PR #20, opened 2024-02-14
Fix doc-tests.

xorgy:


dfrg (approved):
Thanks! I missed these when I pulled in fount.

We should probably also update the peniko dep from git to crates. I’m fine with adding that to this PR if you’d like. Otherwise we can do it as a followup.

xorgy:
> ...Otherwise we can do it as a followup.

I'll do it that way instead, for clarity.

PR #22, opened 2024-03-06
Make Cursor fields public, and remove accessors.

### LINKED ISSUE #23: `is_rtl` never set on `Cursor` when it is made with `from_point`
When a `Cursor` is created with `Cursor::from_point`, `is_rtl` is never set, so is always `false`.

xorgy:
Resolves #23

### DIFF
--- a/src/layout/cursor.rs
+++ b/src/layout/cursor.rs
@@ -5,15 +5,24 @@ use super::*;
 /// Represents a position within a layout.
 #[derive(Copy, Clone, Default, Debug)]
 pub struct Cursor {
-    path: CursorPath,
-    baseline: f32,
-    offset: f32,
-    advance: f32,
-    text_start: usize,
-    text_end: usize,
-    is_rtl: bool,
-    is_leading: bool,
-    is_inside: bool,
+    /// Path to the target cluster.
+    pub path: CursorPath,
+    /// Offset to the baseline.
+    pub baseline: f32,
+    /// Offset to the target cluster along the baseline.
+    pub offset: f32,
+    /// Advance of the target cluster.
+    pub advance: f32,
+    /// Start of the target cluster.
+    pub text_start: usize,
+    /// End of the target cluster.
+    pub text_end: usize,
+    /// Insert point of the cursor (leading or trailing).
+    pub insert_point: usize,
+    /// `true` if the target cluster is in a right-to-left run.
+    pub is_rtl: bool,
+    /// `true` if the cursor was created from a point or position inside the layout
+    pub is_inside: bool,
 }
 
 impl Cursor {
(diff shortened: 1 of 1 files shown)


nicoburns:
Really happy to see the added documentation (field description comments). Makes it much easier to understand what's going on :)

dfrg:
Is there any particular need for the fields to be public? Looking at this now, I had planned to make layouts versioned and have the cursor capture that version on creation so we’d know if it was stale and could update it appropriately. Making it all public would break any guarantees we can make about the validity of a cursor.

xorgy:
While I recognize it introduces a capacity for misuse, it is much more pleasing to destructure what I need from the struct, rather than calling individual fns. When implementing an editor on this interface, I found myself usually throwing out the `Cursor` after using one or two fields out of it. Maybe it'd be better to make static functions that query these values from a layout and a point/position.

xorgy:
Also when implementing a `Selection` helper, I think it's probably not that useful to ingest `Cursor` anyway.

xorgy:
alternative to #24

src/layout/cursor.rs
```diff
@@ -5,15 +5,24 @@ use super::*;
...
-    path: CursorPath,
-    baseline: f32,
-    offset: f32,
-    advance: f32,
-    text_start: usize,
-    text_end: usize,
-    is_rtl: bool,
-    is_leading: bool,
-    is_inside: bool,
+    /// path to the target cluster
```
dfrg, line 8: Admittedly a nit, but for public docs it’d be nice if they started with a capital and ended with a period. This just feels more polished to me when reading rustdocs.

src/layout/cursor.rs
```diff
@@ -5,15 +5,24 @@ use super::*;
...
+    pub path: CursorPath,
+    /// offset to the baseline
+    pub baseline: f32,
+    /// offset to the target cluster along the baseline
+    pub offset: f32,
+    /// advance of the target cluster
+    pub advance: f32,
+    /// start of the target cluster
+    pub text_start: usize,
```
dfrg, line 17: I’m torn on whether to make these fields an actual `Range<usize>`. I think it makes sense now, but I have some planned changes that will make these types a bit more complicated to handle text that spans multiple source objects and `Range` may not fit that. 

I think we’ll leave it as is for now.

dfrg (approved):
Just a couple notes: one nitpick on doc comments and another that’s a bit of rambling about future changes. Otherwise LGTM.

src/layout/cursor.rs
xorgy: `Range` is annoying because it doesn't do `Copy` and breaks the derive macro on this struct. Bare fields are more useful, and you prettymuch always need to reconstruct ranges anyway to use them in more than one place.

src/layout/cursor.rs
xorgy: Fair enough, I'll make that change.

PR #24, opened 2024-03-07
Fix issues in Cursor and simplify logic.

### LINKED ISSUE #23: `is_rtl` never set on `Cursor` when it is made with `from_point`

xorgy:
Resolves #23


xorgy:
#22 was preferred

PR #28, opened 2024-04-11
Add basic Android SystemFonts impl.

xorgy:


xorgy:
I've moved my work over here from my fount repo. I have the fallback stuck to `sans-serif` for everything for the moment, I need to work out proper fallback and there are a few more things that are unclear regarding what parts of fonts.xml to interpret and which parts to ignore.

xorgy:
This stub implementation makes it possible to build and run Xilem, and will be improved afterward, with improvements including:
- Language and script fallback matching, with reasonable behavior for Android expectations.
- Family definitions related back by postScriptName and file name when necessary (the latter may involve adding an index or feature to `scan` to do properly.
- Family-specific fallback with `fallbackFor`, particularly for generic families, e.g. `serif` mapping to _Noto Serif CJK_ (明體) instead of _Noto Sans CJK_ (黑體), or conceivably mapping `cursive` to a 楷書 font if that is called for in the configs.

src/fontique/backend/android.rs
```diff
@@ -0,0 +1,123 @@
...
+                                if let Some((name, to)) =
+                                    child.attribute("name").zip(child.attribute("to"))
+                                {
+                                    if child.attribute("weight").is_some() {
+                                        // weight aliases are an Android quirk and are not in­
+                                        // teresting for use cases other than Android legacy.
+                                        continue;
+                                    }
+                                    let to_n = name_map.get_or_insert(to);
+                                    name_map.add_alias(to_n.id(), name);
```
dfrg, line 76: Just a note: aliases to generic family names will not be resolved as might be expected when parsing font families during layout. That is, if we add an alias for “serif” and the user requests (unquoted) serif, it will never match. Aliases should generally be reserved for family names in alternate languages. Perhaps the `add_alias` method should be renamed to `add_localized_name` to make this more clear.

dfrg (approved):
This looks like a great start, thanks!

PR #29, opened 2024-04-13
Add basic language and script fallback for the Android backend.

xorgy:


xorgy:
This suffers from limitations in the other fallback code, namely that `Hans` and `Hant` are not tried because everything gets collapsed into `Hani`. The specific way that this should be handled depends on the order of system locales on some platforms.

xorgy:
I went ahead and added a workaround that checks if some `lang` fallback includes the `Hant` script, and adds that also for `Hani`. This should be locale-dependent in practice but this makes ambiguous han characters at least show up for now.

xorgy:
One of the limitations of this is that it only functions when `postScriptName` is present on at least one of the `<font...` elements in a fallback family. This is most of the language and script fallback families on common devices, but not all of them; on my device this misses the fallback families for Hebrew.

src/fontique/backend/android.rs
```diff
@@ -87,11 +94,59 @@ impl SystemFonts {
...
+                                                    locale_fallback.push((locale, *family));
+                                                }
+                                            }
+                                        }
+
+                                        // TODO: handle mapping to family names from file names
+                                        //       when postScriptName is unavailable.
+                                    }
+
+                                    // family-specific fallback families, currently unimplimented
```
waywardmonkeys, line 145: "unimplemented"

waywardmonkeys (changes requested):
I don't know enough about what you're actually doing here, but I know enough to spot a typo!

DJMcNab:
I'd be happy to test this on my device if there's a procedure for that, but I don't have the expertise to work out what's happening very quickly, and so could only give a rubber stamping review.

In which case, it should probably be given by Chad.

src/fontique/backend/android.rs
```diff
@@ -87,11 +94,59 @@ impl SystemFonts {
...
+                                                    LanguageIdentifier::try_from_bytes(
+                                                        lang.as_bytes(),
+                                                    )
+                                                {
+                                                    if let Some(scr) = locale.script {
+                                                        script_fallback
+                                                            .push((scr.as_str().into(), *family));
+                                                        if "Hant" == scr.as_str() {
+                                                            // This works around ambiguous han char­
+                                                            // acters going unmapped with current
```
DJMcNab, line 128: I'm not sure I'd put a line break there.
This is also getting very nested, although I haven't reviewed enough surrounding code to see that there's an easy boundary to break things up

src/fontique/backend/android.rs
```diff
@@ -109,15 +164,37 @@ impl SystemFonts {
...
-        Some(self.generic_families.get(GenericFamily::SansSerif)[0])
+    pub fn fallback(&self, key: impl Into<FallbackKey>) -> Option<FamilyId> {
+        let key: FallbackKey = key.into();
+        let script = key.script();
+
+        key.locale()
+            .and_then(|li| {
+                self.locale_fallback
+                    .iter()
+                    .find(|(lid, _)| li == lid.to_string())
```
DJMcNab, line 184: Maybe `find_map` with `some` here?

DJMcNab, line 184: I imagine `to_string` allocates? That's going to be a lot of allocator pressure, I think.

Can we either re-use the same string throughout, or even make a comparison method which doesn't allocate?

src/fontique/backend/android.rs
```diff
@@ -87,11 +94,59 @@ impl SystemFonts {
...
+                                            hasnt_for.iter().partition(|c| {
+                                                c.attribute("postScriptName").is_some()
+                                            });
+
+                                        if let Some(family) = ps_named.iter().find_map(|x| {
+                                            postscript_names
+                                                .get(x.attribute("postScriptName").unwrap())
+                                        }) {
+                                            for lang in langs {
+                                                if let Some(scr) = lang.strip_prefix("und-") {
```
DJMcNab, line 116: Can we have a comment expanding `und` here please? Trying to search for it only brings up the README for https://github.com/behdad/fontconfig, which is not at all enlightening

DJMcNab (approved):
Nothing obviously entirely broken in the Rust code. Hopefully this behaviour will be fully covered in tests once we add those...

src/fontique/backend/android.rs
dfrg: `und` is just a BCP-47 language subtag for “undefined”. It’s used when fallback only depends on script.

dfrg (approved):
In general, I think most of the logic here can be vastly simplified by just using the existing `FallbackMap` type which already handles most of the locale mapping internally. I don’t mind landing this as-is as an intermediate step though if it keeps things moving.

src/fontique/backend/android.rs
xorgy: It's the 99 column limit, my impression was that we were keeping that but...

xorgy:
> I'd be happy to test this on my device if there's a procedure for that, but I don't have the expertise to work out what's happening very quickly, and so could only give a rubber stamping review.
> 
> In which case, it should probably be given by Chad.

There is https://github.com/xorgy/xilem/tree/andtaffy if you have a local parley hooked up in the Cargo.toml. I primarily test this stuff on desktop (you just need to swap the `"fontconfig/mod.rs"` path for `"android.rs"` in your target's cfg of `mod system;`... and then you need to set `$ANDROID_ROOT` to a directory which has the equivalent structure (`adb pull` `/system/etc/fonts.xml` and `/system/fonts/` into a dir, and point `$ANDROID_ROOT` TO THE `system` dir that you create).

src/fontique/backend/android.rs
xorgy: I've swapped it to do the string conversion at parse time, now it just takes a ref from a `Box<str>` instead of converting a `LanguageIdentifier` to `String` at match time.

xorgy: I think in this case that doesn't have any compiler or readability benefit. It would end up being an `if` explicitly returning `Some(...)` or `None`.

src/fontique/backend/android.rs
nicoburns: I second @DJMcNab's suggestion that this is the sort of thing that belongs in the code as a comment. I think we should be aiming for the entire codebase to be explained in such detail, along with a higher-level architectural explanation in a (markdown?) doc (such that it's approachable for someone who doesn't know much if anything about text, but is willing to learn). Obviously it's going to take a while to backfill that for existing code, but we could probably do it more easily for new code.

Perhaps others consider this unnecessary or onerous to the point that it will prevent people from contributing? But if I was reviewing code at work, that's the standard I would expect. And I also feel like it's what I will need to be able to meaningfully contribute to this project (given enough time, I might be able to do a "deep reading" of the code and fill in some this myself, but even that can be tricky around some of the stuff like font fallback and bidi resolution which involve tricky domain-specific knowledge - it's easy to end up understanding what the code is doing, but not *why*, which can make it difficult to work out how to safely modify the code without breaking things).

PR #31, opened 2024-04-15
deps: Update core-text to 0.20, drop foreign-types.

waywardmonkeys:
The foreign-types crate wasn't used directly here, so isn't needed.


dfrg (approved):
The CoreText code was more complex at some point but looks like the dependency remained after simplification. Good catch, thanks!

waywardmonkeys:
Meant to rebase that on rather than a merge commit, but alas ...

PR #34, opened 2024-04-19
Fix typos.

waywardmonkeys:


src/fontique/collection/mod.rs
```diff
@@ -108,7 +108,7 @@ impl Collection {
         self.inner.generic_families(family)
     }
 
-    /// Replaces the set of family identifers associated with the given generic
+    /// Replaces the set of family identifiers associated with the given generic
```
DJMcNab, line 111: I think this is a correct change - at least, searching to see if `identifer` is a font term didn't bring anything up

DJMcNab (approved):

xStrom:
Merging this so it won't conflict with the upcoming CI merge.

PR #36, opened 2024-04-20
Bump fontconfig-cache-parser

jneem:
This cuts down on transitive dependencies by moving things to dev-dependencies.

It also marked the `Object` enum as `#[non_exhaustive]`, hence the version bump.


waywardmonkeys (approved):
❤️ 

(But should we change the MSRV back a few versions now that `clap` isn't forcing it to be so recent?)

DJMcNab:
I think that this pr should probably have included the MSRV decrease, if this was the only blocker

PR #38, opened 2024-04-20
Don't track variations at default location

dfrg:
For variable fonts, when all normalized coordinates are equal to 0, that is equivalent to the default location in variation space and all variation processing can be disabled.

Detect this when pushing a new shaped run and don't store the unnecessary coordinates.

This "fixes" a current performance problem in vello where glyphs from variable fonts are not cached. This effectively forces variable glyphs at the default location to be treated as non-variable allowing them to be cached.


xStrom (approved):
Haven't run the code, but it looks fine.

PR #42, opened 2024-04-26
[deps] Update fontique dep `roxmltree` to 0.19.0

waywardmonkeys:


DJMcNab (approved):
This version has been out for 5 months, so it should be fine

I have not run this code. We really need to work on tests for Parley :lab_coat:

PR #43, opened 2024-04-26
[deps] Revert fontique dep `roxmltree` to 0.18.1

waywardmonkeys:
The update to 0.19.0 broke the masonry tests in the xilem repo on Ubuntu.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -426,9 +426,12 @@ dependencies = [
 
 [[package]]
 name = "roxmltree"
-version = "0.19.0"
+version = "0.18.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "3cd14fd5e3b777a7422cca79358c57a8f6e3a703d9ac187448d0daf220c2407f"
+checksum = "862340e351ce1b271a378ec53f304a5558f7db87f3769dc655a8f6ecbb68b302"
+dependencies = [
+ "xmlparser",
+]
 
 [[package]]
 name = "serde"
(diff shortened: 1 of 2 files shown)


waywardmonkeys:
This reverts #42.

DJMcNab:
This is a not a very compelling fix. Can you test xilem#221 from 9a519ba644eda13783a7326539dd0305dc206ba8 instead, which should have this same tree.

I would much prefer us to just fix this failure, if at all possible. cc @xorgy

waywardmonkeys:
I bisected the tree by running it in CI outside of the Linebender org (my own fork). The memmap2 update and previous commits are okay. Only this last one that updated `roxmltree` makes things sad.

It does show that we really need actual tests for things.

dfrg (approved):
I’m confused as to why updating an xml dependency would break tests. Unfortunately, I probably won’t have time to investigate this today.

waywardmonkeys:
It was several tests failing with:

```
---- widget::align::tests::centered stdout ----
thread 'widget::align::tests::centered' panicked at /home/runner/.cargo/git/checkouts/parley-7d89eb78db3c5efe/9e2be05/src/layout/line/greedy.rs:516:37:
range end index 1 out of range for slice of length 0
```

And that line of code was:

```
    for (i, run_data) in layout.runs[state.runs.clone()].iter().enumerate() {
```

My guess is that the fontconfig failed to load and then things went wrong after that.

dfrg:
At a glance, this might have to do with changes to DTD parsing in roxmltree. Changing the fontconfig backend to use `Document::parse_with_options` with `allow_dtd` set to true might fix it. If not, we’ll likely need to consider a different xml library.

waywardmonkeys:
This isn’t needed now!

PR #44, opened 2024-04-26
[fontique] fontconfig: allow_dtd for xml config

dfrg:
Appears to fix the issue with the roxmltree update.

### DIFF
--- a/fontique/src/backend/fontconfig/config.rs
+++ b/fontique/src/backend/fontconfig/config.rs
@@ -17,7 +17,13 @@ pub fn parse_config(path: &Path, sink: &mut impl ParserSink) {
     let Ok(text) = std::fs::read_to_string(path) else {
         return;
     };
-    let Ok(doc) = roxmltree::Document::parse(&text) else {
+    let Ok(doc) = roxmltree::Document::parse_with_options(
+        &text,
+        roxmltree::ParsingOptions {
+            allow_dtd: true,
+            nodes_limit: u32::MAX,
+        },
+    ) else {
         return;
     };
     let root = doc.root_element();


DJMcNab:
The commit used in https://github.com/linebender/xilem/pull/224 (4e6109f2ff5847a72dc77971b6fa0942b8474d88) doesn't match the head commit of this PR (3dd6f9b823b007724256ce19f78ad1f7ffe5d4cc)

Indeed, 4e6109f2ff5847a72dc77971b6fa0942b8474d88 is apparently 3 weeks old?

fontique/src/backend/fontconfig/config.rs
```diff
@@ -17,7 +17,13 @@ pub fn parse_config(path: &Path, sink: &mut impl ParserSink) {
     let Ok(text) = std::fs::read_to_string(path) else {
         return;
     };
-    let Ok(doc) = roxmltree::Document::parse(&text) else {
+    let Ok(doc) = roxmltree::Document::parse_with_options(
+        &text,
+        roxmltree::ParsingOptions {
+            allow_dtd: true,
+            nodes_limit: u32::MAX,
```
DJMcNab, line 24: I'm guessing we don't care about a malicious file here which is too big to parse?

DJMcNab (approved):
Approving based on xilem#224

My previous comments still stand, but are not blocking

dfrg:
Thanks for catching the bad rev. I have no idea how I ended up with that value.

fontique/src/backend/fontconfig/config.rs
dfrg: This simply retains current behavior. I'm not opposed to placing a limit here but I'm not exactly sure what it would be.

PR #46, opened 2024-04-28
fontique: Remove unused deps.

waywardmonkeys:
Fontique wasn't actually using `anyhow`, `bytemuck`, or `thiserror`.

PR #47, opened 2024-04-28
fontique: Enable builds for iOS and other Apple targets.

waywardmonkeys:
Backend selection for iOS wasn't hooked up.

Also added support for tvOS as it is the same as iOS here.

### DIFF
--- a/fontique/Cargo.toml
+++ b/fontique/Cargo.toml
@@ -35,10 +35,10 @@ dwrote = "0.11.0"
 winapi = { version = "0.3.9", features = ["dwrite", "dwrite_1", "dwrite_3", "winnt", "unknwnbase", "libloaderapi", "winnls"] }
 wio = "0.2.2"
 
-[target.'cfg(any(target_os="macos", target_os="ios"))'.dependencies]
+[target.'cfg(target_vendor="apple")'.dependencies]
 core-text = "20.1.0"
 core-foundation = "0.9.4"
 
-[target.'cfg(not(any(target_os="macos", target_os="ios", target_family="windows")))'.dependencies]
+[target.'cfg(not(any(target_vendor="apple", target_family="windows")))'.dependencies]
 fontconfig-cache-parser = "0.2.0"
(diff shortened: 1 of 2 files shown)


waywardmonkeys:
Opened this as a draft for now for 2 reasons:

1) It'll be in conflict with #46.
2) We don't currently have iOS (or tvOS) CI and I don't think that should block this, but I wanted to call that out.
3) There's a reasonable argument that we should use `target_family = "apple"` rather than `any(target_os = "macos", target_os = "ios", target_os = "tvos")` (and in the future, adding `target_os = "visionos"`.

On the other hand, I should bring up points 2 and 3 in Zulip.

waywardmonkeys:
Zulip thread: https://xi.zulipchat.com/#narrow/stream/419691-linebender/topic/Apple.20family.20target.20configuration

cc: @madsmtm

waywardmonkeys:
Going to adjust this one ...

xorgy:
@waywardmonkeys Out of curiosity, have you actually tried this on tvOS (or whatever they are calling it by the time you reply).

waywardmonkeys:
This just uses CoreText which is the same on all Apple targets. I've built it for tvOS but haven't used it yet. (Need to get wgpu and other things updated.)

Not sure what the "or whatever they're calling it by the time you reply" is meant to mean or imply in any constructive sense.

waywardmonkeys:
This follows along with https://github.com/rust-lang/rust/pull/124491

xorgy:
> This just uses CoreText which is the same on all Apple targets. I've built it for tvOS but haven't used it yet. (Need to get wgpu and other things updated.)

Fair enough. A lot of other things differ, so I didn't assume that CoreText was the same on all the platforms.. as for trying it on tvOS/iOS, you should be able to run it without a renderer as far as I'm aware. That's how the original fontique examples (such as `dump.rs`) did it, and they'd give information about generic families.

xorgy (approved):

PR #53, opened 2024-05-20
lints: Enable `warn(unused_qualifications)`

waywardmonkeys:
This helps point out where the code can be more concistent with an identifier already being in scope and not needing qualifications.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -7,6 +7,7 @@ members = [
 [workspace.package]
 edition = "2021"
 # Keep in sync with RUST_MIN_VER in .github/workflows/ci.yml and with the relevant README.md files.
+# TODO: When this hits 1.74, move lint configuration into this file via a lints table.
 rust-version = "1.70"
 license = "Apache-2.0 OR MIT"
 repository = "https://github.com/linebender/parley"
--- a/fontique/src/backend/fontconfig/config.rs
+++ b/fontique/src/backend/fontconfig/config.rs
@@ -175,10 +175,7 @@ fn include_config(path: &Path, sink: &mut impl ParserSink) -> std::io::Result<()
     Ok(())
 }
 
-fn resolve_dir(
-    node: Node,
-    config_file_path: impl AsRef<std::path::Path>,
-) -> Option<std::path::PathBuf> {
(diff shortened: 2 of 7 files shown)


fontique/src/backend/fontconfig/config.rs
```diff
@@ -175,10 +175,7 @@ fn include_config(path: &Path, sink: &mut impl ParserSink) -> std::io::Result<()
     Ok(())
 }
 
-fn resolve_dir(
-    node: Node,
-    config_file_path: impl AsRef<std::path::Path>,
-) -> Option<std::path::PathBuf> {
+fn resolve_dir(node: Node, config_file_path: impl AsRef<Path>) -> Option<std::path::PathBuf> {
```
waywardmonkeys, line 178: There's a consistency argument to make for this remaining as `std::path::Path` and the other usages of that to do the same and remove the `use std::path::Path` since all other `std` stuff comes in via the fully qualified name.

waywardmonkeys:
This is similar to one that is in Peniko: https://github.com/linebender/peniko/pull/33 and another in Kurbo: https://github.com/linebender/kurbo/pull/346

fontique/src/backend/fontconfig/config.rs
nicoburns: IMO we probably ought to go the other way and have `use` imports for everything in `std` (unless there's a name clash)

nicoburns (approved):
Seems fine to me

Cargo.toml
```diff
@@ -7,6 +7,7 @@ members = [
 [workspace.package]
 edition = "2021"
 # Keep in sync with RUST_MIN_VER in .github/workflows/ci.yml and with the relevant README.md files.
+# TODO: When this hits 1.74, move lint configuration into this file via a lints table.
```
xStrom, line 10: We don't need to wait for the MSRV to be that high to use the lints table. The MSRV `cargo check` commands won't fail, they'll just warn that there is an unused manifest key but the CI will still pass. The stable toolchain will do the actual linting.

This won't be a problem for projects higher up the stack either, as cargo doesn't do linting of dependencies.

PR #54, opened 2024-05-21
Implement simple example using swash to render to png

nicoburns:
This is basically working. But it could definitely do with some actual review and cleanup before merging.

## Output

output

## Notes

- ~Emoji rendering is not working~
- Font fallback for arabic text does seem to be working (note: I pulled the arabic text from cosmic text's sample files. Google translate translate the included arabic text as "arabic", but I cannot otherwise vouch for it's content).
- Subpixel rendering is not enabled (does this make sense when rendering to image?)
- The rendered text doesn't look fantastic. Possibly my blending code (used to apply the Swash alpha mask to an RGBA ImageBuffer from the image buffer is incorrect / too naive?). Cosmic-text pulls in tiny-skia in it's examples. I was hoping to avoid that here, but perhaps it's necessary?
- ~Converting from a `parley::Font` to a `swash::FontRef` is very awkward, specifically when the font file is a collection. The Parley `Font` stores the index of the font within the collection, but Swash's `FontRef` expect the byte offset of the font within the collection. I could not find this exposed anywhere in parley/fontique/skrifa/read-fonts so I'm currently doing pointer arithmetic to get it.~
- ~I'm currently writing to a file called `output.png` in the users current working directory. Possibly we could do better than that (some kind of gitignored "data" directory in the parley repo?)~
- ~I've added an impl of `Brush` for `peniko::Color`, but possibly we ought to just blanket impl for all types implementing the requisite super traits?~
- ~I'm not using the `NormalizedCoords` as they always seemed to be empty in my example (what are they for?)~


examples/simple.rs
```diff
@@ -0,0 +1,236 @@
...
+
+    // Build the builder into a Layout
+    let mut layout: Layout<Color> = builder.build();
+
+    // Perform layout (including bidi resolution and shaping) with start alignment
+    layout.break_all_lines(max_advance, Alignment::Start);
+    let width = layout.width().ceil() as u32;
+    let height = layout.height().ceil() as u32;
+
+    let mut img = RgbaImage::new(width + (padding * 2), height + (padding * 2));
```
waywardmonkeys, line 74: I think you can do something like:

```
let mut img = RgbaImage::from_pixel(width + (padding * 2), height + (padding * 2), Rgba([background_color, ....]));
```

examples/simple.rs
```diff
@@ -0,0 +1,236 @@
...
+    let run = glyph_run.run();
+
+    // Resolve properties of the Run
+    let font = run.font();
+    let font_size = run.font_size();
+
+    // Get byte offset of font within collection (0 if font file is not a collection)
+    // TODO: expose directly in read-fonts
+    let font_collection_ref = font.data.as_ref();
+    let raw_font_ref = ReadFontsRef::from_index(font_collection_ref, font.index).unwrap();
```
dfrg, line 120: Swash has this method: https://docs.rs/swash/0.1.15/swash/struct.FontRef.html#method.from_index

dfrg:
> Emoji rendering is not working

We lost this in the switch to fontique. I can submit a small patch to fix it. 

> Subpixel rendering is not enabled (does this make sense when rendering to image?)

Probably not unless we’re just doing it for effect. Might be nice to add an option at some point. 

> The rendered text doesn't look fantastic.

Black on white without gamma correction is going to look “dirty”. Something simple like `alpha.pow(1.0/2.2)` would improve this a bit. 

> I'm not using the `NormalizedCoords` as they always seemed to be empty in my example (what are they for?)

These are the parameters for controlling interpolation in variable fonts. You should be able to feed the slice directly into `ScalerBuilder::normalized_coords`.

nicoburns:
> We lost this in the switch to fontique. I can submit a small patch to fix it.

Ah. Yes, it seems important!

> Black on white without gamma correction is going to look “dirty”. Something simple like `alpha.pow(1.0/2.2)` would improve this a bit.

Perhaps using these functions? https://observablehq.com/@toja/color-blending-with-gamma-correction.

If I'm understanding correctly the correct process is:

- Convert to float
- Convert to linear rgb
- Perform computation
- Convert srgb
- Convert to u8

> These are the parameters for controlling interpolation in variable fonts. You should be able to feed the slice directly into `ScalerBuilder::normalized_coords`.

I see. Does this also need to be plugged in with skrifa?

dfrg:
See #56 for a patch that might get emoji working with this example

dfrg:
> Perhaps using these functions? https://observablehq.com/@toja/color-blending-with-gamma-correction.
> 
> If I'm understanding correctly the correct process is:
> 
> * Convert to float
> * Convert to linear rgb
> * Perform computation
> * Convert srgb
> * Convert to u8

We can do full sRGB if you'd like but I think just applying the power function to the alpha component will give us reasonably good results for the example.

> > These are the parameters for controlling interpolation in variable fonts. You should be able to feed the slice directly into `ScalerBuilder::normalized_coords`.
> 
> I see. Does this also need to be plugged in with skrifa?

It does. We use the `LocationRef` type there which is really just a wrapper for `&[NormalizedCoord]`. This unfortunately requires a type conversion and allocation as is done in xilem/masonry. We should probably change parley to use this type in the future.

nicoburns:
> We can do full sRGB if you'd like but I think just applying the power function to the alpha component will give us reasonably good results for the example.

The trouble I was having with apply `pow` to the alpha channel is that it's a u8 and I'm unable to apply a fractional power to a u8.

I have implemented an attempt at sRGB <-> Linear RGB conversion. Not sure if I've done it correctly as the result isn't obviously better (but it's good enough that it looks plausibly correct to me). I've left both functions in the code for the time being if you want to inspect them.

Naive:

output_naive

Gamma corrected:

output_gamma

> It does. We use the `LocationRef` type there which is really just a wrapper for `&[NormalizedCoord]`. This unfortunately requires a type conversion and allocation as is done in xilem/masonry. We should probably change parley to use this type in the future.

Ah, gotcha. I'll have a look at plugging this in tomorrow.

dfrg:
Assuming this is using system-ui on macOS, setting the normalized coordinates should also get you the requested bold on the first four characters since San Francisco is a variable font.

nicoburns:
> Assuming this is using system-ui on macOS, setting the normalized coordinates should also get you the requested bold on the first four characters since San Francisco is a variable font.

Ah, that would be good. Bold was working with Helvetica, but it generally looked pretty bad with Helvetica so I switched it out. And I'd like to leave it on `system-ui` as that will presumably pick something sensible cross-platform. I guess maybe I could also use `sans-serif`?

nicoburns:
Pretty sure my gamma correction function is wrong as the serif font looks much better with the naive blending.

examples/simple.rs
```diff
@@ -0,0 +1,263 @@
...
+                            let color = blend_px_naive(*img.get_pixel(x, y), color, alpha);
+                            img.put_pixel(x, y, color);
+                        }
+                        i += 1;
+                    }
+                }
+            }
+            Content::SubpixelMask => unimplemented!(),
+            Content::Color => {
+                for (off_y, row) in rendered_glyph.data.chunks_exact(glyph_width).enumerate() {
```
dfrg, line 214: Should be `glyph_width * 4` since we want the full stride for each row. The image will also need to be blended otherwise it’ll have a black background.

dfrg:
> Pretty sure my gamma correction function is wrong as the serif font looks much better with the naive blending.

Yeah, this is exceptionally tricky to get right. I’ll play with it later this evening.

examples/simple.rs
nicoburns: Ah yes. With this change (+the same blending as I'm using for the alpha mask + #56 applied) emojis are rendering reasonably.

output

waywardmonkeys (dismissed):
I'm assuming that once the other (tiny-skia) lands, this will move to a crate as well?

nicoburns:
> I'm assuming that once the other (tiny-skia) lands, this will move to a crate as well?

Yes. Hoping to land the tiny-skia one (#55) first. Will update this one once we've settled on a design there.

nicoburns:
> > Pretty sure my gamma correction function is wrong as the serif font looks much better with the naive blending.
> 
> Yeah, this is exceptionally tricky to get right. I’ll play with it later this evening.

Hmm... the `image` crate (that we're already using here) has a blend function (https://docs.rs/image/latest/image/struct.Rgba.html#method.blend). Perhaps we could just use that. Looks the same as my "naive" version to me, but it certainly makes the example shorter.

nicoburns:
> These are the parameters for controlling interpolation in variable fonts. You should be able to feed the slice directly into `ScalerBuilder::normalized_coords`.

@dfrg Hmm.. I've added this, but I still don't seem to get bold text.

dfrg:
> > These are the parameters for controlling interpolation in variable fonts. You should be able to feed the slice directly into `ScalerBuilder::normalized_coords`.
> 
> @dfrg Hmm.. I've added this, but I still don't seem to get bold text.

Was just working on this :) Found a bug in swash's avar table. Update to swash 0.1.16 should fix it.

dfrg:
Overall, this is looking pretty good to me now. I'm happy to punt on the gamma issue, get the tiny-skia version landed and then merge this one after making the crate extraction changes.

examples/simple.rs
nicoburns: done

examples/swash_render/src/main.rs
```diff
@@ -0,0 +1,223 @@
...
+                        let x = (glyph_origin_x + off_x) as u32;
+                        let y = (glyph_origin_y + off_y) as u32;
+                        let alpha = rendered_glyph.data[i];
+                        let color = Rgba([color.r, color.g, color.b, alpha]);
+                        img.get_pixel_mut(x, y).blend(&color);
+                        i += 1;
+                    }
+                }
+            }
+            Content::SubpixelMask => unimplemented!(),
```
nicoburns, line 165: @dfrg What are the semantics of the `SubpixelMask`? (/ how would one apply that to an RGBA image?). I reckon it would be good to demonstrate that even if we don't actually enable subpixel mode in the example.

dfrg: This is an rgba image with an alpha component per color channel (the alpha in the mask is ignored). This document (https://github.com/servo/webrender/blob/main/webrender/doc/text-rendering.md#subpixel-text-blending) has a good explanation of the math.

examples/swash_render/src/main.rs
```diff
@@ -0,0 +1,223 @@
...
+                            panic!("Pixel doesn't have 4 components")
+                        };
+                        let x = (glyph_origin_x + off_x as i32) as u32;
+                        let y = (glyph_origin_y + off_y as i32) as u32;
+                        let color = Rgba([r, g, b, a]);
+                        img.get_pixel_mut(x, y).blend(&color);
+                    }
+                }
+            }
+        };
```
nicoburns, line 183: @dfrg Would you be open to adding this pixel iteration code into swash itself(I'm thinking as a method on `Mask` or `swash::Image`)? It would be a nice simplification for this example and could also be used in cosmic-text which has some similar iteration code.

nicoburns: (I don't think we should block this PR on that)

dfrg: Yes, that seems like a useful addition.

nicoburns:
I've rebased this on top of latest `main` and made it fit the new directory structure. It is ready for another round of review.

examples/simple.rs
```diff
@@ -0,0 +1,223 @@
...
+    // Iterate over laid out lines
+    for line in layout.lines() {
+        // Iterate over GlyphRun's within each line
+        for glyph_run in line.glyph_runs() {
+            render_glyph_run(&mut scale_cx, &glyph_run, &mut img, padding);
+        }
+    }
+
+    // Write image to PNG file
+    let output_file = File::create("output.png").unwrap();
```
dfrg, line 92: Should this match the output path of the tiny-skia example?

examples/simple.rs
```diff
@@ -0,0 +1,223 @@
...
+    font_size: f32,
+    glyph_id: GlyphId,
+    normalized_coords: &[NormalizedCoord],
+    x: f32,
+    y: f32,
+) -> Option<SwashImage> {
+    use zeno::{Format, Vector};
+
+    // Build the scaler
+    let mut scaler = context
```
dfrg, line 200: Nit: the scaler settings are dependent on glyph run properties but invariant per glyph so we could build it once per run and pass it in here. This is what I would suggest in real code as a performance optimization so the example might want to reflect that.

examples/swash_render/src/main.rs
```diff
@@ -0,0 +1,223 @@
+// Copyright 2024 the Parley Authors
```
dfrg, line 1: Is this duplicated intentionally?

dfrg (approved):
LGTM with a couple suggestions and some confusion about the code duplication.

examples/swash_render/src/main.rs
nicoburns: No. This file is the correct one. The other file is a rebase conflict.

examples/simple.rs
nicoburns: Fixed

PR #55, opened 2024-05-21
Implement Tiny-Skia example

nicoburns:
This example uses `skrifa` and `tiny-skia` rather than `swash` and `image` as in #54.

## Output

tiny_skia_render

## Notes

- This has the same issue as #54 around emoji rendering not working / being implemented
- There is an issue with the centers of closed letter forms incorrectly being filled
- The code for this is IMO much nicer than the swash version (at the cost of the heavier tiny-skia dependency)
- This example does not have hinting enabled


waywardmonkeys (commented):
Maybe these should have their own `Cargo.toml` so that these things don't all end up being dev dependencies and polluting stuff. That might be better if this repo were using the multi-crate layout...

nicoburns:
> Maybe these should have their own `Cargo.toml` so that these things don't all end up being dev dependencies and polluting stuff. That might be better if this repo were using the multi-crate layout...

Yeah, that makes sense. I believe you can have example be their own crate and have them still work as examples if you set things up correctly (reference them from the Cargo.toml).

xorgy:
Perhaps understandably, I am partial to the example which renders plain text correctly.. but I agree this looks simpler. I also second Bruce's feedback.

xorgy:
Do you think rendering errors are related to a different default fill rule or something similarly easy to change?

dfrg:
> Do you think rendering errors are related to a different default fill rule or something similarly easy to change?

Yes, the fill rule should be non-zero. Looks like it’s set to even-odd here.

nicoburns:
> > Do you think rendering errors are related to a different default fill rule or something similarly easy to change?
> 
> Yes, the fill rule should be non-zero. Looks like it’s set to even-odd here.

It is EvenOdd. Tiny-Skia has `FillRule::EvenOdd` and `FillRule::Winding` (https://docs.rs/tiny-skia/latest/tiny_skia/enum.FillRule.html), but both seem to produce the same result.

examples/tiny-skia.rs
```diff
@@ -0,0 +1,211 @@
...
+            self.x + cx0,
+            self.y - cy0,
+            self.x + cx1,
+            self.y - cy1,
+            self.x + x,
+            self.y - y,
+        );
+    }
+
+    fn close(&mut self) {
```
dfrg, line 198: This should simply call close on the open path. The actual rendering should be done after the call to `OutlineGlyph::draw` completes.

examples/tiny-skia.rs
```diff
@@ -0,0 +1,211 @@
...
+        self.y = y;
+    }
+
+    fn set_color(&mut self, color: TinySkiaColor) {
+        self.paint.set_color(color);
+    }
+
+    fn draw_glyph(&mut self, glyph: &OutlineGlyph<'_>, size: f32) {
+        let settings = DrawSettings::unhinted(Size::new(size), LocationRef::default());
+        glyph.draw(settings, self).unwrap();
```
dfrg, line 169: … so finish and fill_path should be done here

examples/tiny-skia.rs
nicoburns: Ah, I tried that and the fill rule change individually but not together.

Interestingly just moving the code to that location casued a panic when unwrapping the result of `PathBuilder::finish`. But turning that unwrap into an "if let" seems to produce decent visual results even though there are ~12 invalid paths.

waywardmonkeys:
@nicoburns  You'll need to remove a couple of lines from your example's `Cargo.toml` as well.

.github/workflows/ci.yml
```diff
@@ -11,7 +11,9 @@ env:
   RUST_MIN_VER: "1.70"
   # List of packages that will be checked with the minimum supported Rust version.
   # This should be limited to packages that are intended for publishing.
-  RUST_MIN_VER_PKGS: "-p parley -p fontique"
+  PRIMARY_PKGS: "-p parley -p fontique"
```
waywardmonkeys, line 14: You don't use `PRIMARY_PKGS` below, but you do use `PUBIC_PKGS` ...

Is the purpose of renaming this away from `RUST_MIN_VER_PKGS` to denote which are published and not?

examples/tiny_skia_render/Cargo.toml
```diff
@@ -0,0 +1,16 @@
+[package]
+name = "tiny_skia_render"
+version = "0.1.0"
+edition.workspace = true
+rust-version.workspace = true
+license.workspace = true
+repository.workspace = true
+
```
waywardmonkeys, line 8: I think you should add a `publish = false` here.

examples/tiny_skia_render/src/main.rs
```diff
@@ -0,0 +1,224 @@
...
+    let padded_width = width + padding * 2;
+    let padded_height = height + padding * 2;
+
+    // Create TinySkia Pixmap
+    let mut img = Pixmap::new(padded_width, padded_height).unwrap();
+
+    // Fill background color
+    let mut bg_paint = Paint::default();
+    bg_paint.set_color(to_tiny_skia(background_color));
+    img.fill_rect(
```
waywardmonkeys, line 82: I think you could just use `img.fill(to_tiny_skia(background_color));` ?

waywardmonkeys (changes requested):

.github/workflows/ci.yml
nicoburns: Ee're running `check/build/test/doc` etc over `parley` and `fontique` crates with `--features std` and `--features libm`, neither of which work with the `tiny_skia_render` example crate because it doesn't have those features. So we can't use `--workspace` to run those checks. So I'm using this "primary packages" variable to do specify "parley and fontique".

nicoburns:
@dfrg What's the current situation regarding rendering emoji with `skrifa`?

- I can see the [`ColorPainter`](https://docs.rs/skrifa/latest/skrifa/color/trait.ColorPainter.html) trait which seems to be for COLRv1 (and presumably applies to COLRv0?). Although it doesn't look that easy to implement on top of tiny-skia. I think we'd need to do the transforms and clipping ourselves? And I can't find any example implementation anywhere.
- This issue (https://github.com/googlefonts/fontations/issues/593) suggests that there is not currently any support for bitmap emoji formats in skrifa. Is that correct?

dfrg:
> @dfrg What's the current situation regarding rendering emoji with `skrifa`?

Emoji is going to be a heavy lift in this example.

> * I can see the [`ColorPainter`](https://docs.rs/skrifa/latest/skrifa/color/trait.ColorPainter.html) trait which seems to be for COLRv1 (and presumably applies to COLRv0?). Although it doesn't look that easy to implement on top of tiny-skia. I think we'd need to do the transforms and clipping ourselves? And I can't find any example implementation anywhere.

Yes, this covers COLRv1 and v0. The mapping is not difficult but is going to be very tedious. This is implemented in Skia both here: https://github.com/google/skia/blob/907372c4cceee9ab7cf30c8043eba9afd837681e/src/ports/fontations/src/ffi.rs#L148 and here: https://github.com/google/skia/blob/907372c4cceee9ab7cf30c8043eba9afd837681e/src/ports/SkTypeface_fontations.cpp#L1158

Also, tiny-skia is missing some functionality (full two-point conical gradients, sweep gradients and a couple blend modes, I think). We'll probably need to add this functionality ourselves.

> * This issue ([[skrifa] CBDT/CBLC and SBIX bitmap color font access interfaces googlefonts/fontations#593](https://github.com/googlefonts/fontations/issues/593)) suggests that there is not currently any support for bitmap emoji formats in skrifa. Is that correct?

Correct. We have low-level support in read-fonts and for Skia, the additional logic was implemented directly there because we never settled on an API for skrifa. Most of that bitmap logic is here: https://github.com/google/skia/blob/907372c4cceee9ab7cf30c8043eba9afd837681e/src/ports/fontations/src/ffi.rs#L1037

examples/tiny_skia_render/src/main.rs
```diff
@@ -0,0 +1,217 @@
...
+    let run_y = glyph_run.baseline();
+    let style = glyph_run.style();
+    let color = style.brush;
+
+    // Get the "Run" from the "GlyphRun"
+    let run = glyph_run.run();
+
+    // Resolve properties of the Run
+    let font = run.font();
+    let font_size = run.font_size();
```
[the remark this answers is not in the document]
nicoburns: This is now plumbed in. Bold is working after a cargo update.

nicoburns:
> edit: I don't think we should block this PR on emoji support

I am inclined to agree. And given that, I believe this PR may now be ready :)

src/style/brush.rs
```diff
@@ -4,13 +4,4 @@
...
-impl Brush for () {}
-
-/// Brush for a 4-byte color value.
-impl Brush for [u8; 4] {}
-
-/// Brush for a 3-byte color value.
-impl Brush for [u8; 3] {}
-
-impl Brush for peniko::Brush {}
+impl<T: Clone + PartialEq + Default + core::fmt::Debug> Brush for T {}
```
dfrg, line 7: Big +1 on this

dfrg (approved):
I can’t really speak for the CI stuff (which seems fine?) but the code all LGTM. Thanks for taking this on!

xorgy (approved):
It's looking good to me now. I love the detailed comments in the actual example code.

PR #56, opened 2024-05-21
add emoji family for emoji clusters

dfrg:
Quick and dirty fallback for emoji. This can be done more efficiently (and should be handled in fontique directly) but this should get us functioning emoji mapping for capable renderers.

### DIFF
--- a/src/shape.rs
+++ b/src/shape.rs
@@ -10,7 +10,7 @@ use super::style::{Brush, FontFeature, FontVariation};
 use crate::util::nearly_eq;
 #[cfg(feature = "std")]
 use crate::Font;
-use fontique::{self, Attributes, Query, QueryFont};
+use fontique::{self, Attributes, Query, QueryFamily, QueryFont};
 use swash::shape::*;
 #[cfg(feature = "std")]
 use swash::text::cluster::{CharCluster, CharInfo, Token};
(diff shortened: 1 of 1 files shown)


DJMcNab (dismissed):
Is it meaningful to have a test for this?

cc @nicoburns for turning #55 and/or #54 also into integration tests

src/shape.rs
```diff
@@ -211,8 +211,10 @@ impl<'a, 'b, B: Brush> partition::Selector for FontSelector<'a, 'b, B> {
     type SelectedFont = SelectedFont;
 
     fn select_font(&mut self, cluster: &mut CharCluster) -> Option<Self::SelectedFont> {
+        use fontique::QueryFamily;
         let style_index = cluster.user_data() as u16;
-        if style_index != self.style_index {
+        let is_emoji = cluster.info().is_emoji();
+        if style_index != self.style_index || is_emoji {
             self.style_index = style_index;
```
nicoburns, line 218: Do we also need a `self.is_emoji`? I can see this potentially failing to switch back to a regular font after an emoji...

src/shape.rs
```diff
@@ -211,8 +211,10 @@ impl<'a, 'b, B: Brush> partition::Selector for FontSelector<'a, 'b, B> {
     type SelectedFont = SelectedFont;
 
     fn select_font(&mut self, cluster: &mut CharCluster) -> Option<Self::SelectedFont> {
+        use fontique::QueryFamily;
```
nicoburns, line 214: This should be at the top of the file with the other imports

nicoburns:
Applying this on top of the tiny-skia example I get a blank space for the emoji (which is expected because the `ColorPainter` trait and associated function call aren't implemented):

Screenshot 2024-05-22 at 12 22 06

On top of the swash example I get "interesting" output:

Screenshot 2024-05-22 at 12 25 21

I'm assuming this is an issue with the example rather than this PR as it seems to selecting the correct font/glyph.

src/shape.rs
dfrg: Good catch. Checking for `self.font_id.is_none()` should be sufficient. Ultimately I’d like to rework this logic.

dfrg: Added this additional check which will force update of the current font stack following an emoji cluster.

src/shape.rs
dfrg: Done

dfrg:
> Is it meaningful to have a test for this?

It is meaningful but the structure of this file makes unit testing difficult. I filed an issue for refactoring at #57

PR #58, opened 2024-05-23
Remove unused `StyleData`

waywardmonkeys:
`LayoutData` and other things use `Style` instead, leaving this unused and generating a warning.


dfrg (approved):
Must be a remnant from a discarded design. Thanks!

PR #59, opened 2024-05-23
Migrate to multi-crate layout.

waywardmonkeys:
This moves all `parley` items into a subdirectory next to `fontique`.


waywardmonkeys:
This lacks a top level `README.md` as it seems like maybe there isn't a general template yet for this for Linebender repos.

@nicoburns I'd assumed that maybe `examples` would / could still live at the top level. Maybe I'm wrong!

parley/Cargo.toml
```diff
@@ -0,0 +1,26 @@
+[package]
+name = "parley"
+version = "0.1.0"
+description = "Parley provides an API for implementing rich text layout."
```
DJMcNab, line 4: Not really relevant for this PR, but I think the only useful part of this description is `Rich text layout`.

parley/LICENSE-APACHE
DJMcNab, line 1: I think it's still useful to have a copy of this at the top level.

That is, we should have this in this folder *and* at the top level

DJMcNab, line 1: I'm not sure if we should also have a copy of `AUTHORS` in the crate folders. We don't in Xilem, so leaving it out here is fine. But we should have a quick discussion on that

DJMcNab (commented):
Changes look fine, but we *need* a top-level README, even if it's just a stub.

The pattern to follow in this repository is probably [Android Trace](https://github.com/linebender/android_trace), i.e. list out the crates directly

parley/LICENSE-APACHE
xStrom: The `AUTHORS` file is repo wide, as are the copyright headers. This makes maintenance and package splitting easier.

I don't think we should keep copies of the `AUTHORS` file in package directories. It could be done with symlinks, but given that you don't want those, having to keep potentially 5 different `AUTHORS` files in sync is too much friction.

I think continuing to have only a single `AUTHORS` file like we've had thus far is the way to go.

xStrom:
Yeah the top-level readme template isn't finalized. That said, I think it's fine to just keep a copy of the old top-level readme and update it separately from this PR.

parley/LICENSE-APACHE
DJMcNab: Makes sense to me. Just something that I thought now was the opportunity to check

waywardmonkeys:
Okay, I've rebased on top now that the `tiny-skia` sample landed.

I'll fix the clippy thing with a separate PR and then rebase again.

parley/LICENSE-APACHE
waywardmonkeys: I've left a copy of the licenses at top level now.

parley/Cargo.toml
waywardmonkeys: Sounds like a good PR in the future! :)

waywardmonkeys:
> Yeah the top-level readme template isn't finalized. That said, I think it's fine to just keep a copy of the old top-level readme and update it separately from this PR.

For now, I did this and hopefully we can follow up quickly (a day or so?) with a better top level README.

nicoburns (approved):
Happy to approve this as-is now there's a README. That will need updating, but it will be easier to do that work in parallel with other work if this one is merged.

Compile errors (see clippy check failures) will of course need addressing.

DJMcNab (approved):

waywardmonkeys:
Hmm, guess I have to modify CI now ...

waywardmonkeys:
Okay, I think I got the CI this time. I'd made the example use a workspace dependency on `parley` (that I added), but that didn't have default features enabled while the old build did. So now the example still uses that workspace dependency, but enables default-features.

waywardmonkeys:
@nicoburns @DJMcNab If you could re-approve with that revision, that'd be great! And then I'll merge it!

DJMcNab:
Looks like #63 might also fix this. @xStrom which should we land first?

waywardmonkeys:
> Looks like #63 might also fix this. @xStrom which should we land first?

That's fixing a separate issue ... my problem in this one with CI was just my own brain damage, corrected now. :)

DJMcNab (approved):

PR #69, opened 2024-05-30
Add "Parley text stack" section.

PoignardAzur:
Based on writeup by Nico Burns: https://xi.zulipchat.com/#narrow/stream/205635-text/topic/what.20are.20these.20crates/near/439119470

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,6 +15,61 @@
 Parley provides an API for implementing rich text layout.
 It is backed by [Swash](https://github.com/dfrg/swash).
 
+## The Parley text stack
+
+Currently, Parley directly depends on four crates: Fontique, Swash, Skrifa, and Peniko.
+These crates cover different pieces of the text-rendering process.
+
+### Peniko
+
+Peniko builds on top of [kurbo](https://crates.io/crates/kurbo) and provides vocabulary types for 2D rendering.
+
+Peniko mostly just exports types like `Blob`, `Color`, `Gradient`, `Brush`, `Point`, `Rect`, `Vec2`, etc.
+
+### Fontique
+
+Fontique provides font enumeration and fallback.
+
+**Font enumeration** means listing (enumerating) all the fonts installed on the system.
+It also means collecting metadata about those fonts: whether they are serif, sans-serif, monospace, their weight, the code points they cover, etc.
+The library will also load the full font into memory as required using mmap.
+
+**Font fallback** is matching runs of text to a font.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -15,6 +15,61 @@
...
+
+Peniko mostly just exports types like `Blob`, `Color`, `Gradient`, `Brush`, `Point`, `Rect`, `Vec2`, etc.
+
+### Fontique
+
+Fontique provides font enumeration and fallback.
+
+**Font enumeration** means listing (enumerating) all the fonts installed on the system.
+It also means collecting metadata about those fonts: whether they are serif, sans-serif, monospace, their weight, the code points they cover, etc.
+The library will also load the full font into memory as required using mmap.
```
xorgy, line 35: This is not how mmap works. This should probably be something like “The library will use memory-mapped IO , which allows portions to be loaded into memory once and shared between processes on the system.”.. or maybe no attempt should be made to explain mmap.

xorgy (changes requested):
Overall a good overview, just hope that it is kept up to date. I can foresee the portion on Skrifa bits in Swash aging like milk and not being updated.

README.md
waywardmonkeys: 👍 to not mentioning `mmap`

xorgy: There is interest in the Office Hours call to at least mention here that the fonts are memory mapped rather than loaded wholesale.

PoignardAzur: How about this:

> The library will also load the full font into memory as required, using memory-mapped IO for efficiency.

xorgy: That also is misleading, because part of the point is that it doesn't "load the full font into memory". If done correctly, there will be portions of the font that never get loaded into memory.
They are mapped, not loaded.

xorgy: I can give a sentence that is accurate and good, if you tell me what's wrong with what I suggested.

PoignardAzur: No, I think your sentence works. I thought it was maybe too specific, but I don't really understand the stack enough to judge.

xorgy: The crucial/interesting thing about mmap in this circumstance is that it is the norm for accessing system fonts.
That's why I favored getting rid of this sentence: if you don't already know what you're doing with font technology, it's hard to understand why this is important; but if you do know about this stuff then it says something obvious.
The feeling on office hours is that we should mention it, so I'm okay with that but it should be accurate.

PoignardAzur: I think the current version is accurate?

xorgy (approved):
lgtm

README.md
xorgy: It's good. :shipit:

PR #71, opened 2024-06-02
Search correct paths for fonts on Apple platforms.

### LINKED ISSUE #50: Search paths for fonts on Apple platforms
Currently, the CoreText source is searching in `"/System/Library/Fonts/"` for fonts, but it should probably be searching in more than that.

Fonts can be found in:

* The user's home directory library, but this one is tricky.
* `"/Library/Fonts"`
* `"/System/Library/Fonts"`
* `"/Network/Library/Fonts"`

The user's home directory might be tricky due to sandboxing and so we shouldn't just put together the paths ourselves, but use the system APIs for this instead. (In sandboxed applications, one doesn't have direct access to `~/Library/` but has access to something in `~/Library/Containers/<application id>/data/Library` which contains a symlink to the `Fonts` directory.)

I think the right thing to do would be to use `NSSearchPathForDirectoriesInDomains` (<https://developer.apple.com/documentation/foundation/1414224-nssearchpathfordirectoriesindoma?language=objc>) with `NSLibraryDirectory` or (`NSAllLibrariesDirectory`, I don't know which) and a mask of `NSAllDomainsMask`. We should be able to get this from `objc2_foundation` (<https://docs.rs/objc2-foundation/latest/objc2_foundation/fn.NSSearchPathForDirectoriesInDomains.html>)

And finally ... there are ALSO fonts in `/System/Library/Fonts/Supplemental` which should be scanned.

waywardmonkeys:
Use the Foundation API to get the list of "Library" directories and look in each of them for a "Fonts" directory.

This handles the need to search a variety of paths on macOS to find fonts in the various standard locations like "~/Library/Fonts", "/System/Library/Fonts", "/Library/Fonts", "/Network/Library/Fonts", etc.

Fixes #50.

### DIFF
--- a/fontique/src/backend/coretext.rs
+++ b/fontique/src/backend/coretext.rs
@@ -6,6 +6,10 @@ use super::{
 };
 use alloc::sync::Arc;
 use hashbrown::HashMap;
+use objc2::runtime::Bool;
+use objc2_foundation::{
+    NSSearchPathDirectory, NSSearchPathDomainMask, NSSearchPathForDirectoriesInDomains,
+};
 use {
     core_foundation::{
         base::{CFRange, TCFType},
@@ -37,7 +41,17 @@ pub struct SystemFonts {
 
 impl SystemFonts {
     pub fn new() -> Self {
-        let scanned = scan::ScannedCollection::from_paths(Some("/System/Library/Fonts"), 8);
+        let paths = unsafe {
+            NSSearchPathForDirectoriesInDomains(
+                NSSearchPathDirectory::NSLibraryDirectory,
+                NSSearchPathDomainMask::NSAllDomainsMask,
+                Bool::YES,
+            )
+            .as_ref()
+            .iter()
+            .map(|p| format!("{p}/Fonts/"))
+        };
+        let scanned = scan::ScannedCollection::from_paths(paths, 8);
         let name_map = scanned.family_names;
         let mut generic_families = GenericFamilyMap::default();
         for (family, names) in DEFAULT_GENERIC_FAMILIES {
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -51,6 +51,21 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "bef38d45163c2f1dde094a7dfd33ccf595c92905c8f8f4fdc18d06fb1037718a"
 
 [[package]]
+name = "bitflags"
+version = "2.5.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "cf4b9d6a944f767f8e5e0db018570623c85f3d925ac718db4e06d0187adb21c1"
+
+[[package]]
+name = "block2"
+version = "0.5.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "2c132eebf10f5cad5289222520a4a058514204aed6d791f1cf4fe8088b82d15f"
+dependencies = [
+ "objc2",
+]
+
+[[package]]
 name = "bytemuck"
 version = "1.16.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -104,7 +119,7 @@ version = "0.23.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "c07782be35f9e1140080c6b96f0d44b739e2278479f64e02fdab4e32dfd8b081"
 dependencies = [
- "bitflags",
+ "bitflags 1.3.2",
  "core-foundation",
  "core-graphics-types",
  "foreign-types",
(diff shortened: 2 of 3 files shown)


waywardmonkeys:
I'm not sure about a couple of things here, like using string concatenation, the use of `as_ref` on the `NonNull`, and using `to_string` to convert the `NSString` into a `String`.

When I run this locally, I get these search paths:

```
[fontique/src/backend/coretext.rs:52:22] p.to_string() + "/Fonts/" = "/Users/bruce/Library/Fonts/"
[fontique/src/backend/coretext.rs:52:22] p.to_string() + "/Fonts/" = "/Library/Fonts/"
[fontique/src/backend/coretext.rs:52:22] p.to_string() + "/Fonts/" = "/Network/Library/Fonts/"
[fontique/src/backend/coretext.rs:52:22] p.to_string() + "/Fonts/" = "/System/Cryptexes/App/System/Library/Fonts/"
[fontique/src/backend/coretext.rs:52:22] p.to_string() + "/Fonts/" = "/System/Cryptexes/OS/System/Library/Fonts/"
[fontique/src/backend/coretext.rs:52:22] p.to_string() + "/Fonts/" = "/System/Library/Fonts/"
```

If @madsmtm has any comments, that in particular would be great.

fontique/src/backend/coretext.rs
```diff
@@ -37,7 +41,17 @@ pub struct SystemFonts {
...
-        let scanned = scan::ScannedCollection::from_paths(Some("/System/Library/Fonts"), 8);
+        let paths = unsafe {
+            NSSearchPathForDirectoriesInDomains(
+                NSSearchPathDirectory::NSLibraryDirectory,
+                NSSearchPathDomainMask::NSAllDomainsMask,
+                Bool::YES,
+            )
+            .as_ref()
+            .iter()
+            .map(|p| p.to_string() + "/Fonts/")
```
DJMcNab, line 52: This looks like a potential double-allocation hazard - if the path is nearly a power of two length, this will allocate a second time (at least according to https://github.com/rust-lang/rust/issues/16415#issuecomment-65233945).

I that this should be:
```suggestion
            .map(|p| format!("{p}/Fonts/"))
```

waywardmonkeys: Good call! Done and updated.

fontique/src/backend/coretext.rs
```diff
@@ -37,7 +41,17 @@ pub struct SystemFonts {
...
 impl SystemFonts {
     pub fn new() -> Self {
-        let scanned = scan::ScannedCollection::from_paths(Some("/System/Library/Fonts"), 8);
+        let paths = unsafe {
+            NSSearchPathForDirectoriesInDomains(
+                NSSearchPathDirectory::NSLibraryDirectory,
+                NSSearchPathDomainMask::NSAllDomainsMask,
+                Bool::YES,
+            )
+            .as_ref()
```
madsmtm, line 50: `as_ref` in this context is fine as long as you do not return it from this function (the array is placed in the autorelease pool, so that would be unsound). Another alternative would be `Retained::retain_autoreleased(NSSearchPathForDirectoriesInDomains(...).as_ptr()).unwrap()`.

At some point this will be solved in `objc2`.

fontique/src/backend/coretext.rs
```diff
@@ -37,7 +41,17 @@ pub struct SystemFonts {
...
+            .iter()
+            .map(|p| format!("{p}/Fonts/"))
```
madsmtm, line 52: An alternative would be:
```rust
p.stringByAppendingPathComponent(ns_string!("Fonts")).to_string()
```

But the solution you've chosen is probably fine too.

madsmtm (approved):
Memory safety looks fine, and I think using [`NSSearchPathForDirectoriesInDomains`](https://developer.apple.com/documentation/foundation/1414224-nssearchpathfordirectoriesindoma?language=objc) is the correct approach. An alternative, as the docs say, would've been [`-[NSFileManager URLsForDirectory:inDomains:]`](https://developer.apple.com/documentation/foundation/nsfilemanager/1407726-urlsfordirectory?language=objc).

xorgy:
I'm happy to approve this when mads' comments are addressed (either an answer or a code change).

fontique/src/backend/coretext.rs
waywardmonkeys: We're not returning it, so I think just using `as_ref` is fine for now (and certainly shorter than `Retained::retain_autoreleased(...)`)

fontique/src/backend/coretext.rs
waywardmonkeys: If there's an advantage to that, I'm fine making this change ... I don't care either way though, just need a decision.

xorgy (approved):

PR #72, opened 2024-06-03
fontique: Make `icu_properties` optional

waywardmonkeys:
This doesn't seem like everyone using `fontique` will need this (or already have it in their dependency graph).

I left this in `default` since that's the current behavior, but I think there's a good case for not doing so.

### DIFF
--- a/fontique/Cargo.toml
+++ b/fontique/Cargo.toml
@@ -16,6 +16,7 @@ workspace = true
 default = ["system"]
 std = ["skrifa/std", "peniko/std", "dep:memmap2"]
 libm = ["skrifa/libm", "peniko/libm", "dep:core_maths"]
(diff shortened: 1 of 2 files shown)


DJMcNab (approved):
Thanks.

Given how the only usage is in two way mapping, this makes sense.

waywardmonkeys:
I'm going to wait and see if @dfrg thinks we should just remove it from the default features since I'm not in a rush.

waywardmonkeys:
@dfrg `parley` doesn't need this feature to build / run. So I'll just remove it from the default features and people can enable if they find it useful?

dfrg:
Removing from default sounds good to me.

waywardmonkeys:
Updated with it removed from default features.

dfrg (approved):

PR #73, opened 2024-06-06
fontique: Make an explicit feature for `unicode_script`.

waywardmonkeys:
Despite the crate name using a hyphen, the feature uses an underscore to make it match the other Linebender crate features.


waywardmonkeys:
This is preparation for the 2024 edition: https://github.com/rust-lang/cargo/issues/12826

xorgy (approved):

waywardmonkeys:
When we do a changelog, this should get called out.

PR #74, opened 2024-06-06
fontique: Use `icu_properties` for feature name.

waywardmonkeys:
This makes it match the value in `Cargo.toml` and was an error in my previous changes.


xorgy (approved):
Whoopse, I should have caught that too.

PR #75, opened 2024-06-08
Fix cluster index for rtl cursors

msiglreith:
The `cluster_index` field of `CursorPath` is used for relative indexing within cluster range of the corresponding `Run`.
For RTL runs the logical `cluster_index` is calculated incorrectly from the visual cluster index as it uses the global cluster range as reference.

This can be reproduced with the `We will لقاء في 09:35 في ال 🏖️` snippet., where `cluster_index` for cursors within RTL segments may reach values of 20 and higher, while the the individual runs only span a few clusters.


xorgy (approved):
Thank you!

PR #76, opened 2024-06-12
Tree style builder

nicoburns:
- Builds on top of https://github.com/linebender/parley/pull/84 which should be merged first to preserve history

## Changes made

- ### Entire unresolved style struct
     - Adds an "unresolved style struct" (`TextStyle`) to `style/mod.rs`
     - The new `TreeBuilder` allows you to directly pass an entire style struct (rather than just "changed styles" relative to a previous node in the tree) as this is easier for integrating with Stylo which already resolves inherited styles. So I have added an "unresolved" version of the style struct to allow this API to still plug into the other parts of Parley's style resolution functionality.
     - Adds a corresponding`resolve_entire_style_set` method to `LayoutContext` to convert `TextStyle` into `ResolvedStyle`.
- ### Moved code
   - Moves the `RangedBuilder` from `context.rs` to a new `builder.rs` module (as there are now two builders, justifying a separate module).
   - Extract most of `RangedBuilder::build_into` into a standalone `build_into_layout` function (`builder.rs`) that can be shared between the ranged and tree builders.
   - Moves the `RangedStyle` and `RangedProperty` types from `resolve/range.rs` to `resolve/mod.rs`. These types are shared between the `RangedBuilder` and the `TreeBuilder`.
- ### Tree builder
   - Adds a `TreeBuilder` (also to `builder.rs`). This mostly delegates to `TreeStyleBuilder`
   - Add a `TreeStyleBuilder` (`resolve/tree.rs`). This is the vast majority of the new code 
   - The `TreeStyleBuilder` implements HTML-style whitespace collapsing (opt-in). This probably ought to become a style rather than being a flag on the style builder.
   - Updated swash example to optionally use the tree builder (depending on command line opt)

### DIFF
--- a/parley/src/builder.rs
+++ b/parley/src/builder.rs
@@ -1,0 +1,197 @@
+// Copyright 2021 the Parley Authors
+// SPDX-License-Identifier: Apache-2.0 OR MIT
+
+//! Context for layout.
+
+use super::context::*;
+use super::style::*;
+use super::FontContext;
+
+#[cfg(feature = "std")]
+use super::layout::Layout;
+
+use core::ops::RangeBounds;
+
+use crate::inline_box::InlineBox;
+
+/// Builder for constructing a text layout with ranged attributes.
+pub struct RangedBuilder<'a, B: Brush> {
+    pub(crate) scale: f32,
+    pub(crate) lcx: &'a mut LayoutContext<B>,
+    pub(crate) fcx: &'a mut FontContext,
+}
+
+impl<'a, B: Brush> RangedBuilder<'a, B> {
+    pub fn push_default(&mut self, property: &StyleProperty<B>) {
+        let resolved = self
+            .lcx
+            .rcx
+            .resolve_property(self.fcx, property, self.scale);
+        self.lcx.ranged_style_builder.push_default(resolved);
+    }
+
+    pub fn push(&mut self, property: &StyleProperty<B>, range: impl RangeBounds<usize>) {
+        let resolved = self
+            .lcx
+            .rcx
+            .resolve_property(self.fcx, property, self.scale);
+        self.lcx.ranged_style_builder.push(resolved, range);
+    }
+
+    pub fn push_inline_box(&mut self, inline_box: InlineBox) {
+        self.lcx.inline_boxes.push(inline_box);
+    }
+
+    #[cfg(feature = "std")]
+    pub fn build_into(&mut self, layout: &mut Layout<B>, text: impl AsRef<str>) {
+        // Apply RangedStyleBuilder styles to LayoutContext
+        self.lcx.ranged_style_builder.finish(&mut self.lcx.styles);
+
+        // Call generic layout builder method
+        build_into_layout(layout, self.scale, text.as_ref(), self.lcx, self.fcx);
+    }
+
+    #[cfg(feature = "std")]
+    pub fn build(&mut self, text: impl AsRef<str>) -> Layout<B> {
+        let mut layout = Layout::default();
+        self.build_into(&mut layout, text);
+        layout
+    }
+}
+
+/// Builder for constructing a text layout with a tree of attributes.
+pub struct TreeBuilder<'a, B: Brush> {
+    pub(crate) scale: f32,
+    pub(crate) lcx: &'a mut LayoutContext<B>,
+    pub(crate) fcx: &'a mut FontContext,
+}
+
+impl<'a, B: Brush> TreeBuilder<'a, B> {
+    pub fn push_style_span(&mut self, style: TextStyle<B>) {
+        let resolved = self
+            .lcx
+            .rcx
+            .resolve_entire_style_set(self.fcx, &style, self.scale);
+        self.lcx.tree_style_builder.push_style_span(resolved);
+    }
+
+    pub fn push_style_modification_span<'s, 'iter>(
+        &mut self,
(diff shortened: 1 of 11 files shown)


nicoburns:
This one is nearly ready, but needs a bit of cleanup in the tree builder code.

parley/src/builder.rs
```diff
@@ -0,0 +1,197 @@
...
+        .data
+        .styles
+        .extend(lcx.styles.iter().map(|s| s.style.as_layout_style()));
+
+    // Sort the inline boxes as subsequent code assumes that they are in text index order.
+    // Note: It's important that this is a stable sort to allow users to control the order of contiguous inline boxes
+    lcx.inline_boxes.sort_by_key(|b| b.index);
+
+    // dbg!(&lcx.inline_boxes);
+
```
PoignardAzur, line 176:
```suggestion
```

nicoburns:
@dfrg @xorgy Pinging you on this one as it's been a couple of months, I can see that other work on parley is starting to happen, and I am keen to get this merged before it starts to generate conflicts. I'm also keen to do some api documentation / re-export work on Parley and then get a 0.2.0 crates.io release out, which I consider this to be blocking. I do understand that you're busy and that this may not be top priority though.

If it helps:

This PR is actually much smaller and simpler than #84. I have also updated the PR description with a high-level description of the changes which may help with reviewing. I could also split the changes out into separate commits if that would help?

If I were to self-review I suspect I might point out:

- That the whitespace collapsing is a bit awkward:
   - It could probably be made a style rather than a flag on the `TreeStyleBuilder` (I could attempt that change if that's considered important)
   - It doesn't currently integrate well with editing functionality (it would be nice to support this properly, but perhaps for now we could simply document this - in the usual case you wouldn't want to mix these two features anyway)
- There is a bunch of commented out debug code (this is in line with other parts of parley atm and is quite helpful for debugging, but could be removed (or converted to tracing calls?) if it's considered a problem).

dfrg:
Thanks for the ping.

After some thought, and given that you’ve been actively using this code in Blitz, I’m leaning toward landing as is with the caveat that much of this is likely to be superseded with work I’ve been doing on refactoring the text analysis code. The downside is more churn in the future on your end but the upside is support for span margins/borders/padding and a new pull style API for styles so you have more control. Hopefully that’s compelling. 

With that in mind, I’d recommend not investing a lot of effort into improving documentation as things are likely to change soon.

nicoburns:
@dfrg 

> I’m leaning toward landing as is with the caveat that much of this is likely to be superseded with work I’ve been doing on refactoring the text analysis code.

That would be great!

> The downside is more churn in the future on your end but the upside is support for span margins/borders/padding and a new pull style API for styles so you have more control.

I am excited to see what you come up with, and definitely keen for changes that lead to a more complete implementation of the web model (and I'm guessing structural improvements to the code too). API churn isn't really a big deal for us: I'm familiar enough with the code at this point that I'm confident in being able to adapt to any changes. Regression in terms of feature set would be more problematic (if there was no longer a way to do something that was previously possible and that we are relying on) .

> With that in mind, I’d recommend not investing a lot of effort into improving documentation as things are likely to change soon.

The doc changes I have in mind are pretty basic:

- Basic one-line description for key types
- Re-exporting a bunch of (most) types from the top-level of the crate, with `#[doc(inline)]` for important types
- A couple of paragraphs of explanation and a usage example (a pared down version of one of the examples) in the `lib.rs` module documentation.

I feel like a couple of hours spent on those kind of changes would put us in pretty good shape for a 0.2 release, and would bring us much more in line with cosmic-text in terms of first impressions and ease of getting started.

---

Beyond that, there are also a few features that I might be interested in working on soon:

- A style property to control emergency line-breaking behaviour (wrap vs overflow)
- Vertical alignment (and a corresponding style property) + user-specified baselines for inline boxes 
- Floats

My feeling is that the first two (and perhaps even floats) probably wouldn't clash too much with your changes as they deal primarily with the layout code? But if you think it would be better to hold off then I will (I don't suppose you have any kind of estimated timeline for these changes landing / being published?)

nicoburns:
I've rebased on latest `main`:

- Fixed conflicts
- Fixed it to work with the new `vello_editor` example
- Updated it to pass the latest clippy lints

xorgy (approved):
Looks good to me and I think I get it now.

The one caveat being I'd prefer this patch not introduce commented-out printf debugging code.

parley/src/builder.rs
```diff
@@ -0,0 +1,197 @@
...
+    layout.data.text_len = text.len();
+
+    // println!("BUILD INTO ({})", text.len());
+    // for span in &lcx.styles {
+    //     let stack = lcx.rcx.stack(span.style.font_stack);
+    //     println!(
+    //         "{:?} weight:{}, family: {:?}",
+    //         span.range, span.style.font_weight, stack
+    //     );
+    // }
```
xorgy, line 155: I'd nix these commented debugging code blocks.

parley/src/resolve/range.rs
```diff
@@ -124,25 +124,17 @@ impl<B: Brush> RangedStyleBuilder<B> {
             }
         }
         styles.truncate(styles.len() - merged_count);
+
+        // for span in styles.iter() {
+        //     println!("{:?} weight:{}", span.range, span.style.font_weight);
+        // }
+
```
xorgy, line 131: Same issue

PR #77, opened 2024-06-15
Update stable Rust version to 1.79

waywardmonkeys:


DJMcNab (approved):

waywardmonkeys:
Blocking this on #78 rather than listen to what clippy suggests here.

PR #78, opened 2024-06-15
fontique: Remove unsafe in fontconfig cache

waywardmonkeys:
Instead of transmuting the `CharSetLeaf` to the underlying value and then doing the operations on that, use the operation as already exposed via `CharSetLeaf::contains_byte`.

### DIFF
--- a/fontique/src/backend/fontconfig/cache.rs
+++ b/fontique/src/backend/fontconfig/cache.rs
@@ -193,9 +193,7 @@ fn parse_font(
                             .extend_from_slice(set.numbers().ok()?.as_slice().ok()?);
                         for leaf in set.leaves().ok()? {
                             let leaf = leaf.ok()?;
-                            font.coverage
-                                .leaves
-                                .push(unsafe { core::mem::transmute(leaf) });
+                            font.coverage.leaves.push(leaf);
                         }
                     }
                 }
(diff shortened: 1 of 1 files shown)


waywardmonkeys:
I'm unable to test this locally, so some care should be exercised.

In #77, updating to Rust 1.79 enabled a new clippy lint about how `transmute` calls are annotated and instead of doing what it suggested, I thought it would be better to get rid of the transmute entirely instead.

One option was to just call `font.coverage.leaves.push(leaf.map)` instead of transmuting, since the `map` has the right type and is `pub`, so no `transmute` would be required.

That felt a bit odd though to do and I thought maybe this approach would be better since it would use the code that already exists within `fontconfig-cache-parser`.

Is there an issue with this not inlining that code? (If so, could flag that method in a future release of `fontconfig-cache-parser` as `#[inline]` perhaps?)

Or, going by my mental track record of late after illness, I could just be entirely wrong here.

waywardmonkeys:
(Another option might have been to use even more code from `fontconfig-cache-parser`, but I don't know enough at that point about this and the requirements to be sure...)

jneem (approved):
Looks fine to me, although I'm not sure that I have approval authority.

dfrg (approved):
I can’t say why I didn’t structure the code this way originally but the change looks good to me. Thanks!

PR #80, opened 2024-06-26
Add a CHANGELOG

DJMcNab:
Based on linebender/vello#604

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,14 +15,24 @@ You can find its changes [documented below](#010---2024-05-01).
 
 ### Added
 
+### Parley
+
 - Example using tiny-skia which renders into a png ([#55] by [@nicoburns])
-    - breaking: There is now a blanket implementation for `Brush`.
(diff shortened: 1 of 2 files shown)


waywardmonkeys (commented):
I'd been thinking about doing this ... but one thing that held me back (other than time) is that I thought that `parley` and `fontique` should have separate change logs.

DJMcNab:
I think we're planning on doing combined releases for these crates, which suggests a combined changelog makes sense.
But I'm no expert

waywardmonkeys:
I thought part of the appeal of splitting them was that people would be able to use fontique by itself...

DJMcNab:
Sure? But the release strategy of doing the releases at the same time.

Consider [wgpu](https://github.com/gfx-rs/wgpu) for example. They release naga and wgpu at the same time, but projects (such as our `vello_shaders`) can use naga without using wgpu.

nicoburns (approved):

PR #83, opened 2024-07-20
Fontique: parse `-apple-system` and `BlinkMacSystemFont` as system-ui on relevant platforms

nicoburns:
This is for web compatibility. I think we're actually not using it in Blitz anymore (as the interaction between stylo parsing things first and how we then parse that into fontique meant that this code path didn't actually get hit). But I thought it might be worth having anyway? (alternatively I won't be too concerned if this gets rejected).

### DIFF
--- a/fontique/src/generic.rs
+++ b/fontique/src/generic.rs
@@ -77,6 +77,13 @@ impl GenericFamily {
             "emoji" => Self::Emoji,
             "math" => Self::Math,
             "fangsong" => Self::FangSong,
+
+            // Legacy web compatibility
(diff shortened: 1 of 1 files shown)


xorgy (commented):
I am not sure about this, I feel like the place for these is somewhere else (i.e. if you're implementing a CSS/browser layout engine, you lower these to the standard `system-ui` at that level).
I'm not dead-set against it, but it feels like the wrong layer for this hack.

xorgy:
I think that this is ultimately the wrong layer for this. I'm going to close it. Thanks!

PR #84, opened 2024-07-20
Use CSS definition of line height

nicoburns:
Implements CSS style line height computation.

This is now directly on top of main and #76 builds on top of this.

### DIFF
--- a/parley/src/layout/line/greedy.rs
+++ b/parley/src/layout/line/greedy.rs
@@ -472,6 +472,7 @@ impl<'a, B: Brush> BreakLines<'a, B> {
                         // Default vertical alignment is to align the bottom of boxes with the text baseline.
                         // This is equivalent to the entire height of the box being "ascent"
                         line.metrics.ascent = line.metrics.ascent.max(item.height);
+                        line.metrics.line_height = line.metrics.line_height.max(item.height);
 
                         // Mark us as having seen non-whitespace content on this line
                         have_metrics = true;
@@ -504,12 +505,10 @@ impl<'a, B: Brush> BreakLines<'a, B> {
                         // Compute the run's vertical metrics
                         let run = &self.layout.runs[line_item.index];
                         let line_height = line_item.compute_line_height(self.layout);
-                        line.metrics.ascent =
-                            line.metrics.ascent.max(run.metrics.ascent * line_height);
-                        line.metrics.descent =
-                            line.metrics.descent.max(run.metrics.descent * line_height);
-                        line.metrics.leading =
-                            line.metrics.leading.max(run.metrics.leading * line_height);
+                        line.metrics.line_height = line.metrics.line_height.max(line_height);
+                        line.metrics.ascent = line.metrics.ascent.max(run.metrics.ascent);
+                        line.metrics.descent = line.metrics.descent.max(run.metrics.descent);
+                        line.metrics.leading = line.metrics.leading.max(run.metrics.leading);
 
                         // Mark us as having seen non-whitespace content on this line
                         have_metrics = true;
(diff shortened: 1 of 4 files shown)


nicoburns:
The actual change here is ready, but will conflict with https://github.com/linebender/parley/pull/76 which isn't. If we land https://github.com/linebender/parley/pull/67 then we can swap the order of this and #76 so that this lands first if we want to.

xorgy:
I'll take a look at this again after we get `InlineBox` in. I think that there's some question as to whether this is correct to do at this layer, since it introduces questions about rounding.

nicoburns:
I've rebased this to be directly on top of `main` (so we can potentially land it before https://github.com/linebender/parley/pull/76).

I don't see how we'd do this outside of this layer as it controls the spacing between lines (a key part of layout!). And without this change parley doesn't let the user control that as it computes it from the font metrics (in a slightly different way to this - letting ascent + descent + leading control the line height, rather than letting line_height - (ascent + descent) control the leading).

xorgy:
I took a look and I think it's fine the way it is, except the comment on the `line_height` property seems to say that it can be a ‘px value’ which is incorrect.

parley/src/layout/line/mod.rs
```diff
@@ -90,6 +90,8 @@ pub struct LineMetrics {
     pub descent: f32,
     /// Typographic leading.
     pub leading: f32,
+    /// CSS definition of line height (px value or multiplicative value * font size)
```
xorgy, line 93: ‘px value’ is not supported, so should be removed from doc comment.

parley/src/layout/mod.rs
```diff
@@ -197,7 +197,7 @@ pub struct Style<B: Brush> {
     pub underline: Option<Decoration<B>>,
     /// Strikethrough decoration.
     pub strikethrough: Option<Decoration<B>>,
-    /// Multiplicative line height factor.
+    /// Baked line height (multiplicative line height multiplied by font size)
```
xorgy, line 200: I might call this ‘absolute line height (in layout units)’.

parley/src/resolve/mod.rs
```diff
@@ -423,7 +423,7 @@ impl<B: Brush> ResolvedStyle<B> {
             brush: self.brush.clone(),
             underline: self.underline.as_layout_decoration(&self.brush),
             strikethrough: self.strikethrough.as_layout_decoration(&self.brush),
-            line_height: self.line_height,
+            line_height: self.line_height * self.font_size,
```
xorgy, line 426: Should be rounded here or at the other rounding stage.

parley/src/layout/line/greedy.rs
```diff
@@ -559,7 +561,7 @@ impl<'a, B: Brush> BreakLines<'a, B> {
             // Round block/vertical axis metrics
             line.metrics.ascent = line.metrics.ascent.round();
             line.metrics.descent = line.metrics.descent.round();
-            line.metrics.leading = (line.metrics.leading * 0.5).round() * 2.;
+            line.metrics.leading = line.metrics.line_height - (line.metrics.ascent + line.metrics.descent);
```
xorgy, line 564: `line.metrics.line_height` is not rounded at any point as it is, should be rounded here or when it is computed.

xorgy (changes requested):

parley/src/layout/line/greedy.rs
nicoburns: I have gone with rounding here as this is where it is done for the other metrics

parley/src/resolve/mod.rs
nicoburns: I rounded at the other stage

parley/src/layout/mod.rs
nicoburns: Updated

parley/src/layout/line/mod.rs
nicoburns: Removed

xorgy (approved):
:shipit:

PR #85, opened 2024-07-21
[fontique] use windows-rs for dwrite backend

dfrg:
Swaps out winapi, wio and dwrote for windows-core and windows.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -36,6 +36,7 @@ This release has an [MSRV] of 1.70.
 #### Fontique
 
 - Removed unsafe code from fontconfig cache ([#78] by [@waywardmonkeys])
+- Switched to `windows-rs` for `dwrite` backend ([#85] by [@dfrg])
 
 ### Fixed
 
@@ -69,6 +70,7 @@ This release has an [MSRV] of 1.70.
 [#78]: https://github.com/linebender/parley/pull/78
 [#71]: https://github.com/linebender/parley/pull/71
 [#72]: https://github.com/linebender/parley/pull/72
+[#85]: https://github.com/linebender/parley/pull/85
 
 [Unreleased]: https://github.com/linebender/parley/compare/v0.1.0...HEAD
 [0.1.0]: https://github.com/linebender/parley/releases/tag/v0.1.0
--- a/fontique/Cargo.toml
+++ b/fontique/Cargo.toml
@@ -36,9 +36,8 @@ icu_locid = "1.5.0"
 hashbrown = "0.14.5"
 
 [target.'cfg(target_family="windows")'.dependencies]
-dwrote = "0.11.0"
-winapi = { version = "0.3.9", features = ["dwrite", "dwrite_1", "dwrite_3", "winnt", "unknwnbase", "libloaderapi", "winnls"] }
-wio = "0.2.2"
+windows = { version = "0.58.0", features = ["implement", "Win32_Graphics_DirectWrite"] }
+windows-core = { version = "0.58" }
 
 [target.'cfg(target_vendor="apple")'.dependencies]
 core-text = "20.1.0"
--- a/fontique/src/backend/dwrite.rs
+++ b/fontique/src/backend/dwrite.rs
@@ -110,6 +110,8 @@ impl SystemFonts {
     }
 
     pub fn fallback(&mut self, key: impl Into<FallbackKey>) -> Option<FamilyId> {
+        // DirectWrite does not have a function to get the default font for a script and locale pair
+        // so here we provide a sample of the intended script instead.
         let key = key.into();
         let text = key.script().sample()?;
         let locale = key.locale();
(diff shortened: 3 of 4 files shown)


fontique/src/backend/dwrite.rs
```diff
@@ -120,129 +113,270 @@ impl SystemFonts {
...
+                        None,
                         DWRITE_FONT_WEIGHT_REGULAR,
                         DWRITE_FONT_STYLE_NORMAL,
                         DWRITE_FONT_STRETCH_NORMAL,
-                        &mut mapped_length,
-                        &mut font,
-                        &mut scale,
+                        &mut mapped_len,
+                        &mut mapped_font,
+                        &mut 1.0,
```
waywardmonkeys, line 214: For some reason, I feel nervous about this line: `&mut 1.0` especially since it is an `[out]` parameter. (It doesn't get used ... should it be? But the old code didn't use it either.)

fontique/src/backend/dwrite.rs
```diff
@@ -120,129 +113,270 @@ impl SystemFonts {
...
                     )
-                };
-                assert_eq!(hr, 0);
-                if font.is_null() {
-                    i += 1;
+                    .is_ok()
+                {
+                    if let Some(font) = mapped_font {
```
waywardmonkeys, line 218: What happens if there isn't a mapped font?  I assume falling through to `None` case is fine, but the `mapped_len` doesn't seem to be used, instead just skipping a single character.

waywardmonkeys (commented):
I don't have my Windows machine handy, so I can't actually run this, so just going through the patch when DirectWrite isn't my thing, so I have a couple of questions below.

The code looks better overall though!

fontique/src/backend/dwrite.rs
dfrg: We don’t need this value nor do we really need `mapped_len` which could be passed in the same way.

If passing a mutable reference to a literal seems awkward, I can add a temporary but I happen to like this little quirk of Rust :)

fontique/src/backend/dwrite.rs
dfrg: In reality, we’re not making full use of this function. What we actually want is the font DirectWrite would choose for a specific Unicode script and locale. Since there’s no API for this, we cheat by using a table containing sample text for each script (see https://github.com/linebender/parley/blob/caabcf155dc3b8fa43ff4e78945bb87b4c5d8eac/fontique/src/script.rs#L90). That table was generated naively from the UCD and ends up with some characters that DW won’t map so this code walks the given string until a valid font is returned.

Ideally, we’d hand curate the sample table for comprehensive coverage, but so far it’s seemed to work on all platforms (with some tiny hacks like this one) so I haven’t bothered.

waywardmonkeys: That's a great explanation and a subset of this would be a great thing to have in a comment in this code. :)

fontique/src/backend/dwrite.rs
waywardmonkeys: I'm fine with leaving it. Just the disadvantage of having spent so many (so many) years in C / C++. :)

fontique/src/backend/dwrite.rs
dfrg: Fair point! I’ll add a comment describing this glorious hack to hopefully avoid confusion for future readers.

fontique/Cargo.toml
```diff
@@ -33,9 +33,8 @@ icu_locid = "1.4.0"
 hashbrown = "0.14.5"
 
 [target.'cfg(target_family="windows")'.dependencies]
-dwrote = "0.11.0"
-winapi = { version = "0.3.9", features = ["dwrite", "dwrite_1", "dwrite_3", "winnt", "unknwnbase", "libloaderapi", "winnls"] }
-wio = "0.2.2"
+windows = { version = "0.58.0", features = ["implement", "Win32_Graphics_DirectWrite"] }
+windows-core = { version = "0.58" }
```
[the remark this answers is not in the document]
dfrg: Yes, the `implement` derive macro fails to compile without it :/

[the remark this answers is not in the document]
waywardmonkeys: So uncool.

waywardmonkeys:
@dfrg Any chance this could land soon so that it can be part of our next release?

xorgy (approved):

xorgy:
Alright, I'll take a look at this and rebase/merge if it looks good.

CHANGELOG.md
```diff
@@ -36,6 +36,7 @@ This release has an [MSRV] of 1.70.
 #### Fontique
 
 - Removed unsafe code from fontconfig cache ([#78] by [@waywardmonkeys])
+- Switched to `windows-rs` for `dwrite` backend ([#85] by [@dfrg])
```
waywardmonkeys, line 39: You need to add the link for `#85` down below.

fontique/src/backend/dwrite.rs
waywardmonkeys: @xorgy It would be good if some comment about this made its way in.

waywardmonkeys (approved):
Approve with 2 comments (1 of which requires action).

PR #88, opened 2024-07-22
WIP: Import selection and editing code from Masonry

nicoburns:
This is a little earlier than planned as masonry's dependency on vello was making this awkward to use from masonry (when I am not necessary depending on the same version of vello).

- Builds upon #76 which should be merged first.

## Changes made
- Add dependencies on `xi-unicode`, `unicode_segmentation`, `dpi`, and `winit`
- Import `text2` module from masonry
- Import `type ArcStr = Arc<str>` from masonry
- Import event types from masonry 
- Remove vello dependency (drawing is left up to the user)
- Fix imports of other dependencies
- Updated to work with latest parley API

## Todo
- Feature flag editing feature
- Move selection out of editor module?
- Simplify event types
- Add a new (non-masonry) way to be notified when text changes in response to an event


nicoburns:
I've rebased this on main to simplify the diff now that #76 has been merged. I am no longer anticipating this being merged in it's current state, but I'd like to leave it open for now as it contains useful changes that remove dependencies that I think we will want to port to Masonry's version of this code (CC: @PoignardAzur)

This Zulip thread contains my new proposal for making the changes contained in this PR: https://xi.zulipchat.com/#narrow/stream/205635-text/topic/Sharing.20selection.2Fediting.20code.20between.20Blitz.20and.20Masonry

PR #89, opened 2024-07-25
fontique: Fix fontconfig->opentype extra black weight map

tomcur:
OpenType extra black is usually 950. If my understanding is correct, Fontique should map fontconfig's extra black to 950, not 1000.

See also `fontique::Weight::EXTRA_BLACK`

https://github.com/linebender/parley/blob/0f2608cc562d7c2eef0f34819aeb15166d542b67/fontique/src/attributes.rs#L211-L212

### DIFF
--- a/fontique/src/backend/fontconfig/cache.rs
+++ b/fontique/src/backend/fontconfig/cache.rs
@@ -42,7 +42,7 @@ impl Weight {
             (700, 200),
             (800, 205),
             (900, 210),
-            (1000, 215),
+            (950, 215),
         ];
         for (i, (ot, fc)) in MAP.iter().skip(1).enumerate() {
             if weight == *fc {
@@ -58,7 +58,7 @@ impl Weight {
                 return Self::new(ot_a + (ot_b - ot_a) * t);
             }
         }
-        Self::new(1000.0)
+        Weight::EXTRA_BLACK
     }
 }
 


waywardmonkeys (commented):
I don't have a FontConfig based system, but looking at this, I have some questions.

I think the first part with the `MAP` is correct, but the fallback case of returning `Self::new(950.0)` seems wrong to me without having further context.

For one, why not just use `Weight::EXTRA_BLACK`?

But the other things that convert from FontConfig use the default values for style and width (the normal values) rather than one of the extremes (extra black in this case).  Is this the correct behavior?

Further, if the `MAP` were modified a bit, it could store `(Weight, i32)` and then use the const forms to avoid re-hardcoding these constants. That's more than your fix needs to do, but it seems like it would result in better code?

Just some thoughts before @xorgy or @dfrg chime in...

tomcur:
The fallback is hit only in case the input weight is >=215 (fontconfig's equivalent of extra black), so I think it makes sense to default to EXTRA_BLACK in that case (instead of extrapolating).

I'd be happy to make the change to named weights, but can only pick it up after august 12th as I'm currently on holiday.

xorgy (dismissed):

dfrg:
We correctly map extra black to 950 in the `Weight` type so this change looks good to me. If you’d like to replace the magic numbers with named constants, that seems like a solid improvement.

DJMcNab:
(That is the sort of small change which can be made my an approver/merger, so we don't need to wait on the PR author here)

fontique/src/backend/fontconfig/cache.rs
```diff
@@ -58,7 +58,7 @@ impl Weight {
                 return Self::new(ot_a + (ot_b - ot_a) * t);
             }
         }
-        Self::new(1000.0)
+        Self::new(950.0)
```
xorgy, line 61:
```suggestion
        Weight::EXTRA_BLACK
```

xorgy (changes requested):

xorgy (approved):

PR #91, opened 2024-07-25
Update `icu_locid`, `icu_properties`

waywardmonkeys:
This update brings us to zerovec 0.10.4 which addresses a flaw that Dependabot is warning about:

> The affected versions make unsafe memory accesses under the
> assumption that `#[repr(packed)]` has a guaranteed field order.
>
> The Rust specification does not guarantee this, and
> rust-lang/rust#125360 (1.80.0-beta) starts reordering
> fields of `#[repr(packed)]` structs, leading to illegal memory
> accesses.
>
> The patched versions 0.9.7 and 0.10.3 use `#[repr(C, packed)]`,
> which guarantees field order.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -293,9 +293,9 @@ dependencies = [
 
 [[package]]
 name = "icu_collections"
-version = "1.4.0"
+version = "1.5.0"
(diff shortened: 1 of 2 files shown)


DJMcNab (approved):
Seems necessary

PR #97, opened 2024-07-28
fontique: Improve docs

waywardmonkeys:


waywardmonkeys:
I've been writing a fontique example and this fell out of doing that ... more to come before I take it out of draft mode. Comments welcome though.

fontique/src/attributes.rs
```diff
@@ -224,6 +295,16 @@ impl Weight {
...
+    /// # Examples
+    ///
+    /// ```
+    /// # use fontique::Weight;
+    /// assert_eq!(Weight::parse("normal"), Some(Weight::NORMAL));
+    /// assert_eq!(Weight::parse("bold"), Some(Weight::BOLD));
+    /// assert_eq!(Weight::parse("850"), Some(Weight::new(850.0)));
+    /// assert_eq!(Weight::parse("invalid"), None);
+    /// ```
     pub fn parse(s: &str) -> Option<Self> {
```
DJMcNab, line 308: Should `Self` implement `FromStr`?

fontique/src/font.rs
```diff
@@ -140,27 +140,27 @@ impl FontInfo {
         &self.axes
     }
 
-    /// Returns true if the font has a `wght` axis.
+    /// Returns `true` if the font has a `wght` axis.
```
DJMcNab, line 143: Is there a canonical page to link to for these axes?

fontique/src/lib.rs
```diff
@@ -3,6 +3,7 @@
 
 //! Font enumeration and fallback.
 
+#![cfg_attr(docsrs, feature(doc_auto_cfg))]
```
DJMcNab, line 6: :fire: this seems cool

fontique/src/font.rs
waywardmonkeys: My plan is to find one or find somewhere to link to that gives an indication of what you do with this info. (Suggestions welcome.)

fontique/src/attributes.rs
waywardmonkeys: I'm trying to avoid changing actual code here ... but that's a great suggestion, I think for a PR or a help-wanted / easy issue.

fontique/src/font.rs
waywardmonkeys: Found some links, added them.

fontique/src/collection/mod.rs
```diff
@@ -66,6 +66,14 @@ pub struct Collection {
 
 impl Collection {
     /// Creates a new collection with the given options.
+    ///
+    /// If `fontique` was compiled with the `"system"` feature and
+    /// [`CollectionOptions::system_fonts`] was set to `true` when
+    /// creating this collection, then it will register the fonts
+    /// available on the system.
+    ///
+    /// Additional fonts can be registered via [`Collection::register_fonts()`]
```
DJMcNab, line 75:
```suggestion
    /// Additional fonts can be registered via [`Collection::register_fonts`]
```

fontique/src/collection/mod.rs
```diff
@@ -75,12 +83,18 @@ impl Collection {
...
     pub fn family_names(&mut self) -> impl Iterator<Item = &str> + '_ + Clone {
         self.inner.family_names()
     }
 
     /// Returns the family identifier for the given family name.
+    ///
+    /// # See also
+    ///
+    /// * [`Collection::family_name`]
+    /// * [`Collection::family_names`]
```
DJMcNab, line 97: Similarly here

fontique/src/fallback.rs
```diff
@@ -105,6 +105,12 @@ impl FallbackMap {
 }
 
 /// Describes a selector for fallback families.
+///
+/// This is a [`Script`] and optionally, a `locale`, represented
+/// as a `&'static str`.
+///
+/// It can be constructed directly via [`FallbackKey::new()`] or any of
```
DJMcNab, line 112:
```suggestion
/// It can be constructed directly via [`FallbackKey::new`] or any of
```

fontique/src/font.rs
```diff
@@ -230,6 +257,30 @@ const ITALIC_AXIS: u8 = 0x08;
...
+/// * [Width](https://fonts.google.com/knowledge/glossary/width_axis) or `wdth`
+///
+/// # See also
+///
+/// * [`FontInfo::axes()`]
+/// * [`FontInfo::has_italic_axis()`]
+/// * [`FontInfo::has_optical_size_axis()`]
+/// * [`FontInfo::has_slant_axis()`]
+/// * [`FontInfo::has_weight_axis()`]
+/// * [`FontInfo::has_width_axis()`]
```
DJMcNab, line 282: I think we should instead link directly to these at the relevant item in the list above.
With possibly a sentence like:

> FontInfo allows cheaply querying for the presence of these axes using the linked methods.

fontique/src/font.rs
```diff
@@ -253,18 +309,21 @@ pub struct Synthesis {
...
-    /// Returns true if any synthesis suggestions are available.
+    /// Returns `true` if any synthesis suggestions are available.
     pub fn any(&self) -> bool {
         self.len != 0 || self.embolden || self.skew != 0
     }
 
     /// Returns the variation settings that should be applied to match the
     /// requested attributes.
+    ///
+    /// When using `parley`, these can be used to create `FontVariation`
```
DJMcNab, line 320: Maybe? we should link to docs.rs here?

DJMcNab (approved):
Thanks! This looks good.

A few peripheral details about how linking works.

xorgy (commented):
I agree somewhat with Daniel's point about cross linking within the same type. Otherwise I approve.

fontique/src/font.rs
xorgy: How do other crosslinks work on docs.rs? I don't particularly like the idea of absolute links to docs.rs inside the code.

DJMcNab: This is https://github.com/rust-lang/rust/issues/74481
And indeed is similar to https://linebender.org/blog/doc-include/

fontique/src/collection/mod.rs
waywardmonkeys: Done.

fontique/src/font.rs
waywardmonkeys: I've just removed them instead as I didn't like the other options. I think that's a net negative that this info is gone.

fontique/src/font.rs
waywardmonkeys: I left it as-is for someone else to revisit if they like.

fontique/src/attributes.rs
```diff
@@ -77,45 +89,112 @@ impl Stretch {
...
+    ///
+    /// ```
+    /// # use fontique::Stretch;
+    /// assert_eq!(Stretch::from_ratio(1.5), Stretch::EXTRA_EXPANDED);
+    /// ```
+    ///
+    /// # See also
+    ///
+    /// * [`Stretch::from_percentage`]
+    /// * [`Stretch::ratio`]
```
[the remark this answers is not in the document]
waywardmonkeys: I find them very helpful and I think that in some cases, removing them has made these changes worse. When you have something like `Collection` with a variety of methods doing different things, but with some of the same terms in them, it is useful to have an indication of which are similar sets of functionality.

Instead, someone (probably not me) should write a larger doc comment for `Collection` that has sections about the functionality, especially fallbacks, and then links to the right methods.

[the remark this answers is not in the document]
waywardmonkeys: And since your argument is "the Rust std lib doesn't do them", the next move is clear: Get some better cross-linking into the Rust standard lib.

waywardmonkeys:
I think all of the review stuff has been addressed.

xorgy (approved):

fontique/src/attributes.rs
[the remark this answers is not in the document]
DJMcNab: I think the reason that these aren't especially helpful is that these methods are all directly next to each other.

That is, the order of the methods is part of the documentation, and that ordering places the constructors first, and all next to each other, then the corresponding interpretation methods.

Issue #96, opened 2024-07-26
No fallback for styles with missing glyphs


fredizzimo:
Shape the following text
`Some text here. Let's make it a bit longer so that line wrapping kicks in 😊. And also some اللغة العربية arabic text.`
And set the font
`StyleProperty::FontStack(FontStack::Source("DejaVu Serif")))`

The Arabic test is not rendered, even if the DejaVu Serif normal font can support it. 

But due to https://github.com/linebender/parley/issues/92, the style selected is bold! And bold does not support arabic, so the arabic test fails to render.

But even, when bold is desired, I would expect that the normal style is selected with embolden, so that the glyphs can be rendered.

OS: Linux

Issue #95, opened 2024-07-25
Italic CascadiaCove Nerd Font is loaded completely wrong


fredizzimo:
`StyleProperty::FontStack(FontStack::Source("CaskaydiaCove Nerd Font" ))`
`StyleProperty::FontStyle(FontStyle::Italic)`

Loads as semilight with embolden set to true: `CaskaydiaCove NF SemiLight Italic`, but `CaskaydiaCove NF Italic` should be loaded, which has the regular 400 weight, and does not need embolden.

So, this is like a combination of:
* https://github.com/linebender/parley/issues/92
* https://github.com/linebender/parley/issues/93

But in this case embolden is actually set, although a boolean is not very useful for this case, since if no 400 weight was available, it would only need to embolden 50 units, or almost nothing. Compared to almost doubling the weight to make it bold.

OS: Linux

dfrg:
Hazarding a guess, this might have to do with the transform fontconfig applies in attempt to fit all fonts into the RBIZ model. We do try to undo this to achieve proper family grouping but it’s likely that some patterns have been missed leading to poor matching.

nicoburns:
> RBIZ model

For anyone else wondering what this is it's:

- R - Regular
- B - Bold
- I  - Italic
- Z - Bold Italic

And the idea that these are the only 4 font variants (that fits neatly with the typical boolean Bold and Italic buttons in word processors)

Issue #93, opened 2024-07-25
Embolden is not set properly


fredizzimo:
When loading a font that does not contain a bold style (for example ProFontWindows Nerd Font) `StyleProperty::FontStack(FontStack::Source("ProFontWindows Nerd Font"))`  and setting the bold style `StyleProperty::FontWeight(FontWeight::BOLD)`, then the normal font weight is loaded and the embolden synthesis flag is not set.

Note that it is set when using `BLACK` or `EXTRA_BLACK`. 

Furthermore, I think embolden should be a number, not just a boolean, since at least `swash::scale::Render::embolden` takes a floating point strength parameter. So it would be nice to be able to support different weight for the synthesis. Perhaps even optionally embolden when the font weight does not exactly match.

dfrg:
Seems reasonable to change `Synthesis` to report the difference between requested and actual weights so the user can determine a desirable strength.

Issue #92, opened 2024-07-25
Retina weight is loaded instead of regular weight for FiraCode Nerd Font


fredizzimo:
When trying to load "FiraCode Nerd Font" using `StyleProperty::FontStack(FontStack::Source("FiraCode Nerd Font"))` the retina weight (450) is loaded instead of the regular 400 weight.

`FiraCodeNerdFont-Regular.ttf` has the normal 400 weight, so it should be loaded instead.

`NotoSerif Nerd Font` has a similar problem, but there it loads the `Medium` weight instead of `Normal`

This happens at least on Arch Linux with commit 5b60803d1256ab821f79eaa06721a3112de7202a

fredizzimo:
On another Linux machine, it's actually loading the correct weight.

I'm not sure what the difference is, both are Arch Linux install and use the same version of the `FiraCode NerdFont`, and on both installs the Retina weight is available.

xorgy:
@fredizzimo 
Yeah I was getting a bit confused by that, as a fellow Arch Linux user. It's possible that you have overriding configs on one machine that you do not have on the other, or that if you forced cache reconstruction on both they would become the same.
Have you tried running `fc-cache -sf`?

fredizzimo:
I will try to debug exactly what happens during the weekend when I have access to that machine. 

BTW, I have tests for all the bugs I have reported here:
https://github.com/fredizzimo/vide/blob/305d11fd76dd54cd07d7b757172fe65482ec033e/src/test.rs#L305-L756

Those goes through our code, but it would probably be worth introducing something similar here as well.

xorgy:
Thanks, I'll take another look.

fredizzimo:
I have now debugged it, and here's what's happening

The fontconfig information is correct
```
❯ fc-list :family style file weight family | grep FiraCodeNerdFont-
/usr/share/fonts/TTF/FiraCodeNerdFont-Bold.ttf: FiraCode Nerd Font:style=Bold:weight=200
/usr/share/fonts/TTF/FiraCodeNerdFont-Light.ttf: FiraCode Nerd Font,FiraCode Nerd Font Light:style=Light,Regular:weight=50
/usr/share/fonts/TTF/FiraCodeNerdFont-Medium.ttf: FiraCode Nerd Font,FiraCode Nerd Font Med:style=Medium,Regular:weight=100
[...]

PR #104, opened 2024-08-13
fontique: Clear weight, style and stretch when parsing the fontconfig cache

fredizzimo:
The fonts do not always have these pattern elements defined, and then the value of the previous font was wrongly used. Clearing the values, fixes that and ensures that the default values are used.  For a more detailed description see https://github.com/linebender/parley/issues/92#issuecomment-2282674729

* Fixes https://github.com/linebender/parley/issues/93 
* Fixes https://github.com/linebender/parley/issues/95
* Fixes https://github.com/linebender/parley/issues/96


waywardmonkeys (approved):

xorgy (approved):
Thanks!

Issue #109, opened 2024-08-30
Save the range of the text of the corresponding cluster in a glyph


LaurenzV:
Okay, so I've brought this up before in #100, but at that point I didn't really know myself what exactly I need, but now I do, so I wanted to bring it up again as a feature request. :smile: When writing text in PDF, I need to provide a cmap for each glyph, which maps each glyph to a (or multiple) corresponding codepoints.  

So my question is: Is it feasible to extend the 'Glyph' definition so that it stores a range which points to the byte indices of the original string that correspond to the original cluster the glyph stems from?

DJMcNab:
I think this was discussed in #110. Should this issue be closed?

LaurenzV:
Oh yeah, I’ve managed to get something to work, it’s not ideal but it works.

PR #110, opened 2024-08-30
Save range of original text in glyph

LaurenzV:
Okay, so I tried to look into #109 myself, and it actually was very straighforward (I tested it and it seems to work for my purpose), however:
- I had to make `Glyph` not implement copy, since `Range` doesn't.
- I'm a bit unsure why I had to add the text range in two places instead of one and I'm not sure it's 100% correct, so I would appreciate if someone could check this in detail to make sure I didn't mess up anything (assuming that you are okay with this change in the first place). :p


LaurenzV:
Totally understandable, I'll give this a try later, thanks!

PR #111, opened 2024-09-04
parley: add IME support to text editor example

tomcur:
This is a first pass at adding IME support to the editor example.

I don't speak languages that require the use of an IME, so I'm not myself a user of them. I think the implementation is in the right direction.

https://github.com/user-attachments/assets/863677fd-111c-4d39-8e62-b791113a3d68

### DIFF
--- a/examples/vello_editor/src/text.rs
+++ b/examples/vello_editor/src/text.rs
@@ -162,7 +162,7 @@ impl Editor {
             ) {
                 let range = text_at.text_range();
                 self.selection =
-                    Selection::from_index(&self.layout, range.start - 1, Affinity::Upstream);
+                    Selection::from_index(&self.layout, range.start, Affinity::Downstream);
                 self.buffer.replace_range(range, "");
                 self.compose_state = ComposeState::None;
                 // TODO: defer updating layout. If the event itself also causes an update, we now
@@ -329,41 +329,34 @@ impl Editor {
                             return;
                         }
 
-                        let start = self
-                            .delete_current_selection()
-                            .unwrap_or_else(|| self.selection.focus().text_range().start);
-                        self.buffer.insert_str(start, text);
-
-                        {
-                            // winit says the cursor should be hidden when compose_cursor is None.
-                            // Do we handle that? We also don't extend the cursor to the end
-                            // indicated by winit, instead IME composing is currently indicated by
-                            // highlighting the entire preedit text. Should we even update the
-                            // selection at all?
-                            let compose_cursor = compose_cursor.unwrap_or((0, 0));
-                            self.selection = Selection::from_index(
-                                &self.layout,
-                                start - 1 + compose_cursor.0,
-                                Affinity::Upstream,
-                            );
+                        if let Some(start) = self.delete_current_selection() {
+                            self.selection =
+                                Selection::from_index(&self.layout, start, Affinity::Downstream);
                         }
 
                         {
+                            let insertion_index = self.selection.insertion_index();
+                            self.buffer.insert_str(insertion_index, text);
+                            let text_start = Selection::from_index(
+                                &self.layout,
+                                insertion_index,
+                                Affinity::Downstream,
+                            );
                             let text_end = Cursor::from_index(
                                 &self.layout,
-                                start - 1 + text.len(),
-                                Affinity::Upstream,
+                                insertion_index + text.len(),
+                                Affinity::Downstream,
                             );
-                            let ime_cursor = self.selection.extend_to_cursor(text_end);
+                            let text_at = text_start.extend_to_cursor(text_end);
                             self.compose_state = ComposeState::Preedit {
-                                text_at: ime_cursor,
+                                text_at: text_start.extend_to_cursor(text_end),
                             };
 
                             // Find the smallest rectangle that contains the entire preedit text.
                             // Send that rectangle to the platform to suggest placement for the IME
                             // candidate box.
                             let mut union_rect = None;
-                            ime_cursor.geometry_with(&self.layout, |rect| {
+                            text_at.geometry_with(&self.layout, |rect| {
                                 if union_rect.is_none() {
                                     union_rect = Some(rect);
                                 }
@@ -384,6 +377,21 @@ impl Editor {
                                     ),
                                 );
                             }
+                        }
+
+                        {
+                            // winit says the cursor should be hidden when compose_cursor is None.
+                            // Do we handle that? We also don't extend the cursor to the end
+                            // indicated by winit, instead IME composing is currently indicated by
+                            // underlining the entire preedit text, and the IME candidate box
+                            // placement is based on the preedit text location. Should we even
+                            // update the selection based on the compose cursor at all?
+                            let compose_cursor = compose_cursor.unwrap_or((0, 0));
+                            self.selection = Selection::from_index(
+                                &self.layout,
+                                self.selection.insertion_index() + compose_cursor.0,
+                                Affinity::Downstream,
+                            );
                         }
 
                         self.update_layout(self.width, 1.0);
--- a/examples/vello_editor/src/text.rs
+++ b/examples/vello_editor/src/text.rs
@@ -66,12 +66,10 @@ fn shrink_selection(layout: &Layout, selection: Selection, bytes: usize) -> Sele
         focus.index() + shrink
     };
 
-    selection = Selection::from_index(layout, anchor.index(), anchor.affinity());
-    selection = selection.extend_to_cursor(Cursor::from_index(
-        layout,
-        new_focus_index,
-        focus.affinity(),
-    ));
+    selection = Selection::from_cursors(
+        anchor,
+        Cursor::from_index(layout, new_focus_index, focus.affinity()),
+    );
 
     selection
 }
@@ -177,13 +175,10 @@ impl Editor {
             // Send that rectangle to the platform to suggest placement for the IME
             // candidate box.
             let mut union_rect = None;
-            let preedit_selection =
-                Selection::from_index(&self.layout, text_range.start, Affinity::Downstream);
-            let preedit_selection = preedit_selection.extend_to_cursor(Cursor::from_index(
-                &self.layout,
-                text_range.end,
-                Affinity::Downstream,
-            ));
+            let preedit_selection = Selection::from_cursors(
+                Cursor::from_index(&self.layout, text_range.start, Affinity::Downstream),
+                Cursor::from_index(&self.layout, text_range.end, Affinity::Downstream),
+            );
 
             preedit_selection.geometry_with(&self.layout, |rect| {
                 if union_rect.is_none() {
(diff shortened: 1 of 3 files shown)


tomcur:
It might be possible to handle compose keys similarly to IME. That's the initial reason I started to look into this. For example, with altgr composing, dead keys that are about to be composed could be shown. LibreOffice and Firefox do this.

Unfortunately, it seems the events given by winit are not very easy to use for that purpose. For example, when composing, the KeyboardInput event for `"` shows a dead key, but does not indicate it is the logical key `"` nor the text `"`.

DJMcNab:
What windowing system are you using? The compose behaviour is very platform specific.

I implemented this very carefully in https://github.com/linebender/glazier/pull/119, which Winit should probably emulate for Gnome users.
(On KDE it's not possible to do better because the platform runs all the compose through the text input interface)

tomcur:
I've tested on Wayland compositor River 0.3.5 (wlroots). I can try whether Winit on X11 gives different results.

The following from your implementation is exactly what I was going for.

https://github.com/DJMcNab/glazier/blob/76c35367a9eb9f33e49754319ea1580abf0b7dfc/src/text.rs#L470-L484

I've read and also see your changes mention some platforms handle composing as IME – I was hoping that's how this composing behavior was implemented by applications, but unfortunately it doesn't seem to be that simple.

DJMcNab:
Winit absolutely should handle compose as IME, in my estimation (IME!). Ideally, that would lift-and-shift a lot of the code from that pull request being brought into winit.
I'm not going to do that myself, because:
1) I have other priorities
2) KDE, which I currently use, wouldn't use that code.

I do believe that the resulting UX is excellent, though.
I'd recommend trying out compose in the `edit_text` example in Glazier to see what you think.

xorgy:
Thanks. One issue is that you have to replace the preedit text rather than continuing to insert it repeatedly when it gets updated.

tomcur:
Do you mean how in the above example the letters I'm typing are inserted one after the other as preedit text? Typing "ni hao.", the preedit events are:

```
Ime::Enabled
Ime::Preedit("", None)
Ime::Preedit("n", Some((0, 1)))
Ime::Preedit("", None)
Ime::Preedit("ni", Some((0, 2)))
Ime::Preedit("", None)
Ime::Commit("你")
Ime::Preedit("", None)
Ime::Preedit("", None)
Ime::Preedit("h", Some((0, 1)))
Ime::Preedit("", None)
Ime::Preedit("ha", Some((0, 2)))
Ime::Preedit("", None)
Ime::Preedit("hao", Some((0, 3)))
Ime::Preedit("", None)
Ime::Commit("好")
Ime::Preedit("", None)
Ime::Preedit("", None)
Ime::Commit("。")
Ime::Disabled
```

Displaying the full preedit text seems to be the intended behavior here.

(As a side note, I think the preedit events clearing the text in between subsequent preedits are sent by fcitx5. Winit does guarantee one `Ime::Preedit("", None)` to be sent prior to `Ime::Commit`.)

examples/vello_editor/src/text.rs
```diff
@@ -126,7 +140,28 @@ impl Editor {
...
+    pub fn handle_event(&mut self, window: &Window, event: &WindowEvent) {
+        if let ComposeState::Preedit { text_at } = self.compose_state {
+            // Clear old preedit state when handling events that potentially mutate text/selection.
+            // This is a bit overzealous, e.g., pressing and releasing shift probably shouldnt't
+            // clear the preedit.
+            if matches!(
+                event,
+                WindowEvent::KeyboardInput { .. }
+                    | WindowEvent::MouseInput { .. }
+                    | WindowEvent::Ime(..)
```
DJMcNab, line 152: If you remove `::Ime` from here, I think that might be what Aaron meant?

examples/vello_editor/src/text.rs
```diff
@@ -350,6 +466,13 @@ impl Editor {
         if let Some(cursor) = self.selection.focus().weak_geometry(&self.layout, 1.5) {
             scene.fill(Fill::NonZero, transform, Color::LIGHT_GRAY, None, &cursor);
         };
+        if let ComposeState::Preedit { text_at } = self.compose_state {
```
DJMcNab, line 469: Given that this will always do a relayout if this is changing, can we just add a `Style::Underline` to this region when creating the layout? That's the solution we use in Masonry.

examples/vello_editor/src/text.rs
tomcur: The previous preedit state has to be cleared (or otherwise handled) while handling an `Ime::Preedit`. In `Ime::Commit` it's technically not necessary, because Winit guarantees it clears the preedit just prior to a commit.

I started working on a somewhat more performant version where the preedit text is not simply wiped on IME events. Rather, on subsequent preedit events I find the common prefix and update the `Editor::buffer` field accordingly, usually resulting in one fewer memory shift of the editor text (e.g., preedit changing from "n" to "ni", where we don't first have to remove the "n", which shifts everything following to the left), and sometimes no shifts (e.g., preedit changing from "a" to "b", which can be done in-place).

State management becomes more complicated though, so I postponed it until the details have been worked out. Fcitx5 seems to send a clearing preedit between preedits anyway, so there it wouldn't make a practical difference. There (probably?) are IMEs that behave differently.

examples/vello_editor/src/text.rs
```diff
@@ -26,13 +29,24 @@ pub enum ActiveText<'a> {
...
+#[derive(Default)]
+enum ComposeState {
+    #[default]
+    None,
+    Preedit {
+        /// The location of the (uncommitted) preedit text
+        text_at: Selection,
+    },
+}
+
```
tomcur, line 41: If dead key composing is something that we for now hope will silently be converted to IME events by Winit in the future, I don't expect any other `ComposeState` to be added here. In that case this should just become `Option<PreeditState>`.

examples/vello_editor/src/text.rs
tomcur: Thank you, yes, that works great. I hadn't realized underlining was part of Parley, but that makes sense!

parley/src/layout/cursor.rs
```diff
@@ -505,6 +505,16 @@ impl Selection {
...
+    /// Returns a new selection with the focus extended to the given cursor.
+    #[must_use]
+    pub fn extend_to_cursor(&self, focus: Cursor) -> Self {
+        Self {
+            anchor: self.anchor,
+            focus,
+            h_pos: None,
+        }
+    }
+
```
tomcur, line 517: Perhaps this is not necessary. I used `Selection` to get the geometry of the preedit to draw a highlight manually, but now the IME preedit is visually distinguished by pushing a ranged style, that's no longer necessary. The preedit text location can be tracked using a simple byte range.

That does still leave the question of how to handle the IME cursor as given by Winit.

examples/vello_editor/src/text.rs
DJMcNab: I think the main point is that the only case where you need to remove the previous preedit is when a new pre-edit arrives.

tomcur: Different applications handle this differently.

In Firefox when you're entering with IME and then do something else (for example, removing focus from the text field, or changing cursor location within the field by pressing with the pointer), the preedit text is inserted as-is. Firefox also seems to get into a weird state sometimes: after this sequence, starting a new IME input often fails to display preedit text.

In LibreOffice Writer, when you change cursor location while a preedit is shown, the preedit is moved to the new cursor anchor. If you move focus away from the main text area, the preedit is still shown, but when you move focus back, the moment you do anything, the preedit is removed.

Winit correctly keeps the IME state alive when other events are sent; so when changing cursor location, I think we can move the preedit text to the new location, and it will be updated correctly when the user continues typing. It appears we can reset the IME's state when a text field loses focus by sending `Window::set_ime_allowed(false)` (Winit sends an `IME::Disabled` in response.)

tomcur:
I've pushed some improvements. The preedit text now sticks to the selection anchor: if the selection is changed, the preedit text is moved to the new position (e.g., by pressing with the pointer somewhere). When the preedit is started, the selected text is deleted. When a selection is made while preediting, the selected text is deleted when the preedit is continued. When focus is lost (shown at the end of the video), or if the preedit is cleared, the original selection remains.

https://github.com/user-attachments/assets/982978ff-6a0b-4acd-91a2-650ce00c798f

xorgy:
> (This introduces some bugs with some other events like Ctrl+X, need to think about whether there is a nice way to handle all those events. Perhaps some events should just be noops while preediting is active? For example, Firefox blocks pasting while preediting.)

Yes, copy and cut are meaningless during preedit; not sure about paste, but since it's sorta unclear what it should do if the cursor is inside the preedit region, then blocking it during preedit probably makes sense.

examples/vello_editor/src/text.rs
DJMcNab: Hmm, that all matches my understanding, but wasn't really the point I was trying to make. Given the state of this example, i.e. containing a single text field, winit expects you to only change the preedit in response to `Ime::Preedit`.

When changing the cursor position, it really isn't clear what the best behaviour is, although Firefox definitely gets it wrong on my machine.

examples/vello_editor/src/text.rs
tomcur: I've removed the enum in favor of the simpler `preedit_range: Option<Range<usize>>`

examples/vello_editor/src/text.rs
tomcur: Ah, I see. I was talking about the behavior because stopping the preedit from being displayed was a result of not knowing what the correct behavior should be.

I've since changed the behavior so the preedit text is only touched in response to `Ime::Preedit` and `Ime::Disabled`.

tomcur:
Pushed some performance and selection handling improvements. IME preedit now blocks events other than selection change events (so cut/copy/paste, delete, backspace and text insertion—perhaps delete, backspace or text insertion are never sent by Winit during preedit, but I'm not sure).

Maybe selection change should also be blocked during preedit: then all pointer and keyboard events can simply be blocked. There's a case to be made for that:

https://github.com/linebender/xilem/blob/9a3c8e308c3f46f15bb1a5bd48b9119f0178ad98/masonry/src/text/reference/input_component.rs#L160-L166

DJMcNab:
This is running into https://github.com/linebender/vello/pull/618 for me.

However, aside from that, I'm seeing really awful performance when inputting IME. That is, the app gets the "not responding" thing.

tomcur:
> However, aside from that, I'm seeing really awful performance when inputting IME. That is, the app gets the "not responding" thing. I'll try to debug tomorrow morning

Perhaps on your platform it somehow gets into a loop where setting the IME box candidate area triggers an event that triggers setting the area?

DJMcNab:
Yeah, I think that's what's happening.

Worse, this is actually multiplying. That is, every frame an extra iteration appears to be added.

DJMcNab:
Before I can approve, I'd like to see the IME looping resolved. I probably won't have time to dig into that myself in the next couple of weeks, unfortunately. But if that becomes resolved, then I'll happily approve, as this code otherwise looks good!

We *probably* need to cache the last sent IME cursor area, and not send it again if we would be sending the same value. It is a bit stupid of my platform's IME implementation to send an IME preedit event in response to the movement, but we have to deal with that.

tomcur:
> We _probably_ need to cache the last sent IME cursor area, and not send it again if we would be sending the same value.

I'd be happy to make the suggested change. On my platform I cannot reproduce the issue, so I'll have to ping you to see if the issue is resolved.

Hopefully that process doesn't loop.

tomcur:
I have a patch with this solution, but working on it I realized it may also be resolved by checking on `Ime::Preedit` events whether the new selection is different from the old selection. If and only if so, send the new cursor area to the platform. If that works, we don't grow the memory footprint of `Editor`.

examples/vello_editor/src/text.rs
```diff
@@ -265,6 +372,129 @@ impl Editor {
...
+                                // caret just in front of the preedit.
+                                self.selection = Selection::from_index(
+                                    &self.layout,
+                                    preedit_start,
+                                    Affinity::Downstream,
+                                );
+                            }
+                        }
+
+                        if self.selection != old_selection {
```
tomcur, line 473: The main

examples/vello_editor/src/text.rs
```diff
@@ -265,6 +372,129 @@ impl Editor {
...
+                            self.set_ime_cursor_area(window);
+                        }
```
tomcur, line 475: A potential issue I see with this is if for some reason `self.selection == old_selection`, but we have not sent the area yet, the platform is unaware of the area.

Maybe we can unconditionally send the area in `Ime::Enabled`.

tomcur: (In the masonry case, this would have to calculate window coordinates of the selection)

examples/vello_editor/src/text.rs
```diff
@@ -126,11 +168,54 @@ impl Editor {
         // TODO: support clipboard on Android
     }
 
-    pub fn handle_event(&mut self, event: &WindowEvent) {
+    /// Suggest an area for IME candidate box placement based on the current IME state.
+    fn set_ime_cursor_area(&self, window: &Window) {
+        if let Some(ref text_range) = self.preedit_range {
+            // Find the smallest rectangle that contains the entire preedit text.
+            // Send that rectangle to the platform to suggest placement for the IME
+            // candidate box.
```
DJMcNab, line 176: Hmm, I'm not sure that this is quite the right logic. For example, consider the Windows emoji picker; at least ideally it uses the cursor area, but doesn't add a pre-edit text.
I think that this whole statement needs to have an else branch, which uses something based on the selection.
My proposal would be to use one of:
1) The entire selection area;
2) The area of the `focus`ed cluster

DJMcNab (commented):
I don't think this solution for placing the cursor area works for e.g. the windows emoji picker. That is, not all IME inputs would necessarily use a cursor area.

examples/vello_editor/src/text.rs
tomcur: I think you're right. Perhaps the area should always be based on the selection: either there is no preedit, and the selection is the only thing that makes sense, or there is a preedit, and the IME gave us a selection within it (`Ime::Preedit(_, cursor)`). Although, if the preedit wraps lines and the preedit cursor is a partial selection, the candidate box may then overlap.

I'm also a bit confused by the name [`Window::set_ime_cursor_area`](https://docs.rs/winit/latest/winit/window/struct.Window.html#method.set_ime_cursor_area), as that seems to suggest the area of the IME cursor (as given by `Ime::Preedit(_, cursor)`) is to be given, whereas the documentation states: "an example of such area could be a input field in the UI or line in the editor."

DJMcNab: Yeah; this API is underdocumented from the winit side. I don't have good answers here.

I think the IME box should never cover the pre-edit text, so if there is a pre-edit, I think the current answer is fine.

tomcur: The area is now calculated as the minimum rect containing the preedit text, if present. If not, it depends on whether the selection is collapsed. A collapsed selection sends the caret area, a spanning selection is the minimum rect containing the selection.

tomcur:
@DJMcNab as you mentioned IMEs that do not necessarily set preedit text (e.g., emoji pickers), I've made the IME state tracking a bit more complete. `Editor` now also tracks whether the IME is enabled, instead of only tracking whether there is a preedit.

If the IME is enabled, candidate areas are sent. I think the logic in the most recent commit is correct, but I don't have an easy way to test.

xorgy:
I will add the requisite support in `PlainEditor` for Ime, and fix this branch up for that. 0.3 should be a reasonably quick followup.

examples/vello_editor/src/text.rs
```diff
@@ -326,6 +594,56 @@ impl Editor {
...
+                self.update_layout(self.width, 1.0);
+
+                self.selection = self.selection.refresh(&self.layout);
+            }
+        }
+
+        if self.ime != ImeState::Disabled
+            && !matches!(
+                event,
+                WindowEvent::RedrawRequested | WindowEvent::CursorMoved { .. }
```
DJMcNab, line 636: Can  `CursorMoved` change the selection if we're currently dragging?

tomcur: Yes. Currently the selection is also changed if we click somewhere. The editor state remains consistent: the IME preedit is moved to the new anchor position, and the span of the selection itself is mostly a visual thing. The user can't delete, copy, paste, etc., while a preedit is active.

What we should do here is not clear. Moving the preedit text to the new anchor location feels a bit clunky, and that's even with the IME selecting the entire preedit. If only part of the preedit is selected, it'll probably feel even stranger.

tomcur: Ah, I see what you mean. Indeed, I think that won't really work as intended. The comment I wrote in that section: 

```
// TODO: this is a bit overzealous in recalculating the IME cursor area: there are
// cases where it can cheaply be known the area is unchanged. (If the calculated area
// is unchanged from the previous one sent, it is not re-sent to the platform, though).
//
// Ideally this is called only when the layout, selection, or preedit has changed, or
// if the IME has just been enabled.
```

(Clearly it's not only overzealous, it's also slightly buggy.)

Maybe the new `PlainEditor` should make it cheap to know whether the layout changed. I'd prefer calling out to the platform as conservatively as possible, as there's no control over the cost of those calls.

tomcur:
Superseded by https://github.com/linebender/xilem/pull/762, #192.

PR #123, opened 2024-10-04
Improvements to docs 

nicoburns:
- Simplify public module hierarchy (make some modules private and re-export types instead)
- Re-export all types at the top level (making use of wildcard re-exports to avoid cluttering the docs)
- Add a usage example and basic documentation to `lib.rs`

### DIFF
--- a/parley/src/context.rs
+++ b/parley/src/context.rs
@@ -20,7 +20,7 @@ use swash::text::cluster::CharInfo;
 use crate::builder::TreeBuilder;
 use crate::inline_box::InlineBox;
 
-/// Context for building a text layout.
+/// Shared scratch space used when constructing text layouts
 pub struct LayoutContext<B: Brush = [u8; 4]> {
     pub(crate) bidi: bidi::BidiResolver,
     pub(crate) rcx: ResolveContext,
--- a/parley/src/context.rs
+++ b/parley/src/context.rs
@@ -20,7 +20,9 @@ use swash::text::cluster::CharInfo;
 use crate::builder::TreeBuilder;
 use crate::inline_box::InlineBox;
 
-/// Shared scratch space used when constructing text layouts
+/// Shared scratch space used when constructing text layouts.
+///
+/// This type is designed to be a global resource with only one per-application (or per-thread).
 pub struct LayoutContext<B: Brush = [u8; 4]> {
     pub(crate) bidi: bidi::BidiResolver,
     pub(crate) rcx: ResolveContext,
--- a/parley/src/font.rs
+++ b/parley/src/font.rs
@@ -6,9 +6,9 @@ use fontique::Collection;
 #[cfg(feature = "std")]
 use fontique::SourceCache;
 
-/// A font database/cache (wrapper around a the [fontique] crate)
+/// A font database/cache (wrapper around a Fontique [`Collection`] and [`SourceCache`]).
 ///
-/// This type is designed to be a global resource with only one per-application (or per-thread)
+/// This type is designed to be a global resource with only one per-application (or per-thread).
 #[derive(Default)]
 pub struct FontContext {
     pub collection: Collection,
@@ -17,6 +17,7 @@ pub struct FontContext {
 }
 
 impl FontContext {
+    /// Create a new `FontContext`, discovering system fonts if available.
     pub fn new() -> Self {
         Default::default()
     }
--- a/parley/src/font.rs
+++ b/parley/src/font.rs
@@ -6,9 +6,18 @@ use fontique::Collection;
 #[cfg(feature = "std")]
 use fontique::SourceCache;
 
+/// A font database/cache (wrapper around a the [fontique] crate)
+///
+/// This type is designed to be a global resource with only one per-application (or per-thread)
 #[derive(Default)]
 pub struct FontContext {
     pub collection: Collection,
     #[cfg(feature = "std")]
     pub source_cache: SourceCache,
+}
+
+impl FontContext {
+    pub fn new() -> Self {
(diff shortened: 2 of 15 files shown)


parley/src/context.rs
```diff
@@ -20,7 +20,7 @@ use swash::text::cluster::CharInfo;
 use crate::builder::TreeBuilder;
 use crate::inline_box::InlineBox;
 
-/// Context for building a text layout.
+/// Shared scratch space used when constructing text layouts
```
DJMcNab, line 23:
```suggestion
/// Shared scratch space used when constructing text layouts.
///
/// This type is designed to be a global resource with only one per-application (or per-thread).
```
See https://github.com/linebender/xilem/pull/405

parley/src/font.rs
```diff
@@ -6,9 +6,18 @@ use fontique::Collection;
 #[cfg(feature = "std")]
 use fontique::SourceCache;
 
+/// A font database/cache (wrapper around a the [fontique] crate)
```
DJMcNab, line 9: a Fontique [`Collection`]?

parley/src/context.rs
```diff
@@ -20,7 +20,7 @@ use swash::text::cluster::CharInfo;
 use crate::builder::TreeBuilder;
 use crate::inline_box::InlineBox;
 
-/// Context for building a text layout.
```
DJMcNab, line 23: It's not clear to me why you've removed the full stops from these comments. Our docs style (matching `std`) is to have a full stop at the end of the first sentence.

parley/src/font.rs
```diff
@@ -6,9 +6,18 @@ use fontique::Collection;
...
+///
+/// This type is designed to be a global resource with only one per-application (or per-thread)
```
DJMcNab, line 11: This wording is very clear :fire:

parley/src/font.rs
```diff
@@ -6,9 +6,18 @@ use fontique::Collection;
...
 #[derive(Default)]
 pub struct FontContext {
     pub collection: Collection,
     #[cfg(feature = "std")]
     pub source_cache: SourceCache,
 }
+
+impl FontContext {
+    pub fn new() -> Self {
```
DJMcNab, line 20: Given that this does (iirc) search the system font libraries, we should have a doc comment here.

DJMcNab:
```suggestion
    /// Create a new `FontContext`, discovering system fonts if available.
    pub fn new() -> Self {
```

DJMcNab: Obviously this API could use some work, but that's not the job of this PR.

parley/src/lib.rs
```diff
@@ -1,7 +1,76 @@
...
+//! layout.align(MAX_WIDTH, Alignment::Start);
+//!
+//! // Inspect computed layout (see examples for more details)
+//! let width = layout.width();
+//! let height = layout.height();
+//! for line in layout.lines() {
+//!     for item in line.items() {
+//!         match item {
+//!             PositionedLayoutItem::GlyphRun(glyph_run) => {
+//!                 // Do something with  glyph run
```
DJMcNab, line 65: Should we just say render here? Yes, technically you could want to do something else, but the vast majority of our users will want to render

DJMcNab (commented):
Overall looks good, just a few peripheral thoughts

parley/src/font.rs
waywardmonkeys: "a the" ... choose one!

parley/src/lib.rs
```diff
@@ -1,7 +1,76 @@
 // Copyright 2021 the Parley Authors
 // SPDX-License-Identifier: Apache-2.0 OR MIT
 
-//! Rich text layout.
+//! Parley is library for rich text layout.
+//!
+//! Key types are:
```
waywardmonkeys, line 6: "The key types are"

waywardmonkeys: Or maybe not. Up to you!

parley/src/lib.rs
```diff
@@ -1,7 +1,76 @@
...
+//! - [`FontContext`] and [`LayoutContext`] are resources which should be shared globally (or at coarse-grained boundaries).
+//!   - [`FontContext`] is database of fonts.
+//!   - [`LayoutContext`] is scratch space that allows for reuse of allocations between layouts.
+//! - [`RangedBuilder`] and [`TreeBuilder`] which are builders for creating a [`Layout`].
+//!     - [`RangedBuilder`] allows styles to be specified as a flat `Vec` of spans
+//!     - [`TreeBuilder`] allows styles to be specified as a tree of spans
+//!   
+//!   They are constructed using the [`ranged_builder`](LayoutContext::ranged_builder) and [`tree_builder`](LayoutContext::ranged_builder) on [`LayoutContext`].
+//! - [Layout] which represents styled paragraph(s) of text and can perform shaping, line-breaking, bidi-reordering, and alignment of that text.
```
waywardmonkeys, line 15: Missing backticks.

parley/src/font.rs
nicoburns: > a Fontique [Collection]?

Well yes. Although there's also a fontique SourceCache. I'm not quite sure what the purpose of the SourceCache is.

parley/src/context.rs
nicoburns: No particular reason. Just wasn't paying attention to them.

nicoburns:
> please make sure that this doesn't have warnings or (new) clippy lints with a current nightly from your lifetime changes.

The lifetime changes were to make nightly clippy happy in the first place!

PR #128, opened 2024-10-09
CoWify style types

nicoburns:
This:
- Converts parley style types to use `std::borrow::Cow` instead of references
- Removes the lifetime parameter from `PlainEditor` and `PlainEditorOp` by making them using `StyleProperty<'static>`

This isn't ideal from an API point of view, but does at least allow `PlainEditor` to be stored within a struct.

It would be an option to leave the lifetime parameter in `PlainEditor` and make uses who want to store it in a struct specify `PlainEditor<'static>` themselves. I have not done so because I don't think there's much use for the borrowed editor outside of toy examples.

### DIFF
--- a/examples/vello_editor/src/main.rs
+++ b/examples/vello_editor/src/main.rs
@@ -49,7 +49,7 @@ struct SimpleVelloApp<'s> {
     scene: Scene,
 
     // Our text state object
-    editor: text::Editor<'s>,
+    editor: text::Editor,
 }
 
 impl ApplicationHandler for SimpleVelloApp<'_> {
--- a/examples/vello_editor/src/main.rs
+++ b/examples/vello_editor/src/main.rs
@@ -2,6 +2,7 @@
 // SPDX-License-Identifier: Apache-2.0 OR MIT
 
 use anyhow::Result;
+use std::borrow::Cow;
 use std::num::NonZeroUsize;
 use std::sync::Arc;
 use vello::peniko::Color;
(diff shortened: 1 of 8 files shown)


examples/vello_editor/src/main.rs
```diff
@@ -135,7 +136,7 @@ impl ApplicationHandler for SimpleVelloApp<'_> {
                         parley::style::StyleProperty::FontSize(32.0),
                         parley::style::StyleProperty::LineHeight(1.2),
                         parley::style::StyleProperty::FontStack(parley::style::FontStack::Source(
-                            "system-ui",
+                            Cow::Borrowed("system-ui"),
                         )),
```
xorgy, line 140: To me, this seems really ugly, but I do like the benefits at other levels...

nicoburns: This particular case (constructing styles) could be improved by adding constructors to the style types. Perhaps with with some From/Into implementations so that they can be overloaded.

xorgy: That seems like the way to go yeah.
For now this example should just be 
```Rust
                        StyleProperty::FontStack(FontStack::Single(FontFamily::Generic(
                            GenericFamily::SystemUi,
                        ))),
```
anyway I think.

xorgy (approved):

examples/vello_editor/src/main.rs
nicoburns: If we really wanted to make it ergonomic we could probably make something like:
```
FontStack::Single(FontFamily::SYSTEM_UI).into()
```

Not convinced ergonomics are that important for parley though. I doubt many people will be interacting with it directly. We have a similar thing with Taffy: we could add style builder but in most cases it actually makes more sense if that sits at the level above.

xorgy: True...

PR #129, opened 2024-10-10
Make StyleProperty::FontStack more ergonomic.

xorgy:


parley/src/style/font.rs
```diff
@@ -83,6 +83,18 @@ impl<'a> FontFamily<'a> {
...
+        FontFamily::Generic(self)
+    }
+}
+
+impl<'a> Into<FontStack<'a>> for &'a str {
+    fn into(self) -> FontStack<'a> {
+        FontStack::Source(Cow::Borrowed(self))
+    }
+}
+
```
nicoburns, line 97: I believe it's recommend to implement `From` rather than `Into` when you own both types (there's a blanket impl of Into for types that implement From).

examples/tiny_skia_render/src/main.rs
```diff
@@ -55,9 +56,7 @@ fn main() {
     builder.push_default(&brush_style);
 
     // Set default font family
-    let font_stack = FontStack::Source(Cow::Borrowed("system-ui"));
-    let font_stack_style = StyleProperty::FontStack(font_stack);
-    builder.push_default(&font_stack_style);
+    builder.push_default(&GenericFamily::SystemUi.into());
```
nicoburns, line 59: Could make `push_default` take `impl Into<StyleProperty>` and eliminate this `.into()`.

nicoburns (approved):
Looks good. A couple of suggestions but nothing major so also approving.

parley/src/style/font.rs
xorgy: Addressed.

examples/tiny_skia_render/src/main.rs
xorgy: I like that, I'll add it.

PR #130, opened 2024-10-10
Prepare Parley and Fontique 0.2.0

xorgy:


nicoburns (approved):

waywardmonkeys (approved):

fontique/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "fontique"
-version = "0.1.0" # Keep in sync with workspace dependency specification
+version = "0.2.0" # Keep in sync with workspace dependency specification
```
DJMcNab, line 3: This should probably use `version.workspace = true`

xorgy: I think it is likely we end up releasing Fontique less often than Parley.

DJMcNab: Using the same release schedule and tagging for both is probably simpler, anyway.
Additionally, since they share a changelog, and a Fontique release requires a Parley release, mirroring that dependency probably doesn't really hurt things.

Especially since Fontique depends on Skrifa, any Parley release will realistically need a Fontique release anyway -  Skrifa releases extremely often.

PR #136, opened 2024-10-11
Implement IME primitives for PlainEditor.

xorgy:


parley/src/layout/editor.rs
```diff
@@ -21,6 +21,12 @@ pub enum ActiveText<'a> {
     Selection(&'a str),
 }
 
+#[derive(Clone, Default)]
+struct ComposeState {
+    selection: Option<(usize, usize)>,
+    text: Arc<str>,
```
nicoburns, line 27: I feel like this should be a `String`. Then:
- We could mutate it and re-use the allocation when updating the pre-edit
- Setting it to empty would be allocation free

examples/vello_editor/src/text.rs
```diff
@@ -53,6 +53,32 @@ impl Editor {
...
+                    match ime {
+                        Ime::Enabled => vec![],
+                        Ime::Disabled => vec![
+                            PlainEditorOp::SetCompose("".into(), None),
+                            PlainEditorOp::CommitCompose,
+                        ],
+                        // Winit on some platforms delivers an empty Preedit after Commit
+                        // so don't lock into compose when preedit is empty.
+                        Ime::Preedit(text, None) if text.is_empty() => vec![
+                            PlainEditorOp::SetCompose("".into(), None),
```
nicoburns, line 69: A dedicated `ClearCompose` Op for setting to empty might be nice.

nicoburns (changes requested):

parley/src/layout/editor.rs
xorgy: > * Setting it to empty would be allocation free

it may not matter, but it's never empty

examples/vello_editor/src/text.rs
xorgy: That's fair, and definitely the ergonomic choice when moving to methods instead of ops.

parley/src/layout/editor.rs
```diff
@@ -148,28 +162,78 @@ where
...
+                                new_insert.saturating_sub(1),
+                                Affinity::Upstream,
+                            )
+                        } else {
+                            Selection::from_index(&self.layout, new_insert, Affinity::Downstream)
+                        };
+                    }
+                    self.compose = None;
+                    layout_after = true;
+                }
```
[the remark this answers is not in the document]
xorgy: I agree, that will be the final command set as it has been requested by both you and Nico, and it clearly makes sense.

examples/vello_editor/src/main.rs
```diff
@@ -114,6 +116,14 @@ impl ApplicationHandler for SimpleVelloApp<'_> {
...
 
         self.editor.handle_event(event.clone());
+        if let Some(parley::Rect { x0, y0, x1, y1 }) = self.editor.preedit_area() {
+            println!("{x0} : {y0}");
+            println!("{x1} : {y1}");
+            render_state.window.set_ime_cursor_area(
+                LogicalPosition::new(x0, y0),
+                LogicalSize::new(x1 - x0, y1 - y0),
+            );
+        }
```
tomcur, line 126: There's discussion in #111 about when to notify the platform about the IME preedit area. @DJMcNab reported an infinite loop if the area is sent as a response to all IME events: https://github.com/linebender/parley/pull/111#issuecomment-2358707457. The simplest solution is to cache the previously sent area, and only send the current area if it's not the same. Unconditionally send the area on `IME::Enabled`.

To be a bit more conservative in reporting to the platform, the main thing missing is to easily know whether `Layout` has changed in response to an event, see also https://github.com/linebender/parley/pull/111#discussion_r1808975010.

xorgy: Almost all events affect either the cursor position or the Layout, so in the case of the example it doesn't much matter; however it might be good to make the calculation one-shot, maybe give it an observer pattern for ease of use.

tomcur: That is true, though some solution is required for the infinite loop that currently probably happens on some platforms (report area -> platform resends preedit -> report area -> …). Unfortunately I can't test this on my platform. `WindowEvent::{RedrawRequested,CursorMoved}` are possible exeptions.

xorgy: @tomcur I think #143 should make this pretty easy.

xorgy:
This will be updated on the new method interface, and again once #170 enters.

PR #137, opened 2024-10-11
Make `PlainEditor` width `Option<f32>`

nicoburns:
This matches `Layout`, and is useful for single-line text inputs that shouldn't ever wrap.


waywardmonkeys (approved):
Wonder if there should be tests to make sure the desired behavior happens. Probably should be on `Layout` though if we don't already have them.

PR #140, opened 2024-10-19
vello_editor: Add basic Touch handling.

xorgy:


waywardmonkeys:
With my other PR and a local addition to the `Cargo.toml` for configuring `cargo-bundle`, I was able to run and move the cursor around and do selection.

Interestingly, keyboard input doesn't work yet and just beeps at me.

But this PR worked for the thing it aims to solve.

waywardmonkeys (approved):

PR #143, opened 2024-10-22
PlainEditor: add generation counter for lazy drawing.

xorgy:


xorgy:
This makes it easier to prevent IME positioning redraw loops, and can reduce excessive drawing (when nothing visible has changed despite a command, for example extending a selection to a mouse point which produces the same cursor point and affinity).

examples/vello_editor/src/main.rs
```diff
@@ -32,23 +32,26 @@ enum RenderState<'s> {
...
     renderers: Vec<Option<Renderer>>,
 
-    // State for our example where we store the winit Window and the wgpu Surface
+    /// State for our example where we store the winit Window and the wgpu Surface
     state: RenderState<'s>,
 
-    // A vello Scene which is a data structure which allows one to build up a
-    // description a scene to be drawn (with paths, fills, images, text, etc)
-    // which is then passed to a renderer for rendering
+    /// A vello Scene which is a data structure which allows one to build up a
```
waywardmonkeys, line 45: Also need backticks on Scene here.

examples/vello_editor/src/main.rs
```diff
@@ -32,23 +32,26 @@ enum RenderState<'s> {
 }
 
 struct SimpleVelloApp<'s> {
-    // The vello RenderContext which is a global context that lasts for the
-    // lifetime of the application
+    /// The vello `RenderContext` which is a global context that lasts for the
+    /// lifetime of the application
```
waywardmonkeys, line 36: All of these lines are missing a period at the end of the sentences.

examples/vello_editor/src/main.rs
```diff
@@ -137,56 +141,61 @@ impl ApplicationHandler for SimpleVelloApp<'_> {
                         GenericFamily::SystemUi.into(),
                     ])),
                 ]);
+                render_state.window.request_redraw();
```
waywardmonkeys, line 144: Does moving this here change anything or just makes it clear it happens after any state update?

parley/src/layout/editor.rs
```diff
@@ -442,6 +464,18 @@ where
         self.buffer.clone().into()
     }
 
+    /// Get the current generation of the layout, to decide whether to draw
+    pub fn generation(&self) -> usize {
+        self.generation
+    }
+
+    /// Update the layout if it is dirty
```
waywardmonkeys, line 472: Clean feels like an odd word to me here

waywardmonkeys (commented):
Some initial comments…

parley/src/layout/editor.rs
xorgy: I could use the word `refresh` like we do on `Selection`, if that is preferred.

examples/vello_editor/src/main.rs
```diff
@@ -32,23 +32,24 @@ enum RenderState<'s> {
...
     state: RenderState<'s>,
 
-    // A vello Scene which is a data structure which allows one to build up a
-    // description a scene to be drawn (with paths, fills, images, text, etc)
-    // which is then passed to a renderer for rendering
+    /// A `vello::Scene` where the editor layout will be drawn.
     scene: Scene,
 
-    // Our text state object
+    /// Our our editor, which owns a `parley::PlainEditor`
```
DJMcNab, line 48: What's an "our editor"?

examples/vello_editor/src/text.rs
```diff
@@ -285,7 +285,14 @@ impl Editor {
         }
     }
 
-    pub fn draw(&self, scene: &mut Scene) {
+    /// Return the current generation of the layout
+    pub fn generation(&self) -> usize {
```
DJMcNab, line 289: usize?

parley/src/layout/editor.rs
```diff
@@ -49,7 +51,9 @@ where
             selection: Default::default(),
             cursor_mode: Default::default(),
             width: Default::default(),
-            scale: Default::default(),
+            scale: 1.0,
```
DJMcNab, line 54: :laughing:

parley/src/layout/editor.rs
```diff
@@ -49,7 +51,9 @@ where
...
+            dirty_layout: Default::default(),
```
DJMcNab, line 55: I'd probably call this `layout_dirty`; this way around reads a bit strangely to me.

parley/src/layout/editor.rs
```diff
@@ -34,6 +34,8 @@ where
     cursor_mode: VisualMode,
     width: Option<f32>,
     scale: f32,
+    dirty_layout: bool,
```
DJMcNab, line 37: Some docs here would be good; explaining why this is needed etc.

parley/src/layout/editor.rs
```diff
@@ -442,6 +464,18 @@ where
         self.buffer.clone().into()
     }
 
+    /// Get the current generation of the layout, to decide whether to draw
+    pub fn generation(&self) -> usize {
```
DJMcNab, line 468: Can we make this an opaque value which is only `PartialEq` (with an `as_raw` method).

Users would probably be tempted to perform a `>` check, but that's invalid (presuming that we decide performing more than $$2^{32}$$ operations is allowed)

parley/src/layout/editor.rs
```diff
@@ -452,5 +486,7 @@ where
         self.layout.break_all_lines(self.width);
         self.layout.align(self.width, Alignment::Start);
         self.selection = self.selection.refresh(&self.layout);
+        self.dirty_layout = false;
+        self.generation += 1;
```
DJMcNab, line 490:
```suggestion
        // Overflow handling: the generations should be compared only by value, so wrapping is fine.
        // This could break if the check happens to be performed exactly `2^32` generations later (2^64
        // on 64 bit), but that's sufficiently unlikely in practise.
        self.generation = self.generation.wrapping_add(1);
```
[`usize::wrapping_add`](https://doc.rust-lang.org/stable/std/primitive.usize.html#method.wrapping_add)

DJMcNab (commented):
I think that this mechanism should work.

parley/src/layout/editor.rs
```diff
@@ -402,6 +414,16 @@ where
         };
     }
 
+    /// Update the selection, and bump the generation if something other than `h_pos` changed
+    fn set_selection(&mut self, new_sel: Selection) {
+        if new_sel.focus() != self.selection.focus() || new_sel.anchor() != self.selection.anchor()
+        {
+            self.generation += 1;
```
tomcur, line 421: Also `wrapping_add` here.

tomcur (approved):
Lgtm, I like that you also included conservatively drawing. I second Daniel's suggestions to use `wrapping_add` and an opaque type (if you want to minimax, maybe `generation` can then become `u32` to save four bytes on 64 bit).

examples/vello_editor/src/main.rs
```diff
@@ -87,7 +88,7 @@ impl ApplicationHandler for SimpleVelloApp<'_> {
         // Save the Window and Surface to a state variable
         self.state = RenderState::Active(ActiveRenderState { window, surface });
 
-        event_loop.set_control_flow(ControlFlow::Poll);
+        event_loop.set_control_flow(ControlFlow::Wait);
```
tomcur, line 91: Is this here intentionally? It's also part of #144.

parley/src/layout/editor.rs
xorgy: Fair enough, sounds like a verb.

parley/src/layout/editor.rs
xorgy: You're right, some people are bonkers. :+ )

examples/vello_editor/src/main.rs
xorgy: `Poll` is excessive in this demo, `Wait` is appropriate.

examples/vello_editor/src/main.rs
xorgy: Alien technology.

parley/src/layout/editor.rs
```diff
@@ -21,6 +21,21 @@ pub enum ActiveText<'a> {
     Selection(&'a str),
 }
 
+/// Opaque representation of a generation
+// Overflow handling: the generations are only compared,
+// so wrapping is usually fine, except that in the unlikely
+// event that exactly `usize::MAX` generations happen before the
+// application gets around to drawing `Generation(0)`.
+#[derive(PartialEq, Default, Clone, Copy)]
+pub struct Generation(usize);
```
waywardmonkeys, line 30: I think a `u32` is fine here.

parley/src/layout/editor.rs
```diff
@@ -21,6 +21,21 @@ pub enum ActiveText<'a> {
...
+
+impl Generation {
+    /// Make it different from what it currently is
```
waywardmonkeys, line 33: Missing period at EOL

parley/src/layout/editor.rs
```diff
@@ -21,6 +21,21 @@ pub enum ActiveText<'a> {
...
+    pub fn differentiate(&mut self) {
```
waywardmonkeys, line 34: I don't like this name since it makes me think of other (mathematical) things ... but I can see why you don't want to call it `increment` (although plenty of other generation counters do so).

examples/vello_editor/src/text.rs
```diff
@@ -285,7 +286,14 @@ impl Editor {
...
     }
 
-    pub fn draw(&self, scene: &mut Scene) {
+    /// Return the current generation of the layout
+    pub fn generation(&self) -> Generation {
+        self.editor.generation()
+    }
+
+    /// Draw into scene
+    /// Returns drawn generation
```
waywardmonkeys, line 295: Missing periods. Needs a blank doc comment line between these lines.

examples/vello_editor/src/text.rs
```diff
@@ -285,7 +286,14 @@ impl Editor {
         }
     }
 
-    pub fn draw(&self, scene: &mut Scene) {
+    /// Return the current generation of the layout
```
waywardmonkeys, line 289: Missing period.

parley/src/layout/editor.rs
xorgy: shhhh, it's an improvement :+ )

parley/src/layout/editor.rs
DJMcNab: Agreed; I previously asked for clarification about the use of `usize` here.

parley/src/layout/editor.rs
xorgy: It's called `nudge` now. :+ )

DJMcNab: I think semantically it's a bit messy having this on this type, as end-users aren't really expected to use this.

xorgy:
I've gone ahead and added the punctuation that Bruce always wanted. :+ )

parley/src/layout/editor.rs
```diff
@@ -15,12 +15,27 @@ use alloc::{sync::Arc, vec::Vec};
...
 pub enum ActiveText<'a> {
-    /// The selection is empty and the cursor is a caret; this is the text of the cluster it is on
+    /// The selection is empty and the cursor is a caret; this is the text of the cluster it is on.
     FocusedCluster(Affinity, &'a str),
-    /// The selection contains this text
+    /// The selection contains this text.
     Selection(&'a str),
 }
 
+/// Opaque representation of a generation.
```
DJMcNab, line 24:
```suggestion
/// Opaque representation of a generation.
///
/// Obtained from [`PlainEditor::generation`].
```

parley/src/layout/editor.rs
```diff
@@ -432,16 +474,28 @@ where
...
         self.layout.lines()
     }
 
-    /// Get a copy of the text content of the buffer
+    /// Get a copy of the text content of the buffer.
     pub fn text(&self) -> Arc<str> {
         self.buffer.clone().into()
     }
 
+    /// Get the current `Generation` of the layout, to decide whether to draw.
```
DJMcNab, line 487:
```suggestion
    /// Get the current `Generation` of the layout, to decide whether to draw.
    ///
    /// You should store the generation the editor was at when you last drew it, and then redraw
    /// when the generation is different (`Generation` is [`PartialEq`], so supports the equality `==` operation).
```

parley/src/layout/editor.rs
```diff
@@ -15,12 +15,27 @@ use alloc::{sync::Arc, vec::Vec};
...
+    /// The selection contains this text.
     Selection(&'a str),
 }
 
+/// Opaque representation of a generation.
+// Overflow handling: the generations are only compared,
+// so wrapping is usually fine, except that in the unlikely
+// event that exactly `u32::MAX` generations happen before the
+// application gets around to drawing `Generation(0)`.
+#[derive(PartialEq, Default, Clone, Copy)]
```
DJMcNab, line 29:
```suggestion
#[derive(PartialEq, Eq, Default, Clone, Copy)]
```

parley/src/layout/editor.rs
```diff
@@ -49,122 +71,122 @@ where
             selection: Default::default(),
             cursor_mode: Default::default(),
             width: Default::default(),
-            scale: Default::default(),
+            scale: 1.0,
+            layout_dirty: Default::default(),
+            generation: Default::default(),
```
DJMcNab, line 76:
```suggestion
            // We don't use the `default` value to start with, as our consumers
            // will choose to use that as their initial value, but will probably need
            // to redraw if they haven't already.
            generation: Generation(1),
```

parley/src/layout/editor.rs
```diff
@@ -15,12 +15,27 @@ use alloc::{sync::Arc, vec::Vec};
...
-    /// The selection contains this text
+    /// The selection contains this text.
     Selection(&'a str),
 }
 
+/// Opaque representation of a generation.
+// Overflow handling: the generations are only compared,
+// so wrapping is usually fine, except that in the unlikely
+// event that exactly `u32::MAX` generations happen before the
+// application gets around to drawing `Generation(0)`.
```
DJMcNab, line 28:
```suggestion
// so wrapping is fine. This could only fail if exactly
// `u32::MAX` generations happen between drawing
// operations. This is implausible and so can be ignored.
```

DJMcNab (approved):
Some more peripheral tweaks, but overall this is looking really sensible.

parley/src/layout/editor.rs
```diff
@@ -15,12 +15,27 @@ use alloc::{sync::Arc, vec::Vec};
...
+#[derive(PartialEq, Default, Clone, Copy)]
+pub struct Generation(u32);
```
tomcur: A minor detail, but maybe users should not be able to construct `Generation` themselves. OTOH, that makes constructing wrapping structs a bit more cumbersome (e.g., for vello_editor, `PlainEditor` then has to be assigned to a variable before constructing `SimpleVelloApp`). At the same time, currently the editor's first generation and the default generation are 0, that may lead to bugs in user code.

Users can get easy constructing again by using `last_generation: Option<Generation>`, but then generation should ideally wrap [NonZeroU32](https://doc.rust-lang.org/stable/std/num/type.NonZeroU32.html).

examples/vello_editor/src/main.rs
```diff
@@ -201,6 +208,7 @@ fn main() -> Result<()> {
         state: RenderState::Suspended(None),
         scene: Scene::new(),
         editor: text::Editor::default(),
+        last_drawn_generation: Default::default(),
```
waywardmonkeys, line 211: SUGGESTION!

Maybe `last_drawn_generation: Generation::ZERO` and then where Daniel suggested `Generation(1)`, you'd use `Generation::INITIAL` and those are 2 pub consts on `Generation` that can be documented so that the usage and intent is clear.

But I'd get someone else to say that makes sense before you take it too seriously.

parley/src/layout/editor.rs
```diff
@@ -49,122 +73,125 @@ where
             selection: Default::default(),
             cursor_mode: Default::default(),
             width: Default::default(),
-            scale: Default::default(),
+            scale: 1.0,
+            layout_dirty: Default::default(),
+            // We don't use the `default` value to start with, as our consumers
+            // will choose to use that as their initial value, but will probably need
+            // to redraw if they haven't already.
+            generation: Generation(1),
```
tomcur, line 81: With this added `Generation::default` is fine to keep. The choice is purely a semantic one now.

tomcur (approved):

examples/vello_editor/src/main.rs
xorgy: I kinda like the idea, but also I think that `Default` is fine here, and `Generation(1)` is fine in editor.rs because (for now) it lives in the editor module, has no other users, and is generally not intended to be constructed anywhere else.

parley/src/layout/editor.rs
waywardmonkeys: I kind of like this. But comparing `Option<Generation>` and `Generation` is kind of ugh-ly.

waywardmonkeys (approved):

PR #144, opened 2024-10-22
vello_editor: Use PresentMode::Mailbox when available.

xorgy:
AutoVsync/FIFO introduces unnecessary latency, especially on X11. There aren't animations in vello_editor so the downside (extra work) is not significant.


xorgy:
@tomcur  that is certainly odd! `AutoVsync` should fall back to FIFO, which should be supported on all platforms including Wayland... 
As for `AutoNoVsync`, it's possible it selected `Mailbox` on your system.

dfrg (approved):
This LGTM. Fixes intolerable latency on Windows as well.

PR #145, opened 2024-10-23
vello_editor: Always paint on RedrawRequested.

xorgy:
In non-composited environments, damage accumulates if we don't present.

### DIFF
--- a/examples/vello_editor/src/main.rs
+++ b/examples/vello_editor/src/main.rs
@@ -144,56 +144,53 @@ impl ApplicationHandler for SimpleVelloApp<'_> {
 
             // This is where all the rendering happens
             WindowEvent::RedrawRequested => {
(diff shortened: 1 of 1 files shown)


xorgy:
This is, in part, a fix for a regression caused by #143 on non-composited environments (e.g. uncomposited X11, sometimes Windows).

xorgy:
The efficient way to handle this is probably to do a pass with `LoadOp::Load` or some such, since we know the configuration hasn't changed, or somehow to present the old texture again.

waywardmonkeys (approved):

xorgy:
I gave a first try at writing a render pass that would load the old image and do nothing, but the old image is black on my platform, I suspect this is not a real way to do this. Preventing redraw might involve copying the surface texture.

PR #146, opened 2024-10-30
PlainEditor: Add byte selection and navigation operations.

xorgy:


xorgy:
Splitting out cleanup changes in #147 and #148

waywardmonkeys (approved):
Basically a rubber stamp as I don't know what exactly each editor operation should be doing and there's no tests.

Here's hoping that Aaron (or someone) can add tests in a future PR!

xorgy:
I promise that if you file an issue, I will write tests at some point for PlainEditor.

mwcampbell:
A bit late, but LGTM.

parley/src/layout/editor.rs
```diff
@@ -111,6 +111,10 @@ where
     BackdeleteWord,
     /// Move the cursor to the cluster boundary nearest this point in the layout.
     MoveToPoint(f32, f32),
+    /// Move the cursor to a byte index.
+    ///
+    /// No-op if index is not a char boundary.
+    MoveToByte(usize),
```
DJMcNab, line 117: For all of these, we should document the affinity semantics (in that it clobbers them entirely).

parley/src/layout/editor.rs
```diff
@@ -424,20 +441,7 @@ where
...
-                Affinity::Upstream,
-            )
-        } else {
-            Selection::from_index(
-                &self.layout,
-                new_start.min(self.buffer.len()),
-                Affinity::Downstream,
-            )
-        };
+        self.set_selection(self.cursor_at(start.saturating_add(s.len())).into());
```
DJMcNab, line 444: I'm not certain this handles affinity as I'd expect. I'd expect the cursor to be associated with the content you just pasted, not the following cluster.

mwcampbell:
I forgot about affinity semantics. With AccessKit, the client (e.g. screen reader) can theoretically specify the affinity at both ends when setting the selection, though I don't think any of the platform accessibility APIs actually allow that.

xorgy:
@DJMcNab  @mwcampbell Yeah, this is a bit of a mess right now with how `Cursor` currently works, I intend to add some fns to `Cursor` that make it easier to model ‘insertion’ and other behaviors, but the fundamental issue is that Cursors can't be constructed where `index` = `buffer.len()`, which makes all of these behaviors highly conditional.

xorgy:
If we make a `Cursor` type that is based on an insertion index first, like the Parley 0.1 `Cursor`, then all these behaviors become much easier (because we can just always have `Upstream` affinity when inserting, and the neutral empty cursor at `index` = 0 suddenly makes sense)

PR #148, opened 2024-10-30
Remove trailing whitespace in cursor.rs

xorgy:


waywardmonkeys (commented):
You'll hate me a bit more, but could you do this for the other files in the repo that have whitespace at EOL?

* doc/concept.md
* doc/design.md
* fontique/src/script.rs
* parley/src/layout/cursor.rs
* parley/src/layout/line/mod.rs
* parley/src/swash_convert.rs

DJMcNab (commented):
I suspect these indicate that rustfmt formatting is failing for some of these files. At least for `script.rs`. It might be worth double checking some of them to make sure (by inserting definitely wrong formatting and seeing what happens)

Maybe `cursor.rs` is fine, because the trailing whitespace is in doc comments?

PR #152, opened 2024-10-31
Implement cursor blinking in text editor

tannal:
Implemented some kind of timer using `ControlFlow::WaitUntil`.
I believe the timer thing can be handled in a more general way, but I haven't figured it out yet. 
[blink.webm](https://github.com/user-attachments/assets/c28b359e-86b2-4186-8d0d-5c13745a0907)

### DIFF
--- a/examples/vello_editor/src/main.rs
+++ b/examples/vello_editor/src/main.rs
@@ -119,7 +119,7 @@ impl ApplicationHandler for SimpleVelloApp<'_> {
     fn new_events(&mut self, event_loop: &ActiveEventLoop, cause: StartCause) {
         match cause {
             StartCause::Init => {
-                self.editor.init();
+                self.editor.cursor_reset();
                 if let Some(next_time) = self.editor.next_blink_time() {
                     event_loop.set_control_flow(ControlFlow::WaitUntil(next_time));
                 }
--- a/examples/vello_editor/src/main.rs
+++ b/examples/vello_editor/src/main.rs
@@ -4,6 +4,7 @@
 use anyhow::Result;
 use std::num::NonZeroUsize;
 use std::sync::Arc;
+use std::time::{Duration, Instant};
 use vello::peniko::Color;
 use vello::util::{RenderContext, RenderSurface};
 use vello::wgpu;
@@ -50,6 +51,8 @@ struct SimpleVelloApp<'s> {
 
     /// The last generation of the editor layout that we drew.
     last_drawn_generation: text::Generation,
+
+    last_blink_time: Instant
 }
 
 impl ApplicationHandler for SimpleVelloApp<'_> {
@@ -106,7 +109,10 @@ impl ApplicationHandler for SimpleVelloApp<'_> {
         // Save the Window and Surface to a state variable
         self.state = RenderState::Active(ActiveRenderState { window, surface });
 
-        event_loop.set_control_flow(ControlFlow::Wait);
+        let now = Instant::now();
+        let next_blink_time = now + Duration::from_secs_f32(0.5);
+        self.last_blink_time = now;
+        event_loop.set_control_flow(ControlFlow::WaitUntil(next_blink_time));
     }
 
     fn suspended(&mut self, event_loop: &ActiveEventLoop) {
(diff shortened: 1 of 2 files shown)


examples/vello_editor/src/main.rs
```diff
@@ -116,6 +122,31 @@ impl ApplicationHandler for SimpleVelloApp<'_> {
...
+                self.last_blink_time = now;
+                let next_blink_time = now + Duration::from_secs_f32(0.5);
+                event_loop.set_control_flow(ControlFlow::WaitUntil(next_blink_time));
+                
+                if let RenderState::Active(state) = &self.state {
+                    state.window.request_redraw();
+                }
+            }
+            _ => {}
+        }
```
xorgy, line 148: Needs to handle [`WaitCancelled`](<https://docs.rs/winit/latest/winit/event/enum.StartCause.html#variant.WaitCancelled>).

examples/vello_editor/src/text.rs
```diff
@@ -42,6 +42,10 @@ impl Editor {
         self.editor.text()
     }
 
+    pub fn cursor_blink(&mut self) {
+        self.cursor_blink = !self.cursor_blink;
```
xorgy, line 46: Might be better to handle this with a periodic function, instead of a toggle, but I won't block on this.

examples/vello_editor/src/text.rs
```diff
@@ -300,10 +304,10 @@ impl Editor {
             scene.fill(Fill::NonZero, transform, Color::STEEL_BLUE, None, &rect);
         }
         if let Some(cursor) = self.editor.selection_strong_geometry(1.5) {
-            scene.fill(Fill::NonZero, transform, Color::WHITE, None, &cursor);
+            scene.fill(Fill::NonZero, transform, if self.cursor_blink {Color::TRANSPARENT} else {Color::WHITE}, None, &cursor);
```
xorgy, line 307: Since we're painting the whole scene regardless, might as well just add the condition around both of these `if let Some`s.

xorgy (changes requested):

xorgy:
And overall, remember to run `cargo fmt` on your code before committing it. It will fail CI if it is not formatted the way the machine wants.

examples/vello_editor/src/text.rs
tannal: Do you mean something like this?
```rs
pub fn cursor_blink(&mut self) {
    if let Some(start_time) = self.start_time {
        let elapsed = Instant::now().duration_since(start_time);
        self.cursor_blink = (elapsed.as_millis() / self.blink_period.as_millis()) % 2 == 0;
    }
}
```

examples/vello_editor/src/text.rs
tannal: Done.

examples/vello_editor/src/main.rs
tannal: Done.

examples/vello_editor/src/text.rs
```diff
@@ -39,6 +41,27 @@ impl Editor {
         self.editor.text()
     }
 
+    pub fn init(&mut self) {
+        self.start_time = Some(Instant::now());
+        self.blink_period = Duration::from_millis(500);
```
DJMcNab, line 46: We should have a comment here that for simplicity we don't get the system blink period, but in real systems you *must* do so.

We can probably link to [#glazier > Cursor Blink Interval](https://xi.zulipchat.com/#narrow/channel/351333-glazier/topic/Cursor.20Blink.20interval)

examples/vello_editor/src/text.rs
```diff
@@ -39,6 +41,27 @@ impl Editor {
         self.editor.text()
     }
 
+    pub fn init(&mut self) {
+        self.start_time = Some(Instant::now());
```
DJMcNab, line 45: I think that the blinking should be reset if new text is added. Note how in e.g. Firefox the cursor stays a solid line whilst you are inputting text. It seems like it just toggles the cursor on, so the next blink interval would turn it off again? It's slightly unclear what the correct behaviour is, but just blinking all the time probably isn't it.

DJMcNab (commented):
I've not run this myself yet, but this will be good functionality to land. We could also point to [#gpu > Rendering a blinking cursor](https://xi.zulipchat.com/#narrow/channel/197075-gpu/topic/Rendering.20a.20blinking.20cursor)

examples/vello_editor/src/text.rs
tannal: I agree with that the blinking should be reset if the user are typing.
Thanks for the feedback!

examples/vello_editor/src/text.rs
xorgy: Yeah something like that, this also makes it easier to implement solid cursor while actively typing, and to use the timers for more things (I want to add long-touch word selection).

xorgy (approved):
I'm happy with this as it is. Thank you for your contribution @tannal !

tannal:
That's cool.
And thanks for all the feedback!

PR #153, opened 2024-10-31
Update to `hashbrown` 0.15

waywardmonkeys:
Among other things, this changes the default hash from `ahash` to `foldhash`.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -716,6 +716,12 @@ dependencies = [
 ]
 
 [[package]]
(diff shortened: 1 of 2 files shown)


DJMcNab (approved):
Is there a specific reason we want this, or just as a general upgrade?

waywardmonkeys:
> Is there a specific reason we want this, or just as a general upgrade?

Keeping up with the Jones.

PR #154, opened 2024-11-01
Eliminate extraneous allocation and dynamic dispatch in `PlainEditor`

mwcampbell:
I started by wanting to eliminate `Arc<str>`. Then I noticed that `vello_editor` was allocating temporary `Vec`s for the `PlainEditorOp`s. Finally, it occurred to me that we could probably do something altogether more efficient than `PlainEditorOp`. I think my refactored version, with separate functions in `PlainEditorTxn`, will be nicer to maintain as well.

### DIFF
--- a/examples/vello_editor/src/main.rs
+++ b/examples/vello_editor/src/main.rs
@@ -16,7 +16,7 @@ use winit::window::Window;
 
 // #[path = "text2.rs"]
 mod text;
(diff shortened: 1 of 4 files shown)


xorgy:
Thanks, this looks great. I'll give it a quick check.

xorgy (approved):

PR #160, opened 2024-11-05
Get as close as possible to full no_std support

mwcampbell:
To really get no-std support in parley, we need to land dfrg/swash#63. But I believe I've done everything else that's necessary, in both parley and fontique. One can test the build by patching Cargo.toml to use the working branch for that swash PR, and modifying parley/Cargo.toml to specify swash/std or swash/libm as appropriate. If we decide to land dfrg/swash#63 before landing this one, then I can make the latter modification to parley/Cargo.toml before merging this PR.

Some of my changes, e.g. unconditionally using `alloc::vec::Vec`, are designed to minimize the amount of noise when searching the whole source tree for "std".

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1815,6 +1815,7 @@ dependencies = [
 name = "parley"
 version = "0.2.0"
 dependencies = [
+ "core_maths",
  "fontique",
  "peniko",
  "skrifa",
--- a/fontique/src/attributes.rs
+++ b/fontique/src/attributes.rs
@@ -3,7 +3,8 @@
 
 //! Properties for specifying font weight, stretch and style.
 
-#[cfg(all(feature = "libm", not(test)))]
+#[cfg(feature = "libm")]
+#[allow(unused_imports)]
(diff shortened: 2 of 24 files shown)


xorgy (approved):

mwcampbell:
I saw @xorgy's approval. The only reason I'm not merging this immediately is that I wonder if we should land dfrg/swash#63 first, and publish a new version of swash (presumably a semver-breaking release), so we can go all the way with this PR. If people think it would be better to merge this PR immediately, that's fine.

waywardmonkeys:
https://github.com/dfrg/swash/pull/63 and https://github.com/dfrg/swash/pull/61 have both landed now. I have submitted a couple of additional PRs for swash and I don't know what his plans are for doing a new release.

xStrom:
I think it's fine to land this and when `swash` gets released do those steps in a separate PR.

Merged PRs are better than lingering PRs. When work can be split, it should. Helps reduce merge conflicts for yourself and others. Keeps the ball rolling.

PR #162, opened 2024-11-07
Fontique: Use `read-fonts` instead of `skrifa`

nicoburns:
Fontique was previously depending on `skrifa` only to use it's re-export of `read-fonts`. This PR makes `fontique` depend on `read-fonts` directly.

I also bumped the versions of `skrifa` (parley still uses `skrifa`) and `read-fonts` to the latest versions.


waywardmonkeys (commented):
I've asked about this usage before as well and I think it is okay to change `fontique` in this way.

But we don't want to update the version of fontations here until it is also updated in `swash` and that's published to avoid duplicated dependencies (as you can see happening in the changes in `Cargo.toml`).

I've already got the changes to update fontations in `swash` (along with the changes for improved `no_std` support), so it shouldn't be long before we can publish that and update it here.

If you can remove that part from here and do the minimal change, then I think we can approve that.

nicoburns:
> If you can remove that part from here and do the minimal change, then I think we can approve that.

@waywardmonkeys done.

waywardmonkeys (approved):

PR #164, opened 2024-11-08
Remove broken line break condition in right-side handling in `Cursor::previous_visual`

mwcampbell:
That code was clearly adapted from the left-side handling in `Cursor::next_visual`, but it doesn't seem to belong here. It caused this bug: In vello_editor, move down to the first line of the second paragraph (after the blank line), arrow right once, then arrow left. That should leave the cursor back at the start of that line, but instead it jumps back to the blank line.


dfrg:
I believe this was superseded by #170

mwcampbell:
Yes, it was.

PR #166, opened 2024-11-09
Add optional AccessKit integration for text layout and `PlainEditor`

mwcampbell:
This is an optional feature in Parley. `vello_editor` demonstrates how to use it.

### DIFF
--- a/examples/vello_editor/src/main.rs
+++ b/examples/vello_editor/src/main.rs
@@ -1,6 +1,7 @@
 // Copyright 2024 the Parley Authors
 // SPDX-License-Identifier: Apache-2.0 OR MIT
 
+use accesskit::{Node, Rect, Role, Tree, TreeUpdate};
 use anyhow::Result;
 use std::num::NonZeroUsize;
 use std::sync::Arc;
@@ -11,18 +12,55 @@ use vello::{AaConfig, Renderer, RendererOptions, Scene};
 use winit::application::ApplicationHandler;
 use winit::dpi::LogicalSize;
 use winit::event::*;
-use winit::event_loop::{ActiveEventLoop, ControlFlow, EventLoop};
+use winit::event_loop::{ActiveEventLoop, ControlFlow, EventLoop, EventLoopProxy};
 use winit::window::Window;
+
+mod access_ids;
+use access_ids::{TEXT_INPUT_ID, WINDOW_ID};
 
 // #[path = "text2.rs"]
 mod text;
 use parley::{GenericFamily, StyleProperty};
 
+const WINDOW_TITLE: &str = "Vello Text Editor";
+
 // Simple struct to hold the state of the renderer
 pub struct ActiveRenderState<'s> {
-    // The fields MUST be in this order, so that the surface is dropped before the window
+    // The fields MUST be in this order, so that the surface and AccessKit adapter are dropped before the window
     surface: RenderSurface<'s>,
+    access_adapter: accesskit_winit::Adapter,
     window: Arc<Window>,
+    sent_initial_access_update: bool,
+}
+
+impl ActiveRenderState<'_> {
+    fn access_update(&mut self, editor: &mut text::Editor) {
+        self.access_adapter.update_if_active(|| {
+            let mut update = TreeUpdate {
+                nodes: vec![],
+                tree: (!self.sent_initial_access_update).then(|| Tree::new(WINDOW_ID)),
+                focus: TEXT_INPUT_ID,
+            };
+            if !self.sent_initial_access_update {
+                let mut node = Node::new(Role::Window);
+                node.set_label(WINDOW_TITLE);
+                node.push_child(TEXT_INPUT_ID);
+                update.nodes.push((WINDOW_ID, node));
+                self.sent_initial_access_update = true;
+            }
+            let mut node = Node::new(Role::TextInput);
+            let size = self.window.inner_size();
+            node.set_bounds(Rect {
+                x0: 0.0,
+                y0: 0.0,
+                x1: size.width as _,
+                y1: size.height as _,
+            });
+            editor.accessibility(&mut update, &mut node);
+            update.nodes.push((TEXT_INPUT_ID, node));
+            update
+        });
+    }
 }
 
 enum RenderState<'s> {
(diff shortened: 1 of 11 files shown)


mwcampbell:
I'm aware that this may soon be invalidated by @dfrg's planned change from character indices to gap indices. But I thought I should still get this PR open first, since I have something that more or less works.

The part I'm least sure about is how to expose geometry, particularly the bounding rectangles for the runs. I'm not sure if I'm doing the right thing for layouts where the dominant direction is RTL. But then, I'm also not sure if vello_editor's code for rendering the selection and cursor works in that case.

Another thing I need confirmation on: Does `Line::runs` always iterate over runs in textual order?

examples/vello_editor/src/main.rs
```diff
@@ -11,18 +12,55 @@ use vello::{AaConfig, Renderer, RendererOptions, Scene};
...
 // #[path = "text2.rs"]
 mod text;
 use parley::{GenericFamily, StyleProperty};
 
+const WINDOW_TITLE: &str = "Vello Text Editor";
+
 // Simple struct to hold the state of the renderer
 pub struct ActiveRenderState<'s> {
-    // The fields MUST be in this order, so that the surface is dropped before the window
+    // The fields MUST be in this order, so that the surface and AccessKit adapter are dropped before the window
```
DJMcNab, line 29: I don't think this is true for wgpu any more, since the `Arc<Window>` is also owned by the `wgpu` Surface.

This might still be correct for accesskit, however. I'd have to dig into it.

examples/vello_editor/src/text.rs
```diff
@@ -327,6 +340,11 @@ impl Editor {
         }
         self.editor.generation()
     }
+
+    pub fn accessibility(&mut self, update: &mut TreeUpdate, node: &mut Node) {
+        self.editor
+            .accessibility(update, node, next_node_id, INSET as _, INSET as _);
```
DJMcNab, line 346: Can we use `.into()` or `try_into().unwrap()` instead of `as _` here?

parley/src/layout/cursor.rs
```diff
@@ -373,6 +395,27 @@ impl Cursor {
...
+    #[cfg(feature = "accesskit")]
+    pub fn to_access_position<B: Brush>(
+        &self,
+        layout: &Layout<B>,
+        layout_access: &LayoutAccessibility,
+    ) -> Option<TextPosition> {
+        let run_path = (self.path.line_index(), self.path.run_index());
+        let id = layout_access.access_ids_by_run_path.get(&run_path)?;
+        let mut character_index = self.path.logical_index();
+        if self.affinity == Affinity::Upstream {
```
DJMcNab, line 408: This could probably do with a comment. I'm finding it hard to reason about what this is doing.

Part of that is how confusing Upstream/Downstream are, which isn't your fault.

parley/src/layout/mod.rs
```diff
@@ -261,3 +267,149 @@ pub struct Decoration<B: Brush> {
...
+            // or the end of the line. That way, we can set relations between
+            // runs in a line and do anything special that might be required
+            // for the last run in a line.
+            let mut last_node: Option<(NodeId, Node)> = None;
+
+            // Iterate over the runs from left to right, computing their offsets,
+            // then sort them into text order.
+            let runs = {
+                let mut run_offset = metrics.offset;
+                let mut runs = Vec::with_capacity(line.len());
```
DJMcNab, line 311: Should we reuse this vector? Maybe not between calls, but at least between loop iterations.

parley/src/layout/mod.rs
```diff
@@ -261,3 +267,149 @@ pub struct Decoration<B: Brush> {
...
+        let mut ids_to_remove = Vec::<NodeId>::new();
+        let mut run_paths_to_remove = Vec::<(usize, usize)>::new();
+        for (access_id, run_path) in self.run_paths_by_access_id.iter() {
+            if !ids.contains(access_id) {
+                ids_to_remove.push(*access_id);
+                run_paths_to_remove.push(*run_path);
+            }
+        }
+        for id in ids_to_remove {
+            self.run_paths_by_access_id.remove(&id);
```
DJMcNab, line 409: Presumably it's a choice for simplicity to not reuse these ids?

DJMcNab (approved):
I've not tested the code, but it mostly reads as correct to me.

examples/vello_editor/src/main.rs
mwcampbell: The AccessKit adapter definitely doesn't own the window. The platform backends are supposed to deal gracefully with the adapter outliving the window, but still, the adapter _should_ be destroyed first. I'm not so sure about the surface. I'd think that if the surface owned an `Arc<Window>`, we'd see `Arc<Window>` as a parameter on `RenderSurface`, as I did with softbuffer's `Surface` when experimenting with that yesterday.

parley/src/layout/cursor.rs
mwcampbell: I hope that's clear now without being too verbose.

DJMcNab: Thank yeah, that really helps!

mwcampbell:
I've now addressed all of @DJMcNab's comments. Should I wait for anyone else in particular to review this before merging?

parley/src/layout/mod.rs
DJMcNab: Hmm. I guess that makes sense as a pragmatic choice. It's not like there's any shortage of 64 bit ids.

My concern is also that if you e.g. have:
```text
{a bit of text}[0]
{with some importance}[1]
```
(where `{}[id]` marks a run with id `id`)
and then add a newline before the word "text", you end up with:
```text
{a bit of}[0]
{text}[1]
{with some importance}[2]
```
That is, the second run has been completely transformed in content.

I really don't have a good suggestion here, unfortunately. I imagine your best chance of doing better is with something like the editor abstraction, so you know what has changed. But that's an awful lot of bookkeeping work needed.

DJMcNab:
It should probably get a once-over from @dfrg, but I think this can land even if we don't hear from him today.

examples/vello_editor/src/main.rs
DJMcNab: Internally, wgpu stores the raw window in a boxed form. But yeah, this specific comment doesn't matter too much.

parley/src/layout/mod.rs
mwcampbell: You're completely right about the effects of inserting text. I'm not sure how much this matters. Of the platform backends implemented so far, this only makes a difference at all on Windows, where UIA clients can have long-lived references to text range objects, and in our implementation, those objects point to specific run IDs. I don't know if any UIA implementation really handles this ideally.

mwcampbell: It's also worth noting that the egui implementation, which has been my de-facto reference implementation until now, has the same problem with the insertion scenario. So the egui implementation is really no better. It uses line indices as the basis of run node IDs. It just doesn't have an explicit mapping, since it generates AccessKit node IDs via hashing.

xorgy (approved):

mwcampbell:
I got two approvals, but I'm still inclined to wait and see if @dfrg, who seemingly has the most expertise about text and parley specifically, can look at this soon. But if anyone else thinks this should be landed immediately, I can do that.

dfrg:
The code generally LGTM. I think there’s room for improvement if we adjust some of the layout data structures to be more amenable to accesskit integration but that’s something to keep in mind for future work so I’m happy to land this in current form. Thanks!

Issue #142, opened 2024-10-21
Underline & Strikethrough handling not documented well


spirali:
Underline and strikethrough does not draw any line.

I have tried enable it just by:
```rust
    builder.push_default(StyleProperty::Underline(true));
```

or setting all properties:

```rust
    builder.push_default(StyleProperty::Underline(true));
    builder.push_default(StyleProperty::UnderlineBrush(Some(text_color)));
    builder.push_default(StyleProperty::UnderlineSize(Some(5.0)));
    builder.push_default(StyleProperty::UnderlineOffset(Some(1.0)));
```

in both cases, the text is drawn but without the decoration. I have both tested `main` branch and version `0.2.0`.

DJMcNab:
Parley won't emulate underline drawing for you; you need to wire it up yourself using the offset provided by [`GlyphRun::style`](https://docs.rs/parley/latest/parley/layout/struct.GlyphRun.html).

See for example the relevant code in Masonry: https://github.com/linebender/xilem/blob/771eb2f02f3d6629b78fe233fdec76a563ddbbb2/masonry/src/text/render_text.rs#L86-L119

Note that this code is for Parley 0.1.0, although I don't believe that this part has changed.

spirali:
Thank you for your explanation!

waywardmonkeys:
@spirali I think it is reasonable to keep this open and change it to indicate that we need to improve the documentation.

spirali:
Ok. Btw: I think that it would be nice to put a simple implementation into examples `swash_render` and `tiny_skia_renderer` or at least to put there a comment that the user needs to implement this part manually.

nicoburns:
> Btw: I think that it would be nice to put a simple implementation into examples swash_render and tiny_skia_renderer or at least to put there a comment that the user needs to implement this part manually.

Totally agree. I made those examples, and the only reason that this code is not included as part of them is because I also missed the fact that it was necessary!

nicoburns:
Fixed by #167

PR #167, opened 2024-11-09
Added decorations in examples  (underline & strikethrough)

spirali:
This is related to #142.


xorgy (approved):
Thank you, this works and looks fine to me.

PR #168, opened 2024-11-09
Remove duplicated code in positioned_glyphs()

spirali:
Reusing `glyphs()` method that already defines the iterator.


xorgy (approved):
Thanks!

Issue #132, opened 2024-10-10
Unexpected trailing newline behavior with `Cursor` and `Selection`


xorgy:

dfrg:
This is a bug that originates in swash. Essentially, the Unicode line breaking algorithm generates mandatory breaks _after_ the newline sequence so we don’t capture the final empty line at all because we don’t have a cluster where we can attach the state. I have a “fix” for this in the upcoming code that just uses the cluster’s whitespace property to detect newlines and ignores the results of the LBA entirely for these. This actually simplifies the code a bit and removes the nasty “skip mandatory break” logic.

nicoburns:
I think this was fixed by https://github.com/linebender/parley/pull/170. @xorgy please reopen if you disagree.

PR #170, opened 2024-11-11
Cursor rework

dfrg:
This is a draft because some functionality (mostly around word/line selection) is still missing.  I also commented out the fix from #163 because the `skip_mandatory_break` flag is now gone and I haven't dug into the logic from that PR.

Submitting this now so that people can take a look and see if it fixes some of the known bugs, particularly around newline handling.

note: #166 should probably land first


nicoburns:
> I also commented out the fix from https://github.com/linebender/parley/pull/163 because the skip_mandatory_break flag is now gone and I haven't dug into the logic from that PR.

Looks like you're now appending the newline cluster to the previous line rather than waiting to append it to the next line. We'll need to test, but I suspect that means that the bug fixed in #163 won't be present. The key thing not being which line the newline cluster ends up in, but that the cluster is "committed" eagerly rather than deferred to the next iteration of the loop (the fix in #163 was handling that deferred cluster commit handling the case that the next iteration of the loop was an inline box (and thus went down a different code path)).

If you wanted to keep the "newline is at the start of the following line" behaviour then think you could probably just reverse the `append_cluster_to_line` and `try_commit_line` statements (although IMO it's more intuitive if the newline cluster is part of the line it ends).

nicoburns (commented):
I've left some review comments on the layout code that I'm familiar with. I haven't reviewed the selection code which I don't fully understand yet.

spirali:
I did some quick manual testing with 220f1bc027c213599357f47f952d72e1b77125e9 and it seems that problem from #163 is gone. The logic behind #163 was to solve the problem that newline cluster is part of the next line and it broke placing inline boxes.

parley/src/layout/line/greedy.rs
```diff
@@ -252,50 +253,29 @@ impl<'a, B: Brush> BreakLines<'a, B> {
...
-                                // We do not currently handle breaking within a ligature, so we ignore boundaries in such a position.
-                                //
-                                // We also don't record boundaries when the advance is 0. As we do not want overflowing content to cause extra consecutive
-                                // line breaks. We should accept the overflowing fragment in that scenario.
-                                if !is_ligature_continuation && self.state.line.x != 0.0 {
-                                    self.state.mark_line_break_opportunity();
-                                    // break_opportunity = true;
-                                }
+                        } else if is_newline {
+                            self.state.append_cluster_to_line(0.0);
```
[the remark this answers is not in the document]
dfrg: I guess I assumed that this value was the advance of the next cluster but it appears to want to accumulated advance. Good catch.

parley/src/layout/line/greedy.rs
```diff
@@ -252,50 +253,29 @@ impl<'a, B: Brush> BreakLines<'a, B> {
 
                         // Retrieve metadata about the cluster
                         let is_ligature_continuation = cluster.is_ligature_continuation();
-                        let is_space = cluster.info().whitespace().is_space_or_nbsp();
+                        let whitespace = cluster.info().whitespace();
+                        let is_newline = whitespace == Whitespace::Newline;
+                        let is_space = whitespace.is_space_or_nbsp();
                         let boundary = cluster.info().boundary();
```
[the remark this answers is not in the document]
dfrg: The full explanation is at https://github.com/linebender/parley/issues/132#issuecomment-2408754466

tldr; Unicode LBA specifies a break _after_ the newline and a trailing newline doesn’t have a following cluster so we lose that information. 

Re check order: good question. I’m going to assume that a newline sequence is not marked as a break opportunity because the code appears to work as is :)

parley/src/shape.rs
```diff
@@ -38,9 +38,12 @@ pub fn shape_text<'a, B: Brush>(
     infos: &[(CharInfo, u16)],
     levels: &[u8],
     scx: &mut ShapeContext,
-    text: &str,
+    mut text: &str,
     layout: &mut Layout<B>,
 ) {
+    if text.is_empty() {
+        text = " ";
+    }
```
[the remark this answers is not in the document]
dfrg: We’ll fix this properly with the larger refactor.

dfrg:
> I did some quick manual testing with [220f1bc](https://github.com/linebender/parley/commit/220f1bc027c213599357f47f952d72e1b77125e9) and it seems that problem from #163 is gone. The logic behind #163 was to solve the problem that newline cluster is part of the next line and it broke placing inline boxes.

Thanks for confirming.

dfrg:
> If you wanted to keep the "newline is at the start of the following line" behaviour then think you could probably just reverse the append_cluster_to_line and try_commit_line statements (although IMO it's more intuitive if the newline cluster is part of the line it ends).

This was initially done to force a line to be created at the end of the layout when the last character was a newline. This didn’t really work anyway and I agree that it’s more intuitive (and correct!) to keep the newline attached to the line it ends.

parley/src/shape.rs
```diff
@@ -38,9 +38,12 @@ pub fn shape_text<'a, B: Brush>(
...
     // Do nothing if there is no text or styles (there should always be a default style)
     if text.is_empty() || styles.is_empty() {
```
DJMcNab, line 48: This first condition can now never be true, I think...

nicoburns: We should probably leave the check there anyway, seeing as the hack to make text never empty is intended to be temporary?

dfrg: It’s temporary but I don’t expect to fix this hack until we redo shaping at which point this entire module will be rewritten so 🤷🏻‍♂️. 

Looking again, this should probably avoid setting empty text to a space if there are any inline objects.

parley/src/layout/cursor.rs
```diff
@@ -1,472 +1,340 @@
...
-#[derive(Copy, Clone, PartialEq, Default, Debug)]
 pub struct Cursor {
-    path: ClusterPath,
-    index: u32,
-    text_start: u32,
-    text_end: u32,
-    visual_offset: f32,
-    is_rtl: bool,
+    index: usize,
     affinity: Affinity,
```
xorgy, line 11: There's something very familiar about this. :thinking: :thought_balloon:

mwcampbell:
`Cursor::to_access_position` returns `None` when the cursor is at the end of the text (e.g. after pressing Ctrl+End in vello_editor). This is incorrect, as it means we can't report an accessible text selection when the cursor is at the end.

dfrg:
Ah, I see. We won’t have a downstream cluster at the end of the text. I’ll push a fix for this in a bit.

mwcampbell:
Just tried the latest push with a screen reader. If I go to the end of the default text in vello_editor with Ctrl+End, then the accessible selection is where it should be. But if I then press Enter to start a new blank line at the end, then the accessible selection is on the previous line, after the hard line break, rather than on the new, blank line.

dfrg:
This is probably a question of affinity. I pushed an update that _might_ address it.

dfrg:
I wonder if that problem is that there's no actual character on that line and AccessKit doesn't have a text run node associated with it?

mwcampbell:
When inserting a blank line at the end of the text, is there a Parley run for that blank line at the end? It appears that there is now  parley run, and an AccessKit text run node, for the one blank line if the text is empty.

dfrg:
I think there's one if the text is empty but there's no run for a final blank line in non-empty text. I guess we need to add one somehow and make sure it's mapped properly in the to/from acessskit methods.

parley/src/layout/cursor.rs
```diff
@@ -97,304 +73,280 @@ impl Cursor {
...
+            if cluster.is_rtl() {
+                clusters = [None, Some(cluster)];
             } else {
-                self.affinity
-            };
-            Self::from_cluster(prev, affinity)
-        } else {
-            *self
+                clusters = [Some(cluster), None];
+            }
```
xorgy, line 332: Might be better to hoist some of these `clusters = ` assignments.

parley/src/layout/cursor.rs
```diff
@@ -550,153 +474,95 @@ impl Selection {
...
+    pub fn previous_visual_word<B: Brush>(&self, layout: &Layout<B>, extend: bool) -> Self {
+        self.maybe_extend(self.focus.previous_visual_word(layout), extend)
     }
 
     /// Returns a new selection with the focus moved to the next line. The
     /// current horizontal position will be maintained.
     ///
     /// If `extend` is `true` then the current anchor will be retained,
-    /// otherwise the new selection will be collapsed.
+    /// otherwise the new selection will be collapsed.
```
xorgy, line 565: Please remove trailing white space.

parley/src/layout/cursor.rs
```diff
@@ -785,48 +711,62 @@ impl Selection {
...
             }
+            if cur_x != start_x {
+                let width = (cur_x - start_x).max(MIN_RECT_WIDTH);
+                f(Rect::new(start_x, line_min, start_x + width, line_max));
+            }
+            // } else {
+            //     let x = metrics.offset as f64;
+            //     let width = (metrics.advance as f64).max(MIN_RECT_WIDTH);
+            //     f(Rect::new(x, line_min, x + width, line_max));
+            // }
```
xorgy, line 757: Does this belong here anymore?

parley/src/layout/editor.rs
```diff
@@ -554,21 +577,40 @@ where
...
-                .unwrap_or_default();
-            ActiveText::FocusedCluster(self.selection.focus().affinity(), &self.buffer[range])
+            // let range = self
+            //     .selection
+            //     .focus()
+            //     .cluster_path()
+            //     .cluster(&self.layout)
+            //     .map(|c| c.text_range())
+            //     .unwrap_or_default();
+            ActiveText::FocusedCluster(self.selection.focus().affinity(), "")
```
xorgy, line 613: :eye: :eye:

parley/src/layout/line/greedy.rs
```diff
@@ -584,6 +542,11 @@ impl<'a, B: Brush> BreakLines<'a, B> {
                         line.metrics.descent = run.metrics.descent;
                         line.metrics.leading = run.metrics.leading;
                     }
+                } else if let Some(metrics) = prev_line_metrics {
+                    // HACK: copy metrics from previous line if we don't have
+                    // any; this should only occur for an empty line following
+                    // a newline at the end of a layout
```
xorgy, line 548: Is this really a hack?

xorgy (dismissed):
Glad to see this coming along. :+ )

parley/src/layout/editor.rs
```diff
@@ -160,46 +168,59 @@ where
...
-                    }
-                })
-                .and_then(|c| c.map(|x| x.text_range().start))
+                .logical_clusters(&self.editor.layout)[0]
+                .clone()
             {
+                let range = cluster.text_range();
+                let end = range.end;
+                let start = if cluster.info().whitespace() == Whitespace::Newline
+                    || cluster.info().is_emoji()
```
DJMcNab, line 202: `info` is outside of the visibility boundary. That is, I believe the editor component is something we want to support customising externally (obvious example, for rich text editing, which is clearly out-of-scope). Therefore, it shouldn't have a dependency on anything `pub(crate)` in Parley.

One half of this condition is already public; I'd suggest adding a similar helper for the other half:
```suggestion
                let start = if cluster.is_hard_line_break()
                    || cluster.is_emoji()
```

parley/src/layout/editor.rs
```diff
@@ -160,46 +168,59 @@ where
...
+                {
+                    // For newline sequences and emoji, delete the previous cluster
+                    range.start
+                } else {
+                    // Otherwise, delete the previous character
```
DJMcNab, line 207:
```suggestion
                    // Otherwise, delete the previous character (unicode scalar value)
```

parley/src/layout/editor.rs
```diff
@@ -292,46 +311,45 @@ where
 
     /// Move to the next cluster left in visual order.
     pub fn move_left(&mut self) {
-        self.editor
-            .set_selection(self.editor.selection.previous_visual(
-                &self.editor.layout,
-                self.editor.cursor_mode,
-                false,
-            ));
+        self.editor.set_selection(
```
DJMcNab, line 314: This is a pre-existing concern, but I'll raise it now; should these methods call `refresh_layout`? That would be necessary if specifically the text has changed since the last relayout, and I don't think anything protects against that.

parley/src/layout/cursor.rs
```diff
@@ -785,48 +711,62 @@ impl Selection {
...
+        }
+    }
+
+    pub(crate) fn maybe_extend(&self, focus: Cursor, extend: bool) -> Self {
```
DJMcNab, line 761: A function of this name used to be public; it's used within `editor.rs` (with `extend: true`). Can we just expose an `extend` function?

DJMcNab (commented):
I've ported to this in https://github.com/linebender/xilem/pull/755. Overall, this works well, except for the mentioned privacy concerns.

Additionally, there seems to be a new crash if the `layout`/`buffer` is exactly empty when layout is called, but only the first time.
That is, having an empty initial text crashes, but editing so that the text is empty does not. I have not dug into this very far.

mwcampbell:
Is the fix for the problem with accessibility on the last line waiting on something from me?

dfrg:
> Is the fix for the problem with accessibility on the last line waiting on something from me?

No, I plan on adding an empty run to the last line to see if that helps. Just haven’t gotten to it yet. I’ll ping here when the PR is updated.

PoignardAzur:
Pressing `Delete` while the cursor is a the beginning of a line puts the cursor in an invalid position.

parley/src/layout/cursor.rs
dfrg: The repetition is a bit annoying. Fixed.

parley/src/layout/cursor.rs
dfrg: There were a few instances of this.. I thought `cargo fmt --all` fixed this stuff.

parley/src/layout/cursor.rs
dfrg: It does, when the loop actually constrains the expensive code to the first and last lines. Restored all of this code.

parley/src/layout/editor.rs
dfrg: I have some new debugging code that is more useful so I'll come up with a way to fit that in.

parley/src/layout/editor.rs
dfrg: Added public `is_emoji()` and changed these conditions.

parley/src/layout/cluster.rs
```diff
@@ -394,10 +372,12 @@ impl<'a, B: Brush> Cluster<'a, B> {
 /// Determines how a cursor attaches to a cluster.
 #[derive(Copy, Clone, PartialEq, Eq, Default, Debug)]
 pub enum Affinity {
-    /// Left side for LTR clusters and right side for RTL clusters.
+    /// Cursor is attached to the character that is logically following in the
+    /// text stream.
```
DJMcNab, line 376: Nice! This is so much better.

parley/src/layout/cursor.rs
```diff
@@ -101,128 +101,135 @@ impl Cursor {
         self.affinity
     }
 
-    /// Returns a new cursor that is positioned at the next cluster boundary
+    /// Returns a new cursor that is positioned at the previous cluster boundary
```
xorgy, line 104: Is there a reason for swapping the order of these? it makes it kinda messy to read the diff.

dfrg:
> Pressing `Delete` while the cursor is a the beginning of a line puts the cursor in an invalid position.

Fixed this by just removing the selection adjustment code on delete since that operation doesn’t modify the cursor position.

dfrg:
Moving through these bit by bit. Note: this should now handle proper drag selection at word and line granularity.

nicoburns:
> As it allows both the selections {lorem} ipsum and {lorem }ipsum. Firefox only allows the former on my machine.

FWIW, Chrome allows both on my machine.

xStrom:
On my Windows 11 machine both Firefox and Chrome allow both selections.

dfrg:
Word selection and movement is entirely inconsistent across platforms/editors so I chose a set of behaviors that seemed to make sense. Including space-like “words” in mouse selection was intentional because you cannot otherwise extend the selection into a space without using the keyboard, and as noted, there is precedent for this in Chrome.

parley/src/layout/editor.rs
dfrg: I removed `ActiveText` entirely (it was meant for debugging) and replaced the necessary functionality with `PlainEditor::selected_text(&self) -> Option<&str>`

I left my current debugging code in but commented out in `set_selection`. We can clean this up later.

parley/src/layout/cursor.rs
dfrg: Added a new `extend(focus: Cursor)` method to support this.

parley/src/layout/editor.rs
dfrg: I think I'll defer issues around this sort of thing to future work.

mwcampbell:
The phantom run for a blank line at the end seems to work correctly. I can find just one more accessibility problem: if the text is entirely empty, there's a phantom run, but `Cursor::to_accesskit_position` returns `None` because there's no cluster. In that case, we need to return the node ID of the one phantom run and a character index of 0.

dfrg:
Thanks Matt. I added a special case for that.

Assuming this is okay for accessibility, I don't plan on making any further changes unless further issues are raised by others.

mwcampbell:
The AccessKit integration now looks good to me.

dfrg:
> The AccessKit integration now looks good to me.

Great! I appreciate your patience and your help with testing.

parley/src/layout/editor.rs
xorgy: I get it now thanks.

nicoburns:
We have a cursor when there's no text! (rendered with Blitz + Parley):

Screenshot 2024-11-28 at 11 07 02

nicoburns (approved):
I've done some testing of the `vello_editor` example, and this seems to resolve all of the robustness issues around newlines I was seeing before. I've also updated Blitz to use this branch, which seems to work well.

dfrg:
Thanks Nico. I’m going to go ahead and merge this. If there are further issues they can be addressed in a follow up.

PR #172, opened 2024-11-11
vello_editor: Stop blinking and hide cursor when window not focused.

xorgy:


mwcampbell:
I believe the norm is for the cursor to disappear when the window loses focus, not become permanently shown.

xorgy:
You're right, or in the case of overwrite block cursors it usually becomes an empty square. I'll switch it to that.

mwcampbell (approved):
LGTM

PR #173, opened 2024-11-15
Initial version of snapshot testing

spirali:
Hi,

This PR contains an initial version of snapshot testing.

By design it contains only a minimal API and usage inspired by vello snapshot tests.
The ambition of this PR is not to do any test coverage, it contains only two tests for demonstration purposes of the testing environment.

All the bells and whistles (html report, image diffs, not loading fonts in every test) are now missing to make the code more straightforward.

# Usage

```bash
$ cargo test
```

If a test fails, you can compare images in `parley/tests/current` (images created by the current test)
and `parley/tests/snapshots` (the accepted versions).

If you think that everything is ok, you can start tests as follows:

```bash
$ PARLEY_TEST="accept" cargo test
```

It will update snapshots of the failed tests.

# How it works

It reuses `tiny-skia-renderer` and renders the testing layout into an image and compares it pixel by pixel.

This PR also adds fonts into the tests, to have tests independent on the system fonts. I have chosen DejaVu fonts, but it was a random pick.


DJMcNab (changes requested):
Thank you - it will be great to get some testing landing.

However, We definitely shouldn't import several megabytes of fonts; I'd suggest using only two fonts files; presumably we do want to test multi-font cases.
We also need to follow the license for whatever fonts we do import.

spirali:
> We definitely shouldn't import several megabytes of fonts.
> 
> We also need to follow the license for whatever fonts we do import.

The fonts are used only in tests, Parley itself can be build without it.

In the last change I have added license of the fonts.

spirali:
> Thank you - it will be great to get some testing landed.
> 
> However, We definitely shouldn't import several megabytes of fonts - it's pretty much a worst-case scenario for git; I'd suggest using only two fonts files; presumably we do want to test multi-font cases. We also need to follow the license for whatever fonts we do import.

Ok, I will try to prune it. However it would not be that bad for git, these files should only rarely changed, so it just a few constant megabytes. 

I have also attached more variants so we can test e.g. different weights, because variable weighted font does not for me in Parley (I do not know if it a feature or a bug).

spirali:
I have removed (and squashed) other variants and left just DejaVu Sans and DejaVu Sans Bold (= 2x 700KiB)

xorgy:
It might make more sense to use a pinned version of a distro font package in the CI (maybe Noto?)

spirali:
I have completely removed fonts from repository & added  a script that downloads Noto Fonts.

spirali:
Note: CI is now failing, because font download is not part of CI scripts.

parley/Cargo.toml
```diff
@@ -31,3 +31,6 @@ fontique = { workspace = true }
 core_maths = { version = "0.1.0", optional = true }
 accesskit = { workspace = true, optional = true }
 hashbrown = { workspace = true, optional = true }
+
+[dev-dependencies]
+tiny-skia = "0.11"
```
xStrom, line 36: Let's use the full version.

```suggestion
tiny-skia = "0.11.4"
```

DJMcNab:
We had some discussion in office hours; the suggested path forward was to have one or two small font files in this repository. For example, in Vello, we have Roboto and Inconsolata. Those files in total are almost exactly 500KiB; I'd target a megabyte as the hard limit for total font files in the repository initially. An important tool here is subsetting - see for example https://github.com/linebender/vello/tree/main/examples/assets/noto_color_emoji which describes doing that for Emoji.
My preference would be to choose fonts with either the OFL 1.1 or the Apache 2.0 license, because those are well-understood licenses.
Using downloaded fonts has reproducibility and stability challenges, as the download service is a very easy point of failure. 

At a first pass, the rest of the code looks really good, I hope to find time for a more thorough review soon.

This is a collaborative process, so feel free to push back on this - I recognise that we moved on quite quickly from this in the meeting today without getting confirmation from you.

spirali:
Hi,
I agree with the outcome of the discussion.
Last changes:
* I removed the font download script
* I added Roboto fonts used in Vello tests (Apache 2 licence) as a baseline (168KiB). When some tests would need some extra fonts, we can accept other fonts together with PR of these tests.
* I fixed `tiny_skia` to `0.11.4`

parley/src/tests/test_basic.rs
```diff
@@ -0,0 +1,37 @@
...
+    let mut builder = env.builder(text);
+    let mut layout = builder.build(text);
+    layout.break_all_lines(None);
+    layout.align(None, Alignment::Start);
+
+    env.check(&layout);
+}
+
+#[test]
+fn test_placing_inboxes() {
```
DJMcNab, line 20:
```suggestion
fn simple_inline_boxes() {
```

parley/src/tests/test_basic.rs
```diff
@@ -0,0 +1,37 @@
+// Copyright 2024 the Parley Authors
+// SPDX-License-Identifier: Apache-2.0 OR MIT
+
+use crate::{testenv, Alignment, InlineBox};
+
+#[test]
+fn test_plain_multiline_text() {
```
DJMcNab, line 7: The `test_` prefix is not idiomatic in Rust - see https://github.com/rust-lang/rust-clippy/issues/8931; the same probably also applies to the module name, since it's in the `tests` folder.
```suggestion
fn plain_multiline_text() {
```

parley/src/tests/utils/env.rs
```diff
@@ -0,0 +1,227 @@
...
+use fontique::{Collection, CollectionOptions};
+use std::sync::Once;
+use tiny_skia::Pixmap;
+
+static INIT: Once = Once::new();
+
+/// Initialize tests
+/// Removes all images from directory with the current test images
+pub(crate) fn initialize_tests() {
+    INIT.call_once(|| {
```
DJMcNab, line 70: I like the idea. Unfortunately, I don't expect this to work in our set up because we use [`cargo nextest`](https://nexte.st/) for our testing, which runs multiple tests in different processes.

The way we handle the equivalent directory in Vello is to just delete the tests in the corresponding happy path (i.e. when the test succeeds, delete the comparison image).

parley/src/tests/utils/env.rs
```diff
@@ -0,0 +1,227 @@
...
+        let mut n_different_pixels = 0;
+        let mut color_cumulative_difference = 0.0;
+        for (pixel1, pixel2) in snapshot_img.pixels().iter().zip(current_img.pixels()) {
+            if pixel1 != pixel2 {
+                n_different_pixels += 1;
+            }
+            let diff_r = (pixel1.red() as f32 - pixel2.red() as f32).abs();
+            let diff_g = (pixel1.green() as f32 - pixel2.green() as f32).abs();
+            let diff_b = (pixel1.blue() as f32 - pixel2.blue() as f32).abs();
+            color_cumulative_difference += diff_r.max(diff_g).max(diff_b);
```
DJMcNab, line 182: Hmm, so this is the [Chebyshev](https://en.wikipedia.org/wiki/Chebyshev_distance) difference? I'm not very familiar with why you'd use this over a different (ha) difference metric for this case.

parley/src/tests/utils/env.rs
```diff
@@ -0,0 +1,227 @@
...
+            for (path, msg) in &self.errors {
+                write!(
+                    &mut panic_msg,
+                    "{}\nImage written into: {}\n",
+                    msg,
+                    path.display()
+                )
+                .unwrap();
+            }
+            panic!("{}", &panic_msg);
```
DJMcNab, line 224: Does aborting if we're already panicking make sense here? I *think* it does, but I just want to gut check that.

parley/tests/assets/roboto_fonts/LICENSE.txt
DJMcNab, line 1: I think we also need a reference to this assets folder in the READMEs' license section (both the repository and Parley README files). I'd apply the wording from https://github.com/linebender/xilem

parley/tests/assets/roboto_fonts/Roboto-Regular.ttf
DJMcNab, line 1: We should make sure that this file is [excluded](https://doc.rust-lang.org/cargo/reference/manifest.html#the-exclude-and-include-fields) from the package. This can be validated with `cargo package --list`

parley/tests/current/.gitignore
```diff
@@ -0,0 +1 @@
+*.png
```
DJMcNab, line 1:
```suggestion
*.png

```
Our style is to always have a final newline.

parley/src/tests/utils/renderer.rs
```diff
@@ -0,0 +1,217 @@
+// Copyright 2024 the Parley Authors
+// SPDX-License-Identifier: Apache-2.0 OR MIT
+
+//! A simple example that lays out some text using Parley, extracts outlines using Skrifa and
+//! then paints those outlines using Tiny-Skia.
+//!
+//! Note: Emoji rendering is not currently implemented in this example. See the swash example
+//! if you need emoji rendering.
```
DJMcNab, line 8: I am slightly worried by the code duplication with the existing example. However, I don't think there's a good way for the code to be shared, unfortunately.

parley/src/tests/utils/env.rs
```diff
@@ -0,0 +1,227 @@
...
+        builder.push_default(StyleProperty::FontStack(FontStack::Single(
+            FontFamily::Named(DEFAULT_FONT_NAME.into()),
+        )));
+        builder
+    }
+
+    fn image_name(&mut self) -> String {
+        let name = format!("{}-{}.png", self.test_name, self.check_counter);
+        self.check_counter += 1;
+        name
```
DJMcNab, line 152: My preference would be for the name of a test to be passed explicitly.
I say this both so that adding a new test early doesn't move all of the existing tests out of sync, and so that the image names are self descriptive.
If we ended up duplicating the cool side-by-side view you showed us on Zulip, this would also be really nice there. (But that's definitely not a job for this PR).

However, this is a tradeoff, because naming tests is hard. Leaving it as is would be fine for this PR.

parley/src/tests/utils/env.rs
```diff
@@ -0,0 +1,227 @@
...
+    check_counter: u32,
+    font_cx: FontContext,
+    layout_cx: LayoutContext<Color>,
+    foreground_color: Color,
+    background_color: Color,
+    tolerance: f32,
+    errors: Vec<(PathBuf, String)>,
+}
+
+use fontique::{Collection, CollectionOptions};
```
DJMcNab, line 61: Please leave all imports at the top of the file.

DJMcNab (commented):
This looks like some really useful infrastructure, and some of the ideas in here look really useful to bring to other repositories.

There are a few things I'd like to see changed before landing this; the only strict blocker is mentioning the font's license in the top-level READMEs. As it stands currently, if we merged this we would be representing that roboto is available under the MIT license, which is incorrect. I realise that this might seem pedantic, but licensing is unfortunately an area where it's really important to be accurate.

parley/src/tests/utils/env.rs
```diff
@@ -0,0 +1,227 @@
...
+macro_rules! testenv {
+    () => {{
+        // Get name of the current function
+        fn f() {}
+        fn type_name_of<T>(_: T) -> &'static str {
+            std::any::type_name::<T>()
+        }
+        let name = type_name_of(f);
+        let name = &name[..name.len() - 3];
+        let name = &name[name.rfind(':').map(|x| x + 1).unwrap_or(0)..];
```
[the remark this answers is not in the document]
spirali: I agree that it seems slightly fragile, but I did some research and AFAIK there is not other way in the current Rust how to do it and Insta crate use the similar trick.

parley/src/tests/test_basic.rs
spirali: Fixed

parley/src/tests/test_basic.rs
spirali: Fixed

parley/src/tests/utils/env.rs
spirali: Changed to: Each test clears its own images at the beginning of the test (not only happy path). Global init removed.

parley/src/tests/utils/env.rs
spirali: I have no strong argument for this metric, only that when I experimented with tests for Nelsie, it seemed to work better. My intuition: If you go from black to (say) full red, it makes more sense to me that it is the same error as going to full white, rather than having 1/3 error for red (as happens with simple averaging).
In practice, I have observed that the Chebyshev* metric does a better job of distinguishing small changes in font rendering from real font positioning errors. But I have no data for this, it was just a casual observation on some of my tests.

  *) To be honest, I just experimented with some metrics off the top of my head, I never heard this name before :)

parley/src/tests/utils/env.rs
spirali: You are right. We should not panic when we are already panicking. I have fixed this.

parley/tests/assets/roboto_fonts/LICENSE.txt
spirali: Fixed

parley/tests/assets/roboto_fonts/Roboto-Regular.ttf
spirali: Fixed

parley/src/tests/utils/renderer.rs
spirali: I think that reasonable sharing of the code would lead to a less readable example. I also do not think that there is a good way for remove this duplication.

parley/src/tests/utils/env.rs
spirali: The name of the image is automatically generated from the name of the test, so it it easy to find a corresponding test that generates the image. The counter is used for solving the situation that a test generates more images. 

But maybe I am misunderstanding this comment.

parley/src/tests/utils/env.rs
spirali: Fixed

parley/src/tests/utils/env.rs
DJMcNab: As I say, I don't really care. If we know it works, then that's all we need.

parley/src/tests/utils/env.rs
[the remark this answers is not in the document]
DJMcNab: Agreed. I don't think it's the wrong thing to use, just that it feels like it should be.

parley/src/tests/utils/env.rs
DJMcNab: So specifically, my preference would be for the format to be `{test_name}-{snapshot_name}`; i.e. each snapshot gets a descriptive name (each test in turn could then use their own indexing if they wanted to).

parley/README.md
```diff
@@ -52,4 +52,8 @@ Licensed under either of
 
 at your option.
 
+Some files used for tests are under different licenses:
+
+- The font file `Roboto-Regular.ttf` in `/tests/parley/assets/roboto_fonts/` is licensed solely as documented in that folder (and is licensed under the Apache License, Version 2.0).
+
```
DJMcNab, line 58: We also need this section in the top-level README (i.e. the one at `/README.md` - https://github.com/linebender/parley/blob/main/README.md).

parley/src/tests/utils/env.rs
```diff
@@ -0,0 +1,221 @@
...
+        }
+        if color_cumulative_difference > self.tolerance {
+            return Err(format!(
+                "Testing image differs in {n_different_pixels} pixels (color difference = {color_cumulative_difference})",
+            ));
+        }
+        Ok(())
+    }
+
+    pub(crate) fn check(&mut self, layout: &Layout<Color>) {
```
DJMcNab, line 182: I wonder if this could be renamed; check is a very passive name which doesn't suggest filesystem access.

Maybe just `screenshot` or `snapshot`?

DJMcNab (commented):
The last blocker is the top-level README licensing change.
Otherwise I think this is ready to land.

I think that the screenshots should probably be stored in LFS, as is done in Vello. However, since there are only ~30KiB of screenshots in this PR, it's not a blocking concern. It's nice that these screenshots run across all platforms.

parley/src/tests/utils/env.rs
spirali: Renamed to `check_snapshot`.

parley/README.md
spirali: Fixed

parley/src/tests/utils/env.rs
spirali: I like the idea and I originally planed it put it in this PR, but I have stripped it to make it more straightforward.
I added `check_snapshot_with_name`, it is a simple change.

DJMcNab (approved):
Thanks!

Feel free to merge this yourself when you're certain it's ready!

PR #175, opened 2024-11-17
Use `arboard` instead of `clipboard_rs` in `vello_editor` example

nicoburns:
`clipboard_rs` is unmaintained and also unnecessarily pulls in every image format from the `image` crate. This PR switches to the `arboard` crate which is a maintained fork of `clipboard_rs` (and seems to be the most popular fork, although there is not currently one clear "winner" in this space)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -109,12 +109,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "adler"
-version = "1.0.2"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "f26201604c87b1e01bd3d98f8d5d9a8fcbb815e8cedb41ffccbeb4bf593a35fe"
-
-[[package]]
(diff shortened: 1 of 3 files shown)


waywardmonkeys:
I think you've confused `clipboard-rs` with `clipboard`.  `clipboard-rs` is maintained with a release only 2 months ago and the author merged some PRs of mine last month.

I had a commit locally to submit to clean up the `image` dependency, but I see that I forgot to do so. I've updated and submitted it now: https://github.com/ChurchTao/clipboard-rs/pull/53

Let's give him a shot at responding this week.

If we do end up going and merging this, you'll need to edit your PR description to be accurate.

waywardmonkeys:
I coordinated with upstream and a new release of `clipboard-rs` is out that uses fewer `image` features.

nicoburns:
Right, I guess we can close this then.

PR #176, opened 2024-11-18
Fix Cursor::from_point to use the line's offset

DJMcNab:
This is effectively the opposite of #169.

I've also added a test. This test does use system fonts, and so is written so as to be *relatively* independent of the font, assuming it doesn't do any ligature fanciness.

In future, we'll likely want to us an in-repository small font (such as roboto).


jaredoconnell (approved):
I can confirm that this fixes the cursor positioning in Xilem.

DJMcNab:
Review requested because of the test

PR #177, opened 2024-11-18
Make Data types and fields `pub(crate)`

PoignardAzur:
This makes it more clear that these types aren't publicly exported and are meant to be internal types.


DJMcNab (commented):
Can we also enable `unreachable_pub` at the same time?

nicoburns (approved):

PR #180, opened 2024-11-19
Add linebender lint set to project

PoignardAzur:
This comes with a lot of carve-outs to keep the diff minimal. Further PRs can remove these carve-outs.

The `#[allow]` attributes will need to be replaced with `#[expect]` when the MSRV bumps.


parley/src/testing/mod.rs
DJMcNab, line 1: ?

DJMcNab (approved):
Seems reasonable, except for the one artefact

parley/src/testing/mod.rs
PoignardAzur: Oh yeah, added that on accident.

PR #181, opened 2024-11-19
Update clipboard-rs

waywardmonkeys:
This brings in the changes I coordinated with upstream to remove some of the default features from the `images` build.


waywardmonkeys:
This replaces #175.

xorgy:
I wonder how hard it would be to make clipboard-rs feature gate on non-text clips..

PR #182, opened 2024-11-20
Skip adding font family as fallback if it has zero coverage for given script

### LINKED ISSUE #174: Arabic font fallback not working on linux
Hello, I wanted to test out this library, so I've run the swash_render example on my system and the font fallback for arabic text does not work as expected.

Here is rendered swash_output.png from the example:
swash_render

Is there anything I can test to debug this issue?

richardhozak:
Fontconfig can be and is so in some distros, configured wrong. It can report fonts as a fallback for given language when in reality the font family has zero coverage for given language.

This change does preliminary coverage check for fonts loaded from fontconfig and skips font families that have zero coverage for given language it reports.

This PR also contains small fix for calculating coverage, where it could never match font with full coverage for multi-byte chars as computing coverage returns how many characters it matches but we compared it to number of bytes in a string.

Fixes #174

I've tested this on fresh install of PopOS where the font fallback for Arabic script failed and works with this change.
I've also tested this on fresh install of Ubuntu 24.04 where fontconfig is configured correctly, fallback worked correctly without this change and still works correctly with this change.

### DIFF
--- a/fontique/src/backend/fontconfig/mod.rs
+++ b/fontique/src/backend/fontconfig/mod.rs
@@ -158,6 +158,26 @@ impl SystemFonts {
             };
             for &script in scripts {
                 let script = Script(*script);
+
+                // check if fallback family has any coverage, if not skip adding it
+                let any_coverage = script
+                    .sample()
+                    .and_then(|sample| {
+                        raw_families.get(&family_id).map(|raw_family| {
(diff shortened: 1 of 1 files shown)


dfrg (approved):
This is exactly what I had in mind and the additional fix is a great catch. Thanks!

PR #187, opened 2024-11-28
Fix off-by-one error in `PlainEditor::cursor_at`

tomcur:
I believe the cursor should still land at index `self.buffer.len()` (logically following the last cluster).

E.g., in https://github.com/linebender/xilem/pull/762, if used without this change, the selection can't span the last cluster using `cursor_at`.


DJMcNab (approved):
Agreed that this makes sense

PR #191, opened 2024-12-01
Fix whitespace collapsing at the start of inline spans

nicoburns:
Generally, whitespace collapsing collapses sequences of whitespace into a single space character. However, whitespace at the very start or end of a span should be elided entirely.

We use a flag to track if we are in this state. However, there was a bug which was causing this flag to cleared when whitespace itself was pushed, which caused start-of-span whitespace to appear as a single space if it is pushed into the builder in multiple chunks.

This patch prevents the flag from being cleared by whitespace to fix that bug.

This patch fixes the rendering of a number of websites in Blitz including https://alistapart.com, an older snapshot of https://google.com (the latest version doesn't run into this bug), and around 40 WPT tests.

### Blitz rendering google.com with this patch

Screenshot 2024-12-02 at 10 58 06

### DIFF
--- a/parley/src/resolve/tree.rs
+++ b/parley/src/resolve/tree.rs
@@ -107,10 +107,8 @@ impl<B: Brush> TreeStyleBuilder<B> {
         };
         let span_text = span_text.as_ref();
 
-        // Nothing to do if there is no uncommitted text
+        // Nothing to do if there is no uncommitted text.
         if span_text.is_empty() {
-            // This is for the case of an inline box. This possibly ought to be made more explicit.
-            self.is_span_first = false;
             return;
         }
 


DJMcNab:
No comment on the change specifically, but this feature existing here at all is unexpected; these *feel* like higher-level semantics - it is strange that this is supported in the tree builder but not the range builder.

I haven't reasoned about this very carefully, but my reading of the newline behaviour is also quite unintuitive. It seems like if you have a whitespace character before or after a newline, the whole combination gets replaced with a single space?

dfrg:
> No comment on the change specifically, but this feature existing here at all is unexpected; these _feel_ like higher-level semantics

We don’t handle this currently but if we intend to support proper selection and editing in web style content then the layout itself will need to be aware of whitespace collapse (and text transform) to do accurate offset mapping.

> it is strange that this is supported in the tree builder but not the range builder.

Ranged builder will likely go away at some point but even so, I don’t think whitespace collapse is generally considered in “attributed text” rich styling models so the discrepancy makes sense to me. 

> I haven't reasoned about this very carefully, but my reading of the newline behaviour is also quite unintuitive. It seems like if you have a whitespace character before or after a newline, the whole combination gets replaced with a single space?

Yeah, the web is weird and unless you use one of the preserving modes, all whitespace of any form is collapsed to a single space character.

dfrg (approved):
I also haven’t fully reasoned about how this handles whitespace collapse but since this code currently exists primarily to support Blitz, I’m happy to defer to Nico’s judgment here. LGTM

nicoburns:
I recently found this article on HTML whitespace collapsing which I would recommend: https://blog.dwac.dev/posts/html-whitespace

The TL;DR is that sequences of whitespace collapse to either:
1. Nothing (the disappear entirely). This is the case for whitespace immediately following an opening HTML tag, or immediately preceding a closing HTML tag.
2. A single space. This is the case for whitespace everywhere else.

I am not a huge fan of this system. But it turns out to be absolutely essential to getting HTML to render anything close to correctly (which makes sense when you consider what large quantities of spurious whitespace would do to a layout).

The bug this PR fixes is a case which should have been (1) hitting the (2) codepath instead because whitespace itself was clearing the "immediately following an opening tag" flag.

Issue #186, opened 2024-11-28
Crash with no text in layout and very small widths


DJMcNab:
To reproduce, you can apply this patch then select all and press backspace or delete:

<details><summary>Patch to reproduce</summary>
<p>

```diff
diff --git a/examples/vello_editor/src/main.rs b/examples/vello_editor/src/main.rs
index 0422e76..8ab97cf 100644
--- a/examples/vello_editor/src/main.rs
+++ b/examples/vello_editor/src/main.rs
@@ -119,7 +119,7 @@ impl ApplicationHandler<accesskit_winit::Event> for SimpleVelloApp<'_> {
 
         self.editor.transact(|txn| {
             txn.set_scale(1.0);
-            txn.set_width(Some(size.width as f32 - 2f32 * text::INSET));
+            txn.set_width(Some(0.0)); // This can be a higher value and still reproduce
             txn.set_text(text::LOREM);
         });
 
@@ -238,9 +238,8 @@ impl ApplicationHandler<accesskit_winit::Event> for SimpleVelloApp<'_> {
                     .resize_surface(&mut render_state.surface, size.width, size.height);
                 self.editor.transact(|txn| {
                     txn.set_scale(1.0);
-                    txn.set_width(Some(size.width as f32 - 2f32 * text::INSET));
                     txn.set_default_style(Arc::new([
-                        StyleProperty::FontSize(32.0),
+                        StyleProperty::FontSize(16.0),
                         StyleProperty::LineHeight(1.2),
                         GenericFamily::SystemUi.into(),
                     ]));
diff --git a/examples/vello_editor/src/text.rs b/examples/vello_editor/src/text.rs
index 2211c7e..178b033 100644
--- a/examples/vello_editor/src/text.rs
+++ b/examples/vello_editor/src/text.rs
@@ -77,7 +77,7 @@ impl Editor {
     pub fn handle_event(&mut self, event: WindowEvent) {
         match event {
             WindowEvent::Resized(size) => {
-                self.transact(|txn| txn.set_width(Some(size.width as f32 - 2f32 * INSET)));
+                // self.transact(|txn| txn.set_width(Some(1.0)));
             }
             WindowEvent::ModifiersChanged(modifiers) => {
                 self.modifiers = Some(modifiers);
```

</p>
</details> 

On my machine, any width less than an unknown value in [3.52, 3.521) causes a crash. The backtrace is:

<details><summary>Backtrace</summary>
<p>

```text
[...]

PR #194, opened 2024-12-03
Fix crash when laying out an empty layout with small max_advance

nicoburns:
- Fixes https://github.com/linebender/parley/issues/186

I don't really understand why this works (why it crashes with a smaller width but not a larger one in the first place) But seeing as the `max_advance` doesn't really makes any difference for "empty" layouts that only contain a single space, it seems safe to ignore it here.

### DIFF
--- a/parley/src/layout/line/greedy.rs
+++ b/parley/src/layout/line/greedy.rs
@@ -144,6 +144,14 @@ impl<'a, B: Brush> BreakLines<'a, B> {
         }
         self.prev_state = Some(self.state.clone());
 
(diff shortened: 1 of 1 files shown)


dfrg (approved):
We already have a number of hacks for empty layouts so adding another that prevents a crash doesn’t trouble me. We’ll address these when we redo line breaking. Thanks!

nicoburns:
> We already have a number of hacks for empty layouts so adding another that prevents a crash doesn’t trouble me.

Yep, that was also my thought!

PR #195, opened 2024-12-04
WIP: Fix inline box only wrap

nicoburns:
https://github.com/linebender/parley/pull/194 broke line-wrapping for layouts that only contain inline boxes. This PR adds an extra check to the `max_advance` hack to fix that. It also adds a test to check that the line-wrapping is correct (which it is NOT).

Fixing the inline box wrapping is quite easy. Unfortunately every fix I have come up with so far breaks line-wrapping for regular glyph runs. I'm pretty sure there's an off-by-one error (or similar) somewhere that differs between inline boxes and glyph runs (or between some of the cases within each).

It's also possible that this is ultimately the cause of the crash our hack is working around.

Notes:
- The `.saturating_sub(1)` (`state.items.end.saturating_sub(1)`) in `try_commit_line()` is very suspect to me. Removing it fixes the inline box case but causes the first wrapped glyph run to be skipped.
- The snapshot for the test is correct, but the current code in this PR (or `main`) doesn't generate that result.


nicoburns:
Superceded by #207

PR #198, opened 2024-12-05
Handle IME selection cursor hiding

tomcur:
When the IME indicates a cursor of `None`, it means we should hide the cursor. This does that by collapsing the selection, hiding the caret, and clearing the text selection from the AccessKit node.

### DIFF
--- a/parley/src/layout/editor.rs
+++ b/parley/src/layout/editor.rs
@@ -90,15 +90,6 @@ impl<'source> IntoIterator for SplitString<'source> {
     }
 }
 
-/// IME compose state.
-#[derive(Clone)]
-struct Compose {
-    /// Byte offsets of IME composing preedit text in the text buffer.
-    preedit_range: Range<usize>,
-    /// Whether the cursor should be shown. The IME can request to hide the cursor.
-    show_cursor: bool,
-}
-
 /// Basic plain text editor with a single style applied to the entire text.
 ///
 /// Internally, this is a wrapper around a string buffer and its corresponding [`Layout`],
@@ -115,8 +106,11 @@ where
     #[cfg(feature = "accesskit")]
     layout_access: LayoutAccessibility,
     selection: Selection,
-    /// The IME compose state; `None` if the IME is not currently composing.
-    compose: Option<Compose>,
+    /// Byte offsets of IME composing preedit text in the text buffer.
+    /// `None` if the IME is not currently composing.
+    compose: Option<Range<usize>>,
+    /// Whether the cursor should be shown. The IME can request to hide the cursor.
+    show_cursor: bool,
     width: Option<f32>,
     scale: f32,
     // Simple tracking of when the layout needs to be updated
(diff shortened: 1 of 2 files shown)


parley/src/layout/editor.rs
```diff
@@ -1005,11 +1022,15 @@ where
...
-            node.set_text_selection(selection);
+        if self.compose.as_ref().map(|c| c.show_cursor).unwrap_or(true) {
+            if let Some(selection) = self
+                .selection
+                .to_access_selection(&self.layout, &self.layout_access)
+            {
+                node.set_text_selection(selection);
+            }
+        } else {
+            node.clear_text_selection();
```
tomcur, line 1033: Is clearing the text selection, in response to the IME hiding the cursor, the right thing to do? 

(I also wonder (unrelated to this PR), whether doing nothing when `to_access_selection` is `None` is right, though it appears that only happens when in a strange state.)

parley/src/layout/editor.rs
```diff
@@ -90,6 +90,15 @@ impl<'source> IntoIterator for SplitString<'source> {
...
 }
 
+/// IME compose state.
+#[derive(Clone)]
+struct Compose {
+    /// Byte offsets of IME composing preedit text in the text buffer.
+    preedit_range: Range<usize>,
+    /// Whether the cursor should be shown. The IME can request to hide the cursor.
+    show_cursor: bool,
+}
```
tomcur, line 100: As a style note, I could be convinced to move both of these to `PlainEditor`... It's a trade-off between type-level correctness and ergonomics. In fact, "not-composing" could be encoded with an empty `preedit_range`, which with the current layout would shave 8 bytes off `PlainEditor`.

parley/src/layout/editor.rs
```diff
@@ -757,21 +762,33 @@ where
...
+        // We do not check the compose state's `show_cursor` here, as the IME handling code
+        // collapses the selection to a caret in that case.
         self.selection.geometry(&self.layout)
     }
 
     /// Get a rectangle representing the current caret cursor position.
+    ///
+    /// There is not always a caret. For example, the IME may have indicated the caret should be
+    /// hidden.
     pub fn cursor_geometry(&self, size: f32) -> Option<Rect> {
```
DJMcNab, line 774: Handy that this was already an option...

parley/src/layout/editor.rs
DJMcNab: `self.compose.as_ref().map(|c| c.show_cursor).unwrap_or(true)` is so ugly (even `self.compose.as_ref().is_none_or(|c| c.show_cursor)` is still pretty bad) that I think flattening this is a win. It's effectively only set in two functions anyway, so it's unlikely to get out-of-sync.

DJMcNab (approved):
I'm surprised that this isn't a breaking change. I think flattening the new struct is nicer, because the current ergonomics are impressively bad.

parley/src/layout/editor.rs
tomcur: Agreed.

PR #202, opened 2024-12-06
Ignore affinities in Selection::is_collapsed

tomcur:
Currently `is_collapsed` is false when a selection's anchor and focus are at the same byte position, but have different affinities.

This changes `is_collapsed` to ignore the affinities, which appears to be the sense of "collapsed" that is assumed throughout, e.g.,`Selection::next_visual` currently requires doing the operation twice when only affinities differ to actually move and `PlainEditor::selected_text` currently returns `""` when only affinities differ.

### DIFF
--- a/parley/src/layout/cursor.rs
+++ b/parley/src/layout/cursor.rs
@@ -502,7 +502,7 @@ impl Selection {
     /// This means that the selection represents a single position rather than
     /// a range.
     pub fn is_collapsed(&self) -> bool {
-        self.anchor == self.focus
+        self.anchor.index == self.focus.index
     }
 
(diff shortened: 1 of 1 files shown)


DJMcNab (approved):
Does this match the pre-#170 behaviour?

This seems likely to be reasonable

tomcur:
Yes, it seems to match. For example, creating a selection in "AB" by pressing on the right half of "A" and dragging to the left half of "B" is currently considered not-collapsed due to differing affinities. Pre-#170 the anchor and focus indices would be different, but their `text_start` would match, which would be considered collapsed.

PR #203, opened 2024-12-06
Handle inline boxes in cluster geometry and path

tomcur:
Inline boxes are currently silently ignored in some places, making text selection wrong when they are used, as their advance is not used in offset calculation and the run indices are counted wrong as the boxes are skipped.

To see the issues currently present, insert the following in `PlainEditor::update_layout` and run vello_editor.

```rust
builder.push_inline_box(crate::InlineBox {
    id: 0,
    index: 7,
    width: 50.0,
    height: 50.0,
});
```

This PR fixes the issues I've found, though the cursor position when the cursor has a downstream affinity at the byte index of an inline box is still wrong, I believe. For example, in "AB" with an inline box between "A" and "B", when the cursor is between the two characters with downstream affinity, it shows up on the right side of the A, whereas it would be expected to show up at the left side of the B.

### DIFF
--- a/parley/src/layout/cluster.rs
+++ b/parley/src/layout/cluster.rs
@@ -21,11 +21,8 @@ impl<'a, B: Brush> Cluster<'a, B> {
         let mut path = ClusterPath::default();
         if let Some((line_index, line)) = layout.line_for_byte_index(byte_index) {
             path.line_index = line_index as u32;
-            for run_index in 0..line.len() {
-                let Some(run) = line.run(run_index) else {
-                    continue;
-                };
-                path.run_index = run_index as u32;
+            for run in line.runs() {
+                path.run_index = run.index;
                 if !run.text_range().contains(&byte_index) {
                     continue;
                 }
--- a/parley/src/layout/cluster.rs
+++ b/parley/src/layout/cluster.rs
@@ -21,7 +21,10 @@ impl<'a, B: Brush> Cluster<'a, B> {
         let mut path = ClusterPath::default();
         if let Some((line_index, line)) = layout.line_for_byte_index(byte_index) {
             path.line_index = line_index as u32;
-            for (run_index, run) in line.runs().enumerate() {
+            for run_index in 0..line.len() {
+                let Some(run) = line.run(run_index) else {
+                    continue;
+                };
                 path.run_index = run_index as u32;
                 if !run.text_range().contains(&byte_index) {
                     continue;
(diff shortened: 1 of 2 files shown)


parley/src/layout/cluster.rs
```diff
@@ -21,7 +21,10 @@ impl<'a, B: Brush> Cluster<'a, B> {
         let mut path = ClusterPath::default();
         if let Some((line_index, line)) = layout.line_for_byte_index(byte_index) {
             path.line_index = line_index as u32;
-            for (run_index, run) in line.runs().enumerate() {
+            for run_index in 0..line.len() {
```
tomcur, line 24: Unfortunately we can't just use `line.items()`, as `GlyphRunIter` currently appears to skip the run on empty lines.

parley/src/layout/cluster.rs
```diff
@@ -44,13 +44,17 @@ impl<'a, B: Brush> Cluster<'a, B> {
             path.line_index = line_index as u32;
             let mut offset = line.metrics().offset;
             let last_run_index = line.len().saturating_sub(1);
-            for (run_index, run) in line.runs().enumerate() {
+            for run_index in 0..line.len() {
```
tomcur, line 47: Unfortunately we can't just use `line.items()`, as `GlyphRunIter` currently appears to skip the run on empty lines.

parley/src/layout/cluster.rs
```diff
@@ -44,13 +44,17 @@ impl<'a, B: Brush> Cluster<'a, B> {
...
+                let advance = line.item(run_index).unwrap().advance;
+                let Some(run) = line.run(run_index) else {
+                    offset += advance;
```
DJMcNab, line 50: A comment that this handles inline boxes might be worthwhile.

parley/src/layout/cluster.rs
```diff
@@ -354,16 +358,15 @@ impl<'a, B: Brush> Cluster<'a, B> {
...
-            } else {
-                let visual_index = run.logical_to_visual(self.path.logical_index())?;
-                for cluster in run.visual_clusters().take(visual_index) {
-                    offset += cluster.advance();
-                }
-            }
+        for run_index in 0..self.path.run_index() {
+            let item = line.item(run_index)?;
+            offset += item.advance;
+        }
```
DJMcNab, line 364: It isn't completely clear what this is doing. If it were clearly just handling inline boxes, I'd understand it more clearly.

I feel like either this double counts, or needs a comment?

DJMcNab (approved):
Looks reasonable to me. Sorry this has taken so long to get review to. My confusion on visual_offset might block, but I generally trust your judgement here.

tomcur:
> Looks reasonable to me. Sorry this has taken so long to get review to.

Thank you! I believe this has been superseded by https://github.com/linebender/parley/pull/299. The example I gave in the original comment now seems to work.

PR #204, opened 2024-12-06
Stop using wildcard imports.

waywardmonkeys:


Philipp-M (approved):

parley/src/layout/line/greedy.rs
```diff
@@ -8,10 +8,13 @@ use swash::text::cluster::Whitespace;
 
 #[cfg(feature = "libm")]
 #[allow(unused_imports)]
-use core_maths::*;
+use core_maths::CoreFloat;
```
DJMcNab, line 11: `core_maths`' docs recommend using the previous style of import. I'm not sure why, as it does only export one thing

Philipp-M: Yeah doesn't make sense to me, AFAIK the only semantic difference would be that a wildcard import doesn't clash with a `CoreFloat` trait in the current module namespace (which doesn't exist...).

waywardmonkeys: I've seen other things do `use core_maths::CoreFloat as _;` or similar for their traits ...

PR #207, opened 2024-12-06
Misc. inline box layout fixes

nicoburns:
See individual commits:

- Fix whitespace collapsing when first item in a span is an inline box
    - `<p> label</p>` *should* have a space between the image and "label".
- Fix wrapping when layout contains only inline boxes
- Test case for layout with only inline boxes
- Prevent inline boxes from being duplicated when the line ends with an inline box.
- Fix trailing whitespace computation:
  - If the last item on a line is an inline box, then it does not have any trailing whitespace.


nicoburns:
> Is there something specific in here you want me to review - just want to check why my review has been requested? Unfortunately, this code isn't something I have any real understanding of

Not really. Just keen to get it reviewed so we can get it merged (I think this is the only thing blocking a 0.3 release). I thought you had reviewed similar PRs recently, and might be up for reviewing this one. But no worries if not!

DJMcNab (approved):
This can only impact users of the tree style builder and/or inline boxes, both of which I think you're the only user of currently. So since this is unlikely to be swiftly reviewed otherwise, I'm happy to unblock.

PR #209, opened 2024-12-07
Experiment: styled_text crate

waywardmonkeys:
This is an experimental PR that moves some stuff from `fontique` and `parley` into a new `styled_text` crate.  

This crate would get a rich text / attributed text or that could go in another crate.

Both `fontique` and `parley` depend on this new crate, but it only depends on `core_math` and is fully `no_std` compatible.

Nearly all of this should be so general that it could be used by other text systems as well.


waywardmonkeys:
I see that the font config code in fontique does something that will need adjustment ... and there's plenty of stuff that would need to be cleaned up with imports and such ...

fontique/src/backend/fontconfig/cache.rs
```diff
@@ -1,65 +1,62 @@
...
 
-impl Stretch {
-    fn from_fc(width: i32) -> Self {
-        match width {
-            63 => Self::EXTRA_CONDENSED,
-            87 => Self::SEMI_CONDENSED,
-            113 => Self::SEMI_EXPANDED,
-            _ => Self::from_ratio(width as f32 / 100.0),
-        }
+// FIXME(style): There's an argument for putting this into styled_text
```
waywardmonkeys, line 9: I like the idea of `styled_text` having this code as it makes it slightly friendlier from an interop perspective if others start to use it and it centralizes some of the understanding of the values.

waywardmonkeys:
Some things to resolve:

* [ ] Should `WhiteSpaceCollapse` move into `styled_text`?
* [ ] Should `TextStyle` move into `styled_text`?
* [ ] Fontique refers to `Stretch`, `Style`, and `Weight`, but Parley uses `FontStretch`, `FontStyle`, and `FontWeight`. We should choose one naming scheme and export it that way from `styled_text` and just use that.
* [ ] It would be nice to have a better way to distinguish between `styled_text::FontFeature` and `parley::style::FontFeature` (the latter is based on `swash` for internal usage within `parley`).
* [ ] Vendor the parse list code / API for `FontFeature` and `FontVariation` into `styled_text` and then use that from within Parley.
* [ ] This doesn't yet handle inline boxes, although that'll be more at the `RichText` / `RichTextBuilder` level once that arrives.
* [ ] There are still imports to clean up and harmonize.

waywardmonkeys:
I tink there's some interesting directions for this in the future as well.

Like `FontSize` should be an enum, not an `f32` (at this level) so that ti can handle the various things that CSS Fonts 4 does.

This also could grow some easier ways to configure the various OpenType features rather than requiring knowing the details of what the tags are. (This too can be inspired by CSS Fonts 4.)

It could be pretty exciting, but we need to make the first steps ...

nicoburns:
This looks interesting, but I think we ought to do a more conservative Parley 0.3 release before we land this.

waywardmonkeys:
I think that right now, I've avoided any breaking changes in the API ... but I definitely plan to have some, like dealing with `Weight` vs `FontWeight` and friends.  But those are easy... If there is agreement, this could easily land this week.

waywardmonkeys:
But I might also submit some of the breaking changes separately from this and then rebase this on top of them...

nicoburns:
> I think that right now, I've avoided any breaking changes in the API ... but I definitely plan to have some, like dealing with `Weight` vs `FontWeight` and friends. But those are easy... If there is agreement, this could easily land this week.

I'm less concerned about breaking changes and more concerned about whether this crate structure makes sense. I think we ought to give ourselves more time to iterate and settle on a design rather than trying to immediately rush it into a release along with other changes that have been stewing for some time.

PR #211, opened 2024-12-09
Rename `Stretch`, `Style`, `Weight` to have `Font` prefix

waywardmonkeys:
This brings them into line with how they're used in `parley::style` as well as closer to how they're named in CSS where these properties are `font-stretch` (since renamed to `font-width`), `font-style`, and `font-weight`.

This is a precursor to extracting this code into a separate crate for styling text.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,7 +24,11 @@ This release has an [MSRV] of 1.75.
 
 ### Changed
 
-### Parley
+#### Fontique
+
+- Breaking change: `Stretch`, `Style`, and `Weight` renamed to `FontStretch`, `FontStyle`, `FontWeight ([#211][] by [@waywardmonkeys][])
+
+#### Parley
 
 - Breaking change: `PlainEditor`'s semantics are no longer transactional ([#192][] by [@DJMcNab][])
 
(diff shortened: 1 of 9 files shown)


waywardmonkeys:
I'm sort of mixed on this, but:

* If we extract these to a separate crate (#209), then we'll have to agree on a common name for them.
* We already have 2 names for them within these 2 crates and it gets annoying.
* Masonry does a re-export of `parley::fontique::Weight as TextWeight`, but then goes on to use `TextWeight`, `FontWeight` (from `parley::style`) in doc comments, and also `Weight`, all in the same code.

So it would be nice to have a single name for these.

In this PR, I chose `Font*` because of the mapping to what the CSS Fonts specifications use.

nicoburns (approved):

waywardmonkeys:
https://github.com/linebender/xilem/pull/778 shows why having a single name in Parley and Fontique will be good (and why that name should be `FontWeight` and not `Weight`).

dfrg (approved):
> In this PR, I chose `Font*` because of the mapping to what the CSS Fonts specifications use.

Sgtm. I think we should generally try to mirror CSS naming moving forward.

waywardmonkeys:
> Sgtm. I think we should generally try to mirror CSS naming moving forward.

In a subsequent PR, I will suggest renaming `FontStretch` to `FontWidth` and see how we all feel about that.

PR #212, opened 2024-12-09
fontique: Move creation of attributes from fontconfig

waywardmonkeys:
This exposes `from_fontconfig` functions on `FontStretch`, `FontStyle`, and `FontWeight` rather than only defining them locally within the fontconfig code.

This is in preparation for a change similar to #209 where these types and others get pulled into a shared vocabulary crate and this sort of constructor helper would be useful to others.

### DIFF
--- a/fontique/src/attributes.rs
+++ b/fontique/src/attributes.rs
@@ -333,6 +333,8 @@ impl FontWeight {
     ///
     /// [fonts.conf documentation]: https://www.freedesktop.org/software/fontconfig/fontconfig-user.html
     pub fn from_fontconfig(weight: i32) -> Self {
+        // A selection of OpenType weights (first) and their corresponding fontconfig value (second)
+        // Invariant: The fontconfig values are sorted
         const MAP: &[(i32, i32)] = &[
             (0, 0),
             (100, 0),
--- a/fontique/src/attributes.rs
+++ b/fontique/src/attributes.rs
@@ -191,6 +191,28 @@ impl FontStretch {
     }
 }
 
+impl FontStretch {
+    /// Creates a new stretch attribute with the given value from Fontconfig.
+    ///
+    /// The values are determined based on the [fonts.conf documentation].
+    ///
+    /// [fonts.conf documentation]: https://www.freedesktop.org/software/fontconfig/fontconfig-user.html
+    pub fn from_fontconfig(width: i32) -> Self {
+        match width {
+            50 => Self::ULTRA_CONDENSED,
+            63 => Self::EXTRA_CONDENSED,
+            75 => Self::CONDENSED,
+            87 => Self::SEMI_CONDENSED,
+            100 => Self::NORMAL,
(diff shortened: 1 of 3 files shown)


xStrom:
Is there any use-case for these outside of Linux? They were previously only compiled on Linux.

waywardmonkeys:
Partly that the effort to restrict them is currently not correct (since this would be useful on FreeBSD and others as well). They're only 3 small functions, so not a big deal.

I wish we had better platform control over features.

fontique/src/attributes.rs
```diff
@@ -191,6 +191,28 @@ impl FontStretch {
...
+            50 => Self::ULTRA_CONDENSED,
+            63 => Self::EXTRA_CONDENSED,
+            75 => Self::CONDENSED,
+            87 => Self::SEMI_CONDENSED,
+            100 => Self::NORMAL,
+            113 => Self::SEMI_EXPANDED,
+            125 => Self::EXPANDED,
+            150 => Self::EXTRA_EXPANDED,
+            200 => Self::ULTRA_EXPANDED,
+            _ => Self::from_ratio(width as f32 / 100.0),
```
DJMcNab, line 211: If the constant values are different to this value, does that mean that there are discontinuities as you go up the stack? Is it just "hinting" to the more precise value when available?

fontique/src/attributes.rs
```diff
@@ -304,6 +326,46 @@ impl FontWeight {
...
 }
 
+impl FontWeight {
+    /// Creates a new weight attribute with the given value from Fontconfig.
+    ///
+    /// The values are determined based on the [fonts.conf documentation].
+    ///
+    /// [fonts.conf documentation]: https://www.freedesktop.org/software/fontconfig/fontconfig-user.html
+    pub fn from_fontconfig(weight: i32) -> Self {
+        const MAP: &[(i32, i32)] = &[
```
DJMcNab, line 336:
```suggestion
        // A selection of OpenType weights (first) and their corresponding fontconfig value (second)
        // Invariant: The fontconfig values are sorted
        const MAP: &[(i32, i32)] = &[
```

fontique/src/attributes.rs
```diff
@@ -304,6 +326,46 @@ impl FontWeight {
...
+            (700, 200),
+            (800, 205),
+            (900, 210),
+            (950, 215),
+        ];
+        for (i, (ot, fc)) in MAP.iter().skip(1).enumerate() {
+            if weight == *fc {
+                return Self::new(*ot as f32);
+            }
+            if weight < *fc {
```
DJMcNab, line 355:
```suggestion
            // Linear interpolation if not an exact match
            if weight < *fc {
```

DJMcNab (approved):

fontique/src/attributes.rs
waywardmonkeys: I think just to ensure precision

We should probably add a test (not now) that the formatting of the conversions comes out correctly.

PR #213, opened 2024-12-09
Impl bytemuck traits for `GenericFamily`

waywardmonkeys:
This can be used in the future as part of FFI work, but for now, the `Contiguous` trait is used to provide a `MAX_VALUE` used by `GenericFamilyMap`.

This is a backport of part of #209 where the separation between `GenericFamily` and `GenericFamilyMap` has grown to span crates.

The testing here borrows from (copies) what is done for similar testing of `bytemuck` traits in the `color` crate.

These crates already depend on `bytemuck`, so this isn't a new dependency. A decision is made here though to not use the `derive` feature as when we move this type to a vocabulary crate, we will want to avoid the `derive` feature there to avoid a dependency on proc macros.

### DIFF
--- a/fontique/src/generic.rs
+++ b/fontique/src/generic.rs
@@ -4,6 +4,7 @@
 //! Generic font families.
 
 use super::FamilyId;
+use bytemuck::{checked::CheckedBitPattern, Contiguous, NoUninit, Zeroable};
 use core::fmt;
 use smallvec::SmallVec;
 
@@ -50,6 +51,7 @@ pub enum GenericFamily {
     /// Song and cursive-style Kai forms. This style is often used for
     /// government documents.
     FangSong = 12,
+    // NOTICE: If a new value is added, be sure to add modify `MAX_VALUE` in the bytemuck impl.
 }
 
 impl GenericFamily {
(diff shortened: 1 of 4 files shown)


fontique/src/generic.rs
```diff
@@ -148,3 +173,78 @@ impl GenericFamilyMap {
...
+#[cfg(doctest)]
+/// Doctests aren't collected under `cfg(test)`; we can use `cfg(doctest)` instead
+mod doctests {
+    /// Validates that any new variants in `GenericFamily` has led to a change in the `Contiguous` impl.
+    /// Note that to test this robustly, we'd need 256 tests, which is impractical.
+    /// We make the assumption that all new variants will maintain contiguousness.
+    ///
+    /// ```compile_fail,E0080
+    /// use bytemuck::Contiguous;
+    /// use styled_text::GenericFamily;
```
DJMcNab, line 236: Hmm. I don't think that this is the right import path...

Indeed, this test fails on nightly for me.

DJMcNab (approved):
Other than the incorrect test, this looks good to me.

nicoburns (changes requested):
Could we feature flag the `bytemuck` impls? `bytemuck` is pretty small as a dependency. But it should also be easy to feature flag?

waywardmonkeys:
> Could we feature flag the `bytemuck` impls? `bytemuck` is pretty small as a dependency. But it should also be easy to feature flag?

Probably not.

It is already used by the font stack and the reason that I used it here is to provide a better implementation of something for `GenericFamilyMap`.

I did an earlier version where I had a const and used that for `GenericFamilyMap` and used it in the `Contiguous` impl as well which let it be optional, but it didn't seem worth the hassle.

We'll also use this once we look at FFIs (hopefully).

fontique/src/generic.rs
waywardmonkeys: Whoops, copy/paste from my branch. Fixed and force pushed.

nicoburns:
I'm a little uncomfortable with those unsafe impls. Specifically, it seems like it would be very easy for the MIN and MAX constants to get out of sync. I think we ought consider checking those in CI with a macro at some point.

But they do seem to be correct now.

waywardmonkeys:
They are tested in the PR already.

DJMcNab:
We should probably have a comment in `GenericFamily` about adding variants.

nicoburns (approved):
Tests seems to work in both directions (MAX too high and MAX too low). Nice.

PR #215, opened 2024-12-09
Fix `Cursor::{next,previous}_logical_word` and skip over whitespace

tomcur:
Currently, in `foo b|ar` where `|` indicates the cursor position, `Cursor::previous_logical_word` will return `foo| bar`. This doesn't match the behavior of `Cursor::{next_logical_word, previous_visual_word}` which place the cursor at the boundary of the current word. This happens because `Cluster::previous_logical_word` is called from the upstream cluster (the "b") and the cursor then lands between the "o" and the space.

This also renames "left", "right" -> "upstream", "downstream" (naming copied from `Cursor::logical_clusters`) to make it clear we're operating in logical order and not visual, and attaches the cursor to the word whose boundary it lands on.

The upstream/downstream fix on its own with no other changes, would regress how whitespace is handled, e.g., in `foo |bar`, where `previous_logical_word` would then end up at `foo| bar` instead of `|foo bar`. The cleanest way to handle this, I think, is to fold in a fix for #604 at the same time, skipping over whitespace. That's what this PR now proposes. This aligns these logical jumps to the visual `{previous, next}_visual_word`.

### DIFF
--- a/parley/src/editing/selection.rs
+++ b/parley/src/editing/selection.rs
@@ -59,7 +59,11 @@ impl Selection {
                 cluster = prev;
             }
             let anchor = Cursor::from_cluster(layout, cluster.clone(), !cluster.is_rtl());
-            let focus = anchor.next_logical_word(layout);
+            let focus = if let Some(end) = cluster.next_logical_word() {
+                Cursor::from_cluster(layout, end.clone(), !cluster.is_rtl())
+            } else {
+                Cursor::from_byte_index(layout, usize::MAX, Affinity::Downstream)
+            };
             Self {
                 anchor,
                 focus,
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@ This release has an [MSRV] of 1.88.
 #### Parley
 
 - Breaking change: the `Glyph::style_index` field was removed. Use `Cluster::{style, style_index}` or `GlyphRun::{style, style_index}` instead. (#661 by [@tomcur][])
+- `parley::editing::Cursor::{previous,next}_logical_word` now land at the previous/next logical start of a word and skip over whitespace. ([#215][] by [@tomcur][])
 
 ### Fixed
 
@@ -587,6 +588,7 @@ This release has an [MSRV][] of 1.70.
 [#211]: https://github.com/linebender/parley/pull/211
 [#212]: https://github.com/linebender/parley/pull/212
 [#213]: https://github.com/linebender/parley/pull/213
+[#215]: https://github.com/linebender/parley/pull/215
 [#223]: https://github.com/linebender/parley/pull/223
 [#224]: https://github.com/linebender/parley/pull/224
 [#241]: https://github.com/linebender/parley/pull/241
--- a/parley/src/editing/cursor.rs
+++ b/parley/src/editing/cursor.rs
@@ -243,16 +243,23 @@ impl Cursor {
         cur
     }
 
-    /// Returns a new cursor that is positioned at the next word boundary
-    /// in logical order.
+    /// Returns a new cursor that is positioned at the next word boundary in logical order, skipping
+    /// over whitespace.
     #[must_use]
     pub fn next_logical_word<B: Brush>(&self, layout: &Layout<B>) -> Self {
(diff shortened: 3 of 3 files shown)


parley/src/layout/cursor.rs
```diff
@@ -259,10 +260,11 @@ impl Cursor {
     /// in logical order.
     #[must_use]
     pub fn previous_logical_word<B: Brush>(&self, layout: &Layout<B>) -> Self {
-        let [left, right] = self.logical_clusters(layout);
-        if let Some(cluster) = left.or(right) {
+        let [upstream, downstream] = self.logical_clusters(layout);
+        if let Some(cluster) = downstream.or(upstream) {
```
tomcur, line 264: This is interesting as we're checking the same clusters (in logical order) as for the method in the other direction. I believe the asymmetry comes from the word boundary clusters being the logically first cluster of a word and (e.g.) the whitespace logically following the word, e.g., in "foobar baz" the boundary clusters of the first word are "f" and " ".

parley/src/layout/cursor.rs
```diff
@@ -243,14 +243,15 @@ impl Cursor {
...
-        if let Some(cluster) = right.or(left) {
+        let [upstream, downstream] = self.logical_clusters(layout);
+        if let Some(cluster) = downstream.or(upstream) {
             let start = cluster.clone();
             let cluster = cluster.next_logical_word().unwrap_or(cluster);
             if cluster.path == start.path {
                 return Self::from_byte_index(layout, usize::MAX, Affinity::Downstream);
             }
-            return Self::from_cluster(layout, cluster, true);
+            let moving_right = !cluster.is_rtl();
```
DJMcNab: `moving_right` in `from_cluster` calculates the Affinity from the cluster's rtl; to calculate it from the rtl originally is... interesting.

DJMcNab (approved):
Your bracketed note is also captured in https://github.com/linebender/parley/pull/604.

Otherwise seems reasonable.

parley/src/layout/cursor.rs
tomcur: Yeah... I've simplified this a bit by using the `Self::from_byte_index` directly with an explicit `Affinity`, rather than going through the apply-then-unapply directionality dance to get at an affinity. Both `next_logical_word` and `previous_logical_word` now affine downstream, which I think is correct, and I've added comments explaining the behavior (and at least, manually wiring these methods up in the editor to visually inspect their behavior without mutating text seems to behave as I'd expect (and on `main` only backdelete calls `prev_logical_word`, so it's also not clear whether the affinity really matters at all here)).

parley/src/editing/selection.rs
```diff
@@ -59,7 +59,11 @@ impl Selection {
                 cluster = prev;
             }
             let anchor = Cursor::from_cluster(layout, cluster.clone(), !cluster.is_rtl());
-            let focus = anchor.next_logical_word(layout);
+            let focus = if let Some(end) = cluster.next_logical_word() {
+                Cursor::from_cluster(layout, end.clone(), !cluster.is_rtl())
+            } else {
+                Cursor::from_byte_index(layout, usize::MAX, Affinity::Downstream)
+            };
```
tomcur, line 66: With the new whitespace skipping, double-click-to-select can't use `Cursor::next_logical_word` anymore. Note the `anchor` already used `Cluster::previous_logical_word`, as opposed to `focus` that was using `Cursor::next_logical_word` (note: **`Cluster`** vs. **`Cursor`** as the type!).

`Cluster` uses the Unicode word boundaries from analysis, whereas `Cursor` encodes "editor behavior." When selecting a word in an editor by double clicking, you do want to stop at whitespace. Hence, this should use `Cluster`.

tomcur:
This PR has changed a bit so probably needs re-review. The fix I originally proposed regressed whitespace handling. This now also includes a fix for #604, as I believe that's the only clean way to get this to behave as expected for both cases, e.g., `previous_logical_word` navigates as `foo b|ar` -> `foo |bar` -> `|foo bar`.

DJMcNab (approved):
To clarify my understanding, these methods now jump to the "logical start" of the next or previous logical word, right?

That seems as reasonable as anything, but we probably should document it?

PR #218, opened 2024-12-09
Add some tests for from_fontconfig

waywardmonkeys:


waywardmonkeys:
The ones for `FontWeight` have a couple of interpolation checks in them. And they all double as checking that string formatting works correctly.

DJMcNab (approved):

PR #219, opened 2024-12-09
Explain why `FontStretch`, `FontStyle`, `FontWeight` are `f32`

waywardmonkeys:
These are often `u16` in other libraries and it is worth noting why they are different here, especially as that calls out the support that we have for variable fonts and axes of variation.

### DIFF
--- a/fontique/src/attributes.rs
+++ b/fontique/src/attributes.rs
@@ -48,13 +48,20 @@ impl fmt::Display for Attributes {
 ///
 /// The default value is [`FontStretch::NORMAL`] or `1.0`.
 ///
-/// In variable fonts, this can be controlled with the `wdth` [axis].
+/// In variable fonts, this can be controlled with the `wdth` [axis]. This
+/// is an `f32` so that it can represent the same range of values as as
+/// the `wdth` axis.
+///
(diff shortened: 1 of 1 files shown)


nicoburns (approved):
It might be good to mention that it's the XXXX axis **for variable fonts**. Because otherwise people who are not familiar with variable fonts might not know what the "axis" refers to. But otherwise this seems good.

waywardmonkeys:
> It might be good to mention that it's the XXXX axis **for variable fonts**. Because otherwise people who are not familiar with variable fonts might not know what the "axis" refers to. But otherwise this seems good.

Check out just a bit above that in the comment!  :)

DJMcNab (approved):
Following on from that, have we considered merging these with the sentences mentioning the axes? As it stands, this feels a bit disjointed.

The comment about the weight classes could also be enhanced by mentioning that we snap to the closest weight class if the selected family isn't variable.

waywardmonkeys:
Updated the text some and also fixed so that it doesn't indicate that the `ital` axis needs an `f32`.

waywardmonkeys:
Not going to do the bit about the weight class snapping here, that's not really a property of this value (and if this moves to a vocabulary crate, then that's even more so).

PR #223, opened 2024-12-10
Rename `FontStretch` to `FontWidth`

waywardmonkeys:
This brings it inline with CSS Fonts 4. Various things with `stretch` in their name have also been renamed to use `width`.


xorgy:
I am pleased that a multi-line argument list is finally collapsed to a single line again. :+ )

PR #224, opened 2024-12-10
Add `PlainEditor` method to get IME cursor area

tomcur:
The area reported is the area of text the user is currently editing. It is usually used by platforms to place a candidate box for IME near it, while ensuring it is not obscured.

The implementation here ensures the area is usually near the focus, that the area contains at least some surrounding context (in case the platform places something to the side of the area, instead of above or under), and that the IME candidate box usually does not need to jump around when the IME starts or continues composing.

### DIFF
--- a/examples/vello_editor/src/main.rs
+++ b/examples/vello_editor/src/main.rs
@@ -8,16 +8,17 @@
 #![allow(clippy::shadow_unrelated)]
 #![allow(clippy::unseparated_literal_suffix)]
 
-use accesskit::{Node, Rect, Role, Tree, TreeUpdate};
+use accesskit::{Node, Role, Tree, TreeUpdate};
 use anyhow::Result;
 use std::num::NonZeroUsize;
 use std::sync::Arc;
+use vello::kurbo;
 use vello::peniko::Color;
 use vello::util::{RenderContext, RenderSurface};
 use vello::wgpu;
 use vello::{AaConfig, Renderer, RendererOptions, Scene};
 use winit::application::ApplicationHandler;
-use winit::dpi::LogicalSize;
+use winit::dpi::{LogicalSize, PhysicalPosition, PhysicalSize};
 use winit::event::{StartCause, WindowEvent};
 use winit::event_loop::{ActiveEventLoop, ControlFlow, EventLoop, EventLoopProxy};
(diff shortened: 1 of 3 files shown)


examples/vello_editor/src/main.rs
```diff
@@ -340,6 +358,7 @@ fn main() -> Result<()> {
         scene: Scene::new(),
         editor: text::Editor::new(text::LOREM),
         last_drawn_generation: Default::default(),
+        last_sent_ime_cursor_area: None,
```
DJMcNab, line 361: Maybe this should be an arbitrary definitely invalid rect? E.g. `0x0 at (0, 0)` or similar

examples/vello_editor/src/main.rs
```diff
@@ -215,6 +219,20 @@ impl ApplicationHandler<accesskit_winit::Event> for SimpleVelloApp<'_> {
...
         if self.last_drawn_generation != self.editor.generation() {
             render_state.window.request_redraw();
+            let area = self.editor.editor().ime_cursor_area();
+            if self.last_sent_ime_cursor_area != Some(area) {
+                self.last_sent_ime_cursor_area = Some(area);
+                // Note: on X11 `set_ime_cursor_area` may cause the exclusion area to be obscured
+                // until https://github.com/rust-windowing/winit/pull/3966 is in the Winit release
+                // used by this example.
+                render_state.window.set_ime_cursor_area(
+                    PhysicalPosition::new(
```
DJMcNab, line 229: Hmm, is physical right here?

DJMcNab (approved):
Looks good. I've not tested it

examples/vello_editor/src/main.rs
tomcur: I think it is, but I haven't tested it with a non-`1` display scale. The editor gets its width set as the physical width in https://github.com/linebender/parley/blob/85a824a5310781ef5f3b9568f95e0e613cdcc528/examples/vello_editor/src/main.rs#L228-L235

examples/vello_editor/src/main.rs
tomcur: It's now flattened and splatted with NaN, which should definitely never be returned.

PR #227, opened 2024-12-11
Create CursorTest harness

PoignardAzur:
This is a long-awaited feature, letting us write user-friendly tests of Parley's cursor behavior.

### DIFF
--- a/parley/src/cursor_test.rs
+++ b/parley/src/cursor_test.rs
@@ -1,0 +1,267 @@
+// Copyright 2024 the Parley Authors
+// SPDX-License-Identifier: Apache-2.0 OR MIT
+
+use crate::{Affinity, Cursor, FontContext, FontStack, Layout, LayoutContext};
+
+/// Helper struct for creating cursors and checking their values.
+///
+/// This type implements multiple assertion methods which, on failure, will
+/// print the input text with cursor's expected and actual positions highlighted.
+/// This should make test failures more readable than printing the cursor's byte index.
+///
+/// The following are not supported:
+///
+/// - RTL text.
+/// - Multi-line text.
+/// - Any character that doesn't span a single terminal tile.
+/// - Multi-bytes characters.
+///
+/// Some of these limitations are inherent to visually displaying a text layout in the
+/// terminal.
+///
+/// Others will be fixed in the future.
+pub struct CursorTest {
+    text: String,
+    layout: Layout<()>,
+}
+
+impl CursorTest {
+    pub fn single_line(text: &str, lcx: &mut LayoutContext<()>, fcx: &mut FontContext) -> Self {
+        let mut builder = lcx.ranged_builder(fcx, text, 1.0);
+        builder.push(
+            FontStack::Single(crate::FontFamily::Generic(
+                fontique::GenericFamily::Monospace,
+            )),
+            ..,
+        );
+        let mut layout = builder.build(text);
+        layout.break_all_lines(None);
+
+        Self {
+            text: text.to_string(),
+            layout,
+        }
+    }
+
+    /// Returns the text that was used to create the layout.
+    pub fn text(&self) -> &str {
+        &self.text
+    }
+
+    /// Returns the layout that was created from the text.
+    pub fn layout(&self) -> &Layout<()> {
+        &self.layout
+    }
+
+    #[track_caller]
+    fn get_unique_index(&self, method_name: &str, needle: &str) -> usize {
+        let Some(index) = self.text.find(needle) else {
+            panic!(
+                "Error in {method_name}: needle '{needle}' not found in text '{}'",
+                self.text
+            );
+        };
+        dbg!(index);
+        if self.text[index + needle.len()..].find(needle).is_some() {
+            panic!(
+                "Error in {method_name}: needle '{needle}' found multiple times in text '{}'",
+                self.text
+            );
+        }
+        index
+    }
+
+    /// Returns a cursor that points to the first character of the needle, with
+    /// [`Affinity::Downstream`].
+    ///
+    /// The needle must be unique in the text to avoid ambiguity.
+    ///
+    /// ### Panics
+    ///
+    /// - If the needle is not found in the text.
+    /// - If the needle is found multiple times in the text.
+    #[track_caller]
+    pub fn cursor_before(&self, needle: &str) -> Cursor {
+        let index = self.get_unique_index("cursor_before", needle);
+        Cursor::from_byte_index(&self.layout, index, Affinity::Downstream)
+    }
+
+    /// Returns a cursor that points to the first character after the needle, with
+    /// [`Affinity::Upstream`].
+    ///
+    /// The needle must be unique in the text to avoid ambiguity.
+    ///
+    /// ### Panics
+    ///
+    /// - If the needle is not found in the text.
+    /// - If the needle is found multiple times in the text.
+    #[track_caller]
+    pub fn cursor_after(&self, needle: &str) -> Cursor {
+        let index = self.get_unique_index("cursor_after", needle);
+        let index = index + needle.len();
+        Cursor::from_byte_index(&self.layout, index, Affinity::Upstream)
+    }
+
+    fn cursor_to_monospace(&self, cursor: Cursor, is_correct: bool) -> String {
+        fn check_no_color() -> bool {
+            let Some(env_var) = std::env::var_os("NO_COLOR") else {
+                return false;
+            };
+            let env_var = env_var.to_str().unwrap_or_default().trim();
+
+            if env_var == "0" {
+                return false;
+            }
+            if env_var.to_ascii_lowercase() == "false" {
+                return false;
+            }
+            true
+        }
+
+        // NOTE: The background color doesn't carry important information,
+        // so we do a simple implementation, without worrying about
+        // color-blindness and platform issues.
+        let ansi_bg_color = if cfg!(not(unix)) || check_no_color() {
+            ""
+        } else if is_correct {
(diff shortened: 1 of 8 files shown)


parley/src/cursor_test.rs
DJMcNab, line 1: It's unclear why this isn't in `src/tests/utils`

tomcur (commented):
Really good to see this progress. I think the main concern raised in [the Zulip thread](https://xi.zulipchat.com/#narrow/channel/205635-text/topic/Testing.20API) is whether terminal output can work well enough for all the errors that may need to be reported.

One possibility is to just debug-print the actual and expected cursors.

I think spanning selections should also be in scope for testing, but that can be done later.

PoignardAzur:
> One possibility is to just debug-print the actual and expected cursors for now.

The PR already does.

PoignardAzur:
image

parley/src/cursor_test.rs
```diff
@@ -0,0 +1,267 @@
...
+    text: String,
+    layout: Layout<()>,
+}
+
+impl CursorTest {
+    pub fn single_line(text: &str, lcx: &mut LayoutContext<()>, fcx: &mut FontContext) -> Self {
+        let mut builder = lcx.ranged_builder(fcx, text, 1.0);
+        builder.push(
+            FontStack::Single(crate::FontFamily::Generic(
+                fontique::GenericFamily::Monospace,
```
DJMcNab, line 33: I don't think that this should impact anything?

parley/src/cursor_test.rs
```diff
@@ -0,0 +1,270 @@
...
+        let index = self.get_unique_index("cursor_before", needle);
+        Cursor::from_byte_index(&self.layout, index, Affinity::Downstream)
+    }
+
+    /// Returns a cursor that points to the first character after the needle, with
+    /// [`Affinity::Upstream`].
+    ///
+    /// The needle must be unique in the text to avoid ambiguity.
+    ///
+    /// ### Panics
```
DJMcNab, line 94: [Clippy suggests](https://rust-lang.github.io/rust-clippy/master/index.html#missing_panics_doc) one `#`

tomcur:
I mean debug-printing the cursor structs to the terminal: printing a report to the terminal as in the screenshot may not be possible for all the layouts that are interesting to test selection for (RTL text, very long lines, line breaks). A report like that _is_ very useful to have though, perhaps something very similar could be rendered to an image.

parley/src/cursor_test.rs
```diff
@@ -0,0 +1,270 @@
...
+    #[track_caller]
+    fn cursor_assertion(&self, expected: Cursor, actual: Cursor) {
+        if expected == actual {
+            return;
+        }
+
+        // TODO - Check that the tested string doesn't include difficult
+        // characters (newlines, tabs, RTL text, etc.)
+        // If it does, we should still print the text on a best effort basis, but
+        // without visual cursors and with a warning that the text may not be accurate.
```
DJMcNab, line 165: As discussed on Zulip, the alternative/fallback is to do screenshots. I'd suggest that we should have screenshots unconditionally if this fails, and show the in-terminal version if there is no RTL region (we should strive to handle newlines properly, and we can't possibly detect non-tab extra-wide terminal characters so it's not worth trying)

This doesn't need to happen in this PR, but it might be worth leaving these breakcrumbs in the code

parley/src/cursor_test.rs
```diff
@@ -0,0 +1,270 @@
...
+    #[track_caller]
+    pub fn assert_cursor_is(&self, expected: Cursor, cursor: Cursor) {
+        self.cursor_assertion(expected, cursor);
+    }
+
+    /// Prints the TestLayout's text, with the cursor highlighted.
+    ///
+    /// Uses the same format as assertion failures.
+    #[track_caller]
+    pub fn print_cursor(&self, cursor: Cursor) {
```
DJMcNab, line 237: Hmm, this is a bit harder to have with screenshots for debugging. I suppose we can't know if a test will fail until it fails. We could queue these up?

Not a problem for this PR to solve.

parley/src/cursor_test.rs
```diff
@@ -0,0 +1,270 @@
...
+
+// ---
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn cursor_next_visual() {
+        let (mut lcx, mut fcx) = (LayoutContext::new(), FontContext::new());
```
DJMcNab, line 260: Ideally this would use the build-in Roboto font, but of course it doesn't really matter what font is chosen, so long as it is not fundamentally misbehaving

parley/src/lib.rs
```diff
@@ -124,6 +125,7 @@ pub use peniko::Font;
 
 pub use builder::{RangedBuilder, TreeBuilder};
 pub use context::LayoutContext;
+pub use cursor_test::CursorTest;
```
DJMcNab, line 128: What is the advantage of making this public?

parley/src/cursor_test.rs
DJMcNab: Actually of course it matters, for "up/down" button pressing test cases

parley/src/cursor_test.rs
PoignardAzur: I'd like to export it for external users making their own tests. `src/tests` is for internal testing only.

parley/src/lib.rs
PoignardAzur: For writing tests for something like the text editor widget in Masonry. We can always stop exporting it in future versions if we find nobody uses it.

parley/src/cursor_test.rs
PoignardAzur: I think we shouldn't worry about up-down testing for now, until we have a UI for tests with multi-line strings.

Like, yes, in theory we could just stick to comparing the numeric values of cursors, but if we want to do that we don't need this PR at all, just assert_eq.

parley/src/lib.rs
DJMcNab: I can't imagine a user story which would want this exported in its current form, but I can imagine that some of this code would be useful if made more modular.

I do think that the publicly exported "get the debug print of a cursor from the text and its corresponding layout" could be useful (where the layout is used only to find the soft line breaks in the text for debug printing) - caveated of course that it's a best effort. However, if we were actually doing that, we'd probably also want to properly use unicode_width, which I'm not sure we want to do. As it currently stands, this functionality is also tied in with the constructor creating the layout, which is pretty far from ideal.

The "quickly create a cursor from a needle in a string" is also useful to export publicly for people testing e.g. rich text, but again I think it's conflated with the other stuff.

PoignardAzur:
The CI is blowing up. Not sure where these errors are coming from.

PoignardAzur:
To re-iterate what was mentioned in Office Hours:

The purpose of this PR is to be able to write tests for cursor values where the intent of the test can be understood by reading the code of the test alone. This is in contrast to tests where you compare the cursor to a numeric value (the reader needs to visualize what position the number represents) and screenshot tests (the reader needs to open the screenshot file).

In future versions of this API, we'll probably dump screenshots directly to the console where possible, or other similar tricks.

The important part is that in any tests we write, readers should understand the intent of the test form the code alone.

DJMcNab:
> The purpose of this PR is to be able to write tests for cursor values where the intent of the test can be understood by reading the code of the test alone.

Again, I agree entirely that we should have this, and I agree that the "needle" based API achieves this very cleverly and unambiguously. But I still don't believe that this infrastructure as provided can possibly be useful for external users, and so we shouldn't export it.
I'd happily approve as-is if this new code were in the test utils module or otherwise `cfg(test)`ed.

PoignardAzur:
> But I still don't believe that this infrastructure as provided can possibly be useful for external users, and so we shouldn't export it.

Do you think there's a downside to exporting it, besides potential confusion?

The obvious downside would be that people might end up depending on it and be upset if we remove it later, but if you think this will never be useful to external users, then that doesn't really apply.

I'd be in favor of exporting it, *maybe* with a feature flag or a `#[doc(hidden)]` attribute, and if 6 months later we find out nobody uses it we can make it private again.

DJMcNab:
I strongly agree that some of these features would be useful for someone creating their own editor or similar on top of Parley's primitives.  Using a needle to create a cursor is brilliant, nicely simple code but very clear.

However, I'm still really struggling to imagine a user story for someone who would use what you are proposing we export - as this only exposes those features in a very straitjacketed way. I feel that I must be missing something fundamental here.
As far as I can tell, the only possible tests which can be written using this API are for moving a (single) cursor in different ways in entirely plain text - currently on one line, but in the future across many lines. However, the implementations of all the reasonable operations you would want to have (and if we've missed any, it's our bug) live in this crate, and so should be tested here, not by our external users.

My earlier conditional approval still stands. I'm not going to block this if someone else approves.

PoignardAzur:
I've made the implementation private again, and added some TODOs for future reference.

parley/src/tests/mod.rs
```diff
@@ -1,6 +1,9 @@
 // Copyright 2024 the Parley Authors
 // SPDX-License-Identifier: Apache-2.0 OR MIT
 
+// TODO - Figure out which tests can run without std
```
DJMcNab, line 4: Hmm, why do we want this? Afaik, we only enable `no_std` when not running tests.

DJMcNab (approved):

parley/src/tests/mod.rs
nicoburns: IIRC, the test runner itself requires std so tests are always run with std regardless of whether you enable it anyway.

PR #231, opened 2024-12-16
Update to git Vello, new releases of Color, Peniko

waywardmonkeys:


waywardmonkeys:
@DJMcNab I think someone should separate the MSRV update out of this and handle the noisy changes from that separately ... I've started with cleaning up some stuff as separate PRs (#232, #233).

DJMcNab (commented):
I don't really see the need to split them out, as the actual changes are trivial (oh I see, it's allow attributes; I'd be happy to just allow that temporarily). I am however a little hesitant to actually *merge* this upgrade here; I think we should treat it as experiment.

In my mind, the git dependency on Color in Peniko/Vello is useful, because that's needed for developing Vello further. But this upgrade is largely peripheral, only in proving out the upgrade.

That is, my gut reaction is to make this into a draft PR for now (and have a corresponding draft PR to Xilem)

If you need the corresponding PR in Xilem, let me know and I'll make it today.

waywardmonkeys:
I have a corresponding branch in xilem already, just not pushed yet, I think.

waywardmonkeys:
@DJMcNab Now that Color and Peniko are released, I'd like to get this in here so that we're getting closer to what we need for the upcoming Parley release (and once this is in here, I'll do the same for Xilem).

DJMcNab (approved):

DJMcNab:
The PR title might need to change.

waywardmonkeys:
Ah, I'd changed the title in the commit but not in the PR.

waywardmonkeys:
@DJMcNab Updated!

PR #232, opened 2024-12-17
Fix and un-allow `trivial_numeric_casts` lints

waywardmonkeys:


nicoburns (approved):

nicoburns:
I'm less sure of this one. Seems like it could break on some platforms where types are defined differently? Hopefully CI would catch that though.

PR #238, opened 2025-01-03
Update to swash 0.2, same skrifa, read-fonts

waywardmonkeys:
This doesn't yet address `no_std` support in Parley.

The differences in the screenshot testing appear to be trivial.


dfrg (approved):
LGTM, thanks!

Issue #86, opened 2024-07-21
Sort out no_std support


nicoburns:
Currently `parley` has a `std` feature which can be disabled, and we are paying the maintenance cost of maintaining that, but it can't actually do anything useful when that feature is disabled. I would suggest that:

- Nothing in this repo ought to be support in "no alloc" configurations
- Fontique ought to work as "font database" and support querying, manual loading of fonts, etc without std. Backends should require std as needed (probably all of them?).
- Parley ought to to fully function in "alloc but no std" mode. It ought to have no conditional code to enable this (except perhaps some conditional imports from the alloc crate). Parley will still transitively require std in some cases through it's dependencies.

waywardmonkeys:
See problem exposed in PR #124.

xStrom:
I think a good first step is to get things actually compiling with `no_std`. That will allow us to at least not regress thanks to CI checks. Adding actually useful functionality can be a step after that.

With that goal, I managed to get `fontique` compiling in #158.

Now `parley` is more tricky, because `swash` does not yet support `no_std`. I started to add that support, but then found that @waywardmonkeys is already ahead of me with [swash#63](https://github.com/dfrg/swash/pull/63). Once that lands we can move forward with doing the same for `parley`.

DJMcNab:
What is the state of this now? Is it only waiting on a release of swash containing dfrg/swash#63?

waywardmonkeys:
That is at least the next step!

PR #239, opened 2025-01-03
Test `no_std` in CI

waywardmonkeys:
This builds on #238 and contains the same commit.

This has a change to only run `no_std` checks on the libraries since examples and such aren't likely to be `no_std`, nor do they need to be.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -112,7 +112,7 @@ jobs:
           save-if: ${{ github.event_name != 'merge_group' }}
 
       - name: cargo clippy (no_std)
-        run: cargo hack clippy --workspace --locked --optional-deps --each-feature --ignore-unknown-features --features libm --exclude-features ${{ env.FEATURES_DEPENDING_ON_STD }} --target x86_64-unknown-none -- -D warnings
+        run: cargo hack clippy --workspace --locked --optional-deps --each-feature --ignore-unknown-features --features libm --lib --exclude-features ${{ env.FEATURES_DEPENDING_ON_STD }} --target x86_64-unknown-none -- -D warnings
 
       - name: cargo clippy
         run: cargo hack clippy --workspace --locked --optional-deps --each-feature --ignore-unknown-features --features std -- -D warnings
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -112,7 +112,7 @@ jobs:
           save-if: ${{ github.event_name != 'merge_group' }}
 
       - name: cargo clippy (no_std)
-        run: cargo hack clippy --workspace --locked --optional-deps --each-feature --ignore-unknown-features --features libm --lib --exclude-features ${{ env.FEATURES_DEPENDING_ON_STD }} --target x86_64-unknown-none -- -D warnings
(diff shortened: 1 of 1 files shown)


waywardmonkeys:
so, `--lib` doesn't work here. What's the better way to deal with this?

.github/workflows/ci.yml
```diff
@@ -111,10 +111,8 @@ jobs:
         with:
           save-if: ${{ github.event_name != 'merge_group' }}
 
-      # TODO: Add --target x86_64-unknown-none to the no_std check once we solve the compilation issues with it.
-      #       https://github.com/linebender/parley/issues/86
```
DJMcNab, line 115: What's the status of https://github.com/linebender/parley/issues/86 once this PR is merged? I don't think this is currently marked as fixing that issue.

.github/workflows/ci.yml
```diff
@@ -111,10 +111,8 @@ jobs:
...
       - name: cargo clippy (no_std)
-        run: cargo hack clippy --workspace --locked --optional-deps --each-feature --ignore-unknown-features --features libm --exclude-features ${{ env.FEATURES_DEPENDING_ON_STD }} -- -D warnings
+        run: cargo hack clippy ${{ env.RUST_MIN_VER_PKGS }} --locked --optional-deps --each-feature --ignore-unknown-features --features libm --exclude-features ${{ env.FEATURES_DEPENDING_ON_STD }} --target x86_64-unknown-none -- -D warnings
```
DJMcNab, line 115: Hmm. I *think* the idiomatic approach here would be to add a new environment variable? cc @xStrom

DJMcNab (approved):
I'm definitely happy to merge with a new variable added. Happy to alternatively see what Kaur thinks

.github/workflows/ci.yml
xStrom: An even better solution is a brand new `cargo-hack` feature `--must-have-and-exclude-feature std`. Unfortunately it's been stuck in review queue for a month at [cargo-hack#262](https://github.com/taiki-e/cargo-hack/pull/262).

For now I think it's fine to abuse `${{ env.RUST_MIN_VER_PKGS }}` and just get this landed. When the proper solution becomes possible we'll adjust it.

xStrom (approved):
Yes this contains a hack in the form of abusing `${{ env.RUST_MIN_VER_PKGS }}`, but the new state is still much better than the old one so let's land this.

waywardmonkeys:
Until I got distracted by work, I was going to suggest that `RUST_MIN_VER_PKGS` could just as easily be `RUST_PUBLISHED_PKGS` and then the name is fine for both usages and then we could apply that across all repos.

xStrom:
One issue is that we could introduce published packages that don't support `no_std`. [cargo-hack#262](https://github.com/taiki-e/cargo-hack/pull/262) wouldn't have any problem with that.

So my recommendation is to still just merge this as-is to get things moving, and the eventual standard change would be the more robust `--must-have-and-exclude-feature std`.

.github/workflows/ci.yml
waywardmonkeys: Needs to be reviewed!

Issue #240, opened 2025-01-07
Alignment of long text without word wrap


jamesthurley:
As per the Zulip discussion [here](https://xi.zulipchat.com/#narrow/channel/205635-text/topic/Alignment.20in.20Parley.20without.20word.20wrap/near/492067601).

My code currently looks like this:
```rust
    let max_advance = if text_context.wrap {
        width_constraint
    } else {
        Some(f32::MAX)
    };
    layout.break_all_lines(max_advance);
    layout.align(
        width_constraint,
        match text_context.horizontal_alignment {
            HorizontalAlignment::Center => Alignment::Middle,
            HorizontalAlignment::Right => Alignment::End,
            HorizontalAlignment::Left => Alignment::Start,
        },
    );
```
When word wrap is off I've found I still need to call `break_all_lines` and I still need to supply a `max_advance`, so I set it to `f32::MAX`. This is fine.

This code is all working well, except for the cases where word wrap is off, the text is longer than the width constraint, and the text alignment is Middle or End.

The result is as follows:

![image](https://github.com/user-attachments/assets/e4cb0d50-38fc-4788-99f8-6dd4483876e3)

There are 9 tests here. First three left aligned, then right aligned, then centered. Within each set of three tests the middle test case is the problematic one - you can see in all three cases it gets left-aligned.

The expected behaviour was that the left side of the right-aligned text would be cropped, and both sides of the centered text would be cropped. Although with further thought I can see that chopping off the start of the text may not always be desirable.

After the discussion on Zulip it was decided I should create a GitHub issue.

PR #241, opened 2025-01-17
Make alignment when free space is negative configurable

nicoburns:
Untested, but probably fixes https://github.com/linebender/parley/issues/240

### DIFF
--- a/examples/tiny_skia_render/src/main.rs
+++ b/examples/tiny_skia_render/src/main.rs
@@ -96,7 +96,11 @@ fn main() {
 
     // Perform layout (including bidi resolution and shaping) with start alignment
     layout.break_all_lines(max_advance);
-    layout.align(max_advance, Alignment::Start);
+    layout.align(
+        max_advance,
+        Alignment::Start,
+        false, /* align_when_overflowing */
+    );
     let width = layout.width().ceil() as u32;
     let height = layout.height().ceil() as u32;
     let padded_width = width + padding * 2;
--- a/parley/src/layout/alignment.rs
+++ b/parley/src/layout/alignment.rs
@@ -27,11 +27,6 @@ pub(crate) fn align<B: Brush>(
         // Compute free space.
         let free_space = alignment_width - line.metrics.advance + line.metrics.trailing_whitespace;
 
-        // Alignment only applies if free_space is non-zero
-        if free_space == 0.0 {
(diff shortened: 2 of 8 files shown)


parley/src/layout/alignment.rs
```diff
@@ -27,8 +27,8 @@ pub(crate) fn align<B: Brush>(
         // Compute free space.
         let free_space = alignment_width - line.metrics.advance + line.metrics.trailing_whitespace;
 
-        // Alignment only applies if free_space > 0
-        if free_space <= 0. {
+        // Alignment only applies if free_space is non-zero
+        if free_space == 0.0 {
             continue;
         }
```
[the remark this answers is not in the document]
nicoburns: I suppose it could. I was going to say that it was the `Justified` alignment I was concerned about. But that has it's own check anyway, so this one probably isn't necessary.

examples/tiny_skia_render/src/main.rs
```diff
@@ -96,7 +96,11 @@ fn main() {
 
     // Perform layout (including bidi resolution and shaping) with start alignment
     layout.break_all_lines(max_advance);
-    layout.align(max_advance, Alignment::Start);
+    layout.align(
+        max_advance,
+        Alignment::Start,
+        false, /* align_when_overflowing */
```
tomcur, line 102: If `align` gains more options in the future, a boolean may be a bit confusing, especially if people don't document as you've done here. Maybe an `AlignmentOptions` struct could be introduced?

tomcur, line 102: If `align` gains more options in the future, a boolean may be a bit confusing, especially if people don't document as you've done here. Maybe an `AlignmentOptions` struct could be introduced?

tomcur (approved):
This looks good to me, but see my note about the boolean option. Perhaps we want one more person to agree this is sensible to have.

A quick test of single words that overflow the width in `vello_editor` works as expected.

examples/tiny_skia_render/src/main.rs
nicoburns: Agree, but don't have time to work on this now. In the interest of getting this landed for Parley 0.3 I'm going to merge this as-is. And have created an issue for the options struct (https://github.com/linebender/parley/issues/247)

PR #243, opened 2025-01-20
Use Vello 0.4 release

waywardmonkeys:
Update some other related dependencies and APIs as well.


waywardmonkeys:
This is an example of why I have absolutely nothing nice to say about using image-based tests for editing operations.

PR #244, opened 2025-01-21
Add screenshot feature in CursorTest

PoignardAzur:


PoignardAzur:
I've read your PR a few times and I still don't understand how it works, sorry. I'll leave CI uploads as a future feature.

Keep in mind that because this isn't snapshot testing, the screenshots are dumped to a TMP file by default. I think you would want to change that default based on some env variable.

PR #245, opened 2025-01-21
Allow Bidi `base_level` to be determined from text

tomcur:
`Layout::is_rtl` always returns `false` when `base_level` is forced to 0. Can it already safely be determined from the string buffer? It seems to work OK with this change: the fist strong Bidi character encountered determines the `base_level`.

### DIFF
--- a/parley/src/context.rs
+++ b/parley/src/context.rs
@@ -109,7 +109,7 @@ impl<B: Brush> LayoutContext<B> {
             self.bidi.resolve(
                 text.chars()
                     .zip(self.info.iter().map(|info| info.0.bidi_class())),
-                Some(0),
+                None,
             );
         }
     }


nicoburns:
I think for web rendering I will need the ability to set the base level manually to (i.e. to represent the HTML`dir` attribute). I'll also need this at the "span" level, I'm not quite sure how that interacts will automatically detection.

For example, if you have:

```html
<p dir="rtl">Hello world<span dir="rtl">Foo Bar Baz</span></p>
```

where the `dir` attributes are all `rtl` but the text is all `ltr`, then I'm not quite sure how that ought to work.

tomcur:
> I think for web rendering I will need the ability to set the base level manually to (i.e. to represent the HTML `dir` attribute).

Yes, I think so too. I wanted to try wiring that up as well, but am trying to understand more of Parley's internals first, so started small.

I think in that case the `dir` on the outer paragraph would force `base_level` to `Some(1)`, and it probably should just affect alignment and weak-bidi characters. How the internal span would interact with this, and how to represent that in Parley, I'm not yet sure.

There are also the stronger `<bdo>` that actually changes visual letter ordering, and `<bdi>`.

nicoburns (approved):
This makes sense to me, with the caveat that I do not know all that much about bidi.

tomcur:
Let's merge, and we'll see if anything breaks.

PR #249, opened 2025-01-22
Fix linebreaking for lines without text runs

### LINKED ISSUE #246: Incorrect line breaking with line containing just an inline box
I wrote the following test.
```rust
#[test]
fn full_width_inbox() {
    let mut env = testenv!();

    let text = "AB";
        id: 0,
        index: 1,
        width: 100.0,
        height: 10.0,
    });
    layout.break_all_lines(Some(100.));
    layout.align(None, Alignment::Start);
}
```

I expected the result to look like the following.

Image

But it actually looks like this.

Image

wfdewith:
It turns out that there were actually two bugs that caused the behavior as described in #246.

The first bug happens when a line ends with a text run. In this scenario, the text run is repeated at the start of the next line, so that any remaining clusters can be rendered. However, when there are no more remaining clusters to render, the text run essentially uses clusters from the next run. In the example image in #246, this causes the 'B' to be added to the second line.

The second bug happens when a line that does not contain any text runs is committed. In this case, the cluster range in the current line state is bumped, even though no clusters were actually processed. This is the reason that the 'B' wasn't added to the third line in the image in #246.

I've added a test that tests a scenario with a full width inline box both when the previous line ends with an inline box and when it ends with a text run.

I'm not fully satisfied with the solution in this PR, because it feels rather heavy-handed. It is already unclear to me when both existing conditions in the invalid cluster check even apply, so just adding another one makes it even harder to understand.

Fixes #246

### DIFF
--- a/parley/src/layout/line/greedy.rs
+++ b/parley/src/layout/line/greedy.rs
@@ -696,8 +696,6 @@ fn try_commit_line<B: Brush>(
     alignment: Alignment,
     break_reason: BreakReason,
 ) -> bool {
-    let is_empty = layout.data.text_len == 0;
-
     // Ensure that the cluster and item endpoints are within range
     state.clusters.end = state.clusters.end.min(layout.data.clusters.len());
     state.items.end = state.items.end.min(layout.data.items.len());
@@ -760,12 +758,7 @@ fn try_commit_line<B: Brush>(
                     cluster_range.end = state.clusters.end;
                 }
 
-                let overlap = run_data.cluster_range.contains(&cluster_range.start)
-                    && run_data.cluster_range.contains(&(cluster_range.end - 1));
-                if cluster_range.start > cluster_range.end
-                    || (!is_empty && cluster_range.start == cluster_range.end)
-                    || !overlap
-                {
+                if cluster_range.start >= run_data.cluster_range.end {
                     // println!("INVALID CLUSTER");
                     // dbg!(&run_data.text_range);
                     // dbg!(cluster_range);
--- a/parley/src/layout/line/greedy.rs
+++ b/parley/src/layout/line/greedy.rs
@@ -828,7 +828,6 @@ fn try_commit_line<B: Brush>(
     state.num_spaces = 0;
     if committed_text_run {
         state.clusters.start = state.clusters.end;
-        state.clusters.end += 1;
     }
 
     state.items.start = match last_item_kind {
--- a/parley/src/layout/line/greedy.rs
+++ b/parley/src/layout/line/greedy.rs
@@ -724,6 +724,7 @@ fn try_commit_line<B: Brush>(
     // Iterate over the items to commit
     // println!("\nCOMMIT LINE");
     let mut last_item_kind = LayoutItemKind::TextRun;
+    let mut committed_text_run = false;
     for (i, item) in items_to_commit.iter().enumerate() {
         // println!("i = {} index = {} {:?}", i, item.index, item.kind);
 
@@ -759,8 +760,11 @@ fn try_commit_line<B: Brush>(
                     cluster_range.end = state.clusters.end;
                 }
 
+                let overlap = run_data.cluster_range.contains(&cluster_range.start)
+                    && run_data.cluster_range.contains(&(cluster_range.end - 1));
                 if cluster_range.start > cluster_range.end
                     || (!is_empty && cluster_range.start == cluster_range.end)
+                    || !overlap
                 {
                     // println!("INVALID CLUSTER");
                     // dbg!(&run_data.text_range);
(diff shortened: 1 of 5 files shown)


parley/src/layout/line/greedy.rs
```diff
@@ -759,8 +760,11 @@ fn try_commit_line<B: Brush>(
                     cluster_range.end = state.clusters.end;
                 }
 
+                let overlap = run_data.cluster_range.contains(&cluster_range.start)
+                    && run_data.cluster_range.contains(&(cluster_range.end - 1));
                 if cluster_range.start > cluster_range.end
                     || (!is_empty && cluster_range.start == cluster_range.end)
+                    || !overlap
```
tomcur, line 767: Could this be simplified to the following?

```suggestion
                if cluster_range.start >= run_data.cluster_range.end
                    || (!is_empty && cluster_range.start == cluster_range.end)
```

The idea is that if our next cluster index (`state.clusters.start`) is outside of this run, we've already fully consumed the run. I think where it goes wrong with this bug is that `cluster_range.end` is also updated in the case where there's a single inline box on this line.

tomcur (commented):
This works for me, and I think I agree with the solution. I have found a potential simplification and suggested it inline; I'm still wrapping my head around this line breaking code though.

parley/src/layout/line/greedy.rs
```diff
@@ -759,8 +760,11 @@ fn try_commit_line<B: Brush>(
...
                 {
```
tomcur, line 768: Could this be simplified to the following?

```suggestion
                if cluster_range.start >= run_data.cluster_range.end {
```

The idea is that if our next cluster index (`state.clusters.start`) is outside of this run, we've already fully consumed the run. I think where it goes wrong with this bug is that `cluster_range.end` is also updated, not only `cluster_range.start`, because when there's a single inline box on this line, this `TextRun` is both the first and the last of this line.

nicoburns:
> The first bug happens when a line ends with a text run. In this scenario, the text run is repeated at the start of the next line, so that any remaining clusters can be rendered. However, when there are no more remaining clusters to render, the text run essentially uses clusters from the next run.
 
 This makes sense, and the "committed_text_run" fix also makes sense to me.
 
 > It is already unclear to me when both existing conditions in the invalid cluster check even apply, so just adding another one makes it even harder to understand.
 
Totally agree with this. I think it should be possible to refactor things so that invalid clusters aren't possible. For example, when committing a line, we could eagerly check whether we've consumed all clusters in the last run of that line, and only repeat the run on the next line if there are unconsumed clusters remaining in that run. I've had a quick go at this, but I was unable to get it to work in the time I had available to work on it.

parley/src/layout/line/greedy.rs
wfdewith: Your suggestion seems to work and the reasoning is sound, but I'm still hesitant to remove anything until I know why it was there in the first place.

So far, I figured that `(!is_empty && cluster_range.start == cluster_range.end)` is probably for mandatory line breaks, where it avoids adding an empty text run to a line. Removing it breaks the editor tests, because the cursor will be rendered in an incorrect spot, but the text is still rendered correctly.

I assume that `cluster_range.start > cluster_range.end` is an incomplete check for the bug fully fixed in this PR, but I'm not sure yet.

wfdewith:
> Totally agree with this. I think it should be possible to refactor things so that invalid clusters aren't possible. For example, when committing a line, we could eagerly check whether we've consumed all clusters in the last run of that line, and only repeat the run on the next line if there are unconsumed clusters remaining in that run. I've had a quick go at this, but I was unable to get it to work in the time I had available to work on it.

Hmm, this gives me some ideas. I'll give it a shot to see if it clears things up.

parley/src/layout/line/greedy.rs
wfdewith: I'm fairly convinced that your idea is correct, so I've committed it. As far as I can see, there is only one way to have a run with a cluster range that falls outside the line state cluster range, and that is the scenario you've described above. The `cluster_range.start >= run_data.cluster_range.end` condition is just a cleaner way to check for all possible variants of the same problem.

wfdewith:
I've also removed the update of the end of the line cluster range after the line has been committed. It is not necessary since `BreakerState::append_cluster_to_line` will always update it based on the current `cluster_idx`. If there is any bug that causes it to not update properly, I'd rather have a more visible crash where `start > end`.

tomcur (approved):
Looks good to me, thanks a lot for diving into this! I agree that bumping `state.clusters.end` seems unnecessary.

tomcur:
Because test snapshots would quickly grow the repository's size, we're thinking to set up git LFS before landing this. That can probably be done tomorrow.

wfdewith:
> Because test snapshots would quickly grow the repository's size, we're thinking to set up git LFS before landing this. That can probably be done tomorrow.

Sure, I'll update the PR once that's done to put the snapshots in LFS.

wfdewith:
Snapshots should be in LFS now.

tomcur:
Thank you!

PR #250, opened 2025-01-23
Align based on text base direction

tomcur:
Add Alignment::{Left, Right} for bidi-independent alignment.

parley/tests/assets/Roboto-Regular.ttf is missing the Arabic alphabet, so the tests don't render correctly currently, and I've marked this as draft until that's fixed. I also notice the full-stop is hanging in End alignment. I'm not sure that's intentional.


tomcur:
Rebased on #256, which fixed hanging cluster issues in the LTR text.

wfdewith:
Did you intend for fdec767f62db3918b290bc1c8a702a9703c33eb9, which is a separate PR in #258 to be part of this PR as well?

tomcur:
I intended to make that a separate PR, but I see I accidentally based this PR on top of that.

parley/src/layout/mod.rs
```diff
@@ -41,10 +41,18 @@ pub use run::RunMetrics;
...
 #[repr(u8)]
 pub enum Alignment {
+    /// This is [`Alignment::Left`] for LTR text and [`Alignment::Right`] for RTL text.
     #[default]
     Start,
-    Middle,
+    /// This is [`Alignment::Right`] for LTR text and [`Alignment::Left`] for RTL text.
     End,
+    /// Align content to the left edge.
+    Left,
```
DJMcNab, line 50: I'd probably want to strongly encourage users to use `Alignment::Start` instead. They're more likely to get reasonable bidi handling if they do.

parley/src/layout/alignment.rs
```diff
@@ -89,12 +97,21 @@ pub(crate) fn align<B: Brush>(
...
+            // In RTL text, trailing whitespace is on the left. As we hang that whitespace, offset
+            // the line to the left.
+            line.metrics.offset -= line.metrics.trailing_whitespace;
+        }
     }
 }
 
 /// Removes previous justification applied to clusters.
 /// This is part of resetting state in preparation for re-linebreaking the same layout.
 pub(crate) fn unjustify<B: Brush>(layout: &mut LayoutData<B>) {
```
tomcur, line 111: I think `unjustify` is broken (and was broken prior to this, as it calculates `extra` / `free_space` differently). I intend to propose a simplification in a follow-up PR.

parley/src/layout/alignment.rs
```diff
@@ -32,37 +34,43 @@ pub(crate) fn align<B: Brush>(
...
-                    //   - Iterate backwards for odd bidi levels (which represent RTL runs)
+                // Iterate over text runs in the line and clusters in the text run
+                //   - Iterate forwards for even bidi levels (which represent LTR runs)
+                //   - Iterate backwards for odd bidi levels (which represent RTL runs)
+                let line_items: &mut dyn Iterator<Item = &LineItemData> = if is_rtl {
+                    &mut layout.line_items[line.item_range.clone()].iter().rev()
+                } else {
+                    &mut layout.line_items[line.item_range.clone()].iter()
+                };
+                for line_item in line_items.filter(|item| item.is_text_run()) {
```
wfdewith, line 73: I like this trick with a trait object, and it might be applicable in other places that deal with bidi as well. However, and this is probably _very_ premature optimization, but the Rust compiler doesn't optimize this very well (https://godbolt.org/z/sMqsTdfMs). Obviously, without benchmarks it is impossible to say what the actual impact is, but I thought I'd mention it at least.

tomcur: You are right about that, Rust calls through the vtable and unfortunately doesn't know to inline it with a branch. I think ideally this wouldn't be here. It may be possible with e.g. `skip`, but it's probably a good idea to first have tests with mixed LTR/RTL text working.

DJMcNab: I believe that you can use `Iterator::for_each`, which should optimise much more reasonably.

wfdewith: You could do the same for the inner loops over the clusters, because they are essentially duplicated at the moment.

nicoburns: I believe that (in theory) `Iterator::fold` should optimise even better, at the cost of not being able to `break` when `num_spaces` is reached (but I'm guessing that `break` isn't saving us much on typical text).

nicoburns: I've opened https://github.com/linebender/parley/issues/261 about creating benchmarks so that we can actually measure the impact of these kind of changes.

wfdewith:
Should we add some tests for bidi here as well? We don't test bidi anywhere yet, just LTR and this PR adds the first RTL tests.

tomcur:
> Should we add some tests for bidi here as well? We don't test bidi anywhere yet, just LTR and this PR adds the first RTL tests.

It's a good point, there are likely some issues with mixed-direction text (perhaps even caused by something in this PR), but my preference would be to land this PR to get fully LTR and fully RTL text aligning correctly, and then iterate on mixed-direction text.

parley/src/layout/alignment.rs
```diff
@@ -32,69 +37,80 @@ pub(crate) fn align<B: Brush>(
...
+            line.metrics.offset -= line.metrics.trailing_whitespace;
+        }
     }
 }
 
 /// Removes previous justification applied to clusters.
 /// This is part of resetting state in preparation for re-linebreaking the same layout.
 pub(crate) fn unjustify<B: Brush>(layout: &mut LayoutData<B>) {
+    // Whether the text base direction is right-to-left.
+    let is_rtl = layout.base_level & 1 == 1;
```
wfdewith, line 112: Idem `LayoutData::is_rtl`

parley/src/layout/alignment.rs
```diff
@@ -32,69 +37,80 @@ pub(crate) fn align<B: Brush>(
...
-                    // Iterate over clusters in the run
-                    //   - Iterate forwards for even bidi levels (which represent LTR runs)
-                    //   - Iterate backwards for odd bidi levels (which represent RTL runs)
-                    let clusters = &mut layout.clusters[line_item.cluster_range.clone()];
-                    let bidi_level_is_odd = line_item.bidi_level & 1 != 0;
-                    if bidi_level_is_odd {
-                        for cluster in clusters.iter_mut().rev() {
+                    .for_each(|line_item| {
+                        let clusters = &mut layout.clusters[line_item.cluster_range.clone()];
+                        let line_item_is_rtl = line_item.bidi_level & 1 != 0;
```
wfdewith, line 80: Should we add a `LineItemData::is_rtl` for this as well?

wfdewith (commented):
I have just two minor nits around `is_rtl`, but otherwise this seems fine to me. I'll leave the approving up to someone more experienced within Linebender.

parley/src/layout/line/greedy.rs
```diff
@@ -525,18 +525,33 @@ impl<'a, B: Brush> BreakLines<'a, B> {
...
+                    .filter(|item| item.is_text_run());
+                line.metrics.trailing_whitespace = last_run
+                    .and_then(|run| {
+                        self.layout.data.clusters[run.cluster_range.clone()]
+                            .last()
+                            .filter(|cluster| cluster.info.whitespace().is_space_or_nbsp())
+                            .map(|cluster| cluster.advance)
+                    })
+                    .unwrap_or(0.0);
+            }
```
nicoburns, line 554: Non-blocking, but it would be nice to deduplicate this code.

parley/src/layout/alignment.rs
```diff
@@ -104,10 +120,13 @@ pub(crate) fn unjustify<B: Brush>(layout: &mut LayoutData<B>) {
...
-                for line_run in layout.line_items[line.item_range.clone()]
-                    .iter()
-                    .filter(|item| item.is_text_run())
-                {
+
+                let line_items: &mut dyn Iterator<Item = &LineItemData> = if is_rtl {
+                    &mut layout.line_items[line.item_range.clone()].iter().rev()
+                } else {
+                    &mut layout.line_items[line.item_range.clone()].iter()
+                };
```
nicoburns, line 128: I suspect (with a full understanding of the invariants around `num_spaces`) that it's possible to refactor this to iterate in the same order regardless of bidi (probably with a fixup to trim trailing whitespace from the slice before starting? - but I'm speculating that trailing whitespace is why `num_spaces` may not equal the actual number of space char clusters), which would be a small win for both code runtime perf and binary size.

But this seems fine for now.

nicoburns (approved):
Approved. Some possible future improvements noted.

parley/src/layout/alignment.rs
tomcur: Yes, I think that's right. I also remember experimenting a bit, but it wasn't completely straightforward. Something to be revisited.

PR #252, opened 2025-01-24
fontique: Update to the new `objc2` CoreFoundation/Text bindings

waywardmonkeys:
The `objc2` family of crates now provides bindings for CoreFoundation and CoreText (and much more), so we can use these rather than the old `core-foundation` and `core-text` crates.

### DIFF
--- a/fontique/Cargo.toml
+++ b/fontique/Cargo.toml
@@ -25,7 +25,7 @@ unicode_script = ["dep:unicode-script"]
 system = [
 	"std",
 	"dep:windows", "dep:windows-core",
-	"dep:core-text", "dep:core-foundation", "dep:objc2-foundation",
+	"dep:objc2-core-foundation", "dep:objc2-core-text", "dep:objc2-foundation",
 	"dep:fontconfig-cache-parser", "dep:roxmltree",
 ]
 
@@ -46,9 +46,9 @@ windows = { version = "0.58.0", features = ["implement", "Win32_Graphics_DirectW
 windows-core = { version = "0.58", optional = true }
 
 [target.'cfg(target_vendor = "apple")'.dependencies]
-core-text = { version = "21.0.0", optional = true }
-core-foundation = { version = "0.10.0", optional = true }
-objc2-foundation = { version = "0.3.0", features = ["NSArray", "NSEnumerator", "NSPathUtilities", "NSString"], optional = true }
+objc2-foundation = { version = "0.3.0", optional = true, default-features = false, features = ["alloc", "NSArray", "NSEnumerator", "NSPathUtilities", "NSString"] }
+objc2-core-foundation = { version = "0.3.0", optional = true, default-features = false, features = ["CFBase"]  }
+objc2-core-text = { version = "0.3.0", optional = true, default-features = false, features = ["CTFont", "CTFontDescriptor"] }
 
 [target.'cfg(target_os = "linux")'.dependencies]
 fontconfig-cache-parser = { version = "0.2.0", optional = true }
--- a/fontique/src/backend/coretext.rs
+++ b/fontique/src/backend/coretext.rs
@@ -5,20 +5,16 @@ use super::{
     scan, FallbackKey, FamilyId, FamilyInfo, FamilyNameMap, GenericFamily, GenericFamilyMap,
 };
 use alloc::sync::Arc;
+use core::ptr::{null, null_mut};
 use hashbrown::HashMap;
+use objc2_core_foundation::{CFDictionaryCreate, CFRange, CFRetained, CFString, CFStringGetLength};
+use objc2_core_text::{
+    CTFont, CTFontCopyFamilyName, CTFontCreateForString, CTFontCreateForStringWithLanguage,
+    CTFontCreateUIFontForLanguage, CTFontCreateWithFontDescriptor,
+    CTFontDescriptorCreateWithAttributes, CTFontUIFontType,
+};
 use objc2_foundation::{
     NSSearchPathDirectory, NSSearchPathDomainMask, NSSearchPathForDirectoriesInDomains,
-};
-use {
-    core_foundation::{
-        base::{CFRange, TCFType},
-        dictionary::CFDictionary,
-        string::{CFString, CFStringRef},
-    },
-    core_text::{
-        font::{self, kCTFontSystemFontType, CTFont, CTFontRef, CTFontUIFontType},
-        font_descriptor,
-    },
 };
 
 const DEFAULT_GENERIC_FAMILIES: &[(GenericFamily, &[&str])] = &[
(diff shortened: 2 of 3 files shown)


waywardmonkeys:
@madsmtm This may be of interest to you as well.

fontique/src/backend/coretext.rs
```diff
@@ -77,97 +71,39 @@ impl SystemFonts {
...
-    let base_font = {
-        let desc_attrs = CFDictionary::from_CFType_pairs(&[]);
-        let mut desc = font_descriptor::new_from_attributes(&desc_attrs);
-        if prefer_ui {
-            if let Some(ui_desc) = ui_font_for_language(kCTFontSystemFontType, 0.0, None)
-                .map(|font| font.copy_descriptor())
-            {
-                desc = ui_desc;
-            }
-        }
```
waywardmonkeys, line 109: I didn't think we needed to get the UI front, then get the descriptor from that, then get a font from that descriptor. If we want a UI font, can't we just use the UI font that get from `CTFontCreateUIFontForLanguage`?

fontique/src/backend/coretext.rs
```diff
@@ -77,97 +71,39 @@ impl SystemFonts {
...
-            None
+    ) -> Option<CFRetained<CTFont>> {
+        let text = CFString::from_str(text);
+        let text_range = CFRange {
+            location: 0,
+            length: unsafe { CFStringGetLength(&text) },
+        };
+        let locale = locale.map(CFString::from_str);
+        let base_font = if prefer_ui {
+            unsafe { CTFontCreateUIFontForLanguage(CTFontUIFontType::System, 0.0, None).unwrap() }
```
waywardmonkeys, line 92: I assume this can't fail. I could be wrong?

fontique/src/backend/coretext.rs
```diff
@@ -77,97 +71,39 @@ impl SystemFonts {
...
         } else {
-            Some(CTFont::wrap_under_create_rule(font_ref))
-        }
+            unsafe {
+                let attrs = CFDictionaryCreate(None, null_mut(), null_mut(), 0, null(), null());
+                let desc = CTFontDescriptorCreateWithAttributes(&attrs.unwrap());
+                CTFontCreateWithFontDescriptor(&desc, 0.0, null())
```
waywardmonkeys, line 97: It is interesting that we don't pass any attributes here. What is the font that gets returned from this sort of thing?

fontique/Cargo.toml
```diff
@@ -46,10 +46,9 @@ windows = { version = "0.58.0", features = ["implement", "Win32_Graphics_DirectW
 windows-core = { version = "0.58", optional = true }
 
 [target.'cfg(target_vendor = "apple")'.dependencies]
-core-text = { version = "20.1.0", optional = true }
-core-foundation = { version = "0.9.4", optional = true }
-objc2 = { version = "0.5.2", optional = true }
-objc2-foundation = { version = "0.2.2", features = ["NSArray", "NSEnumerator", "NSPathUtilities", "NSString"], optional = true }
+objc2-foundation = { version = "0.3.0", optional = true, default-features = false, features = ["alloc", "NSArray", "NSEnumerator", "NSPathUtilities", "NSString"] }
+objc2-core-foundation = { version = "0.3.0", optional = true, default-features = false, features = ["CFBase"]  }
+objc2-core-text = { version = "0.3.0", optional = true }
```
waywardmonkeys, line 51: I still need to turn off `default-features` here and then update and re-format the above.

fontique/src/backend/coretext.rs
xorgy: I mean, since we are giving no `language` then presumably it is never going to fail in practice... though it is technically fallible.

fontique/src/backend/coretext.rs
xorgy: I believe in practice it gets you 12pt San Francisco (or Helvetica back-when). `CTFontCreateWithFontDescriptor` takes an unspecified descriptor as an opportunity to default on every field.

fontique/src/backend/coretext.rs
xorgy: That seems plausible, not sure why it's done that way right now.

dfrg: At some point, this was more involved and tried to select a font with specific properties thus a descriptor was required. The simplification looks good to me.

fontique/src/backend/coretext.rs
dfrg: I believe you always get the last resort font when matching fails but I try to avoid unwraps in cases where I can’t be 100% certain. My general philosophy is that libraries should never panic, particularly in cases like this where the user has no control over the behavior.

madsmtm (approved):
I cannot say anything about the specifics of the API (I kinda just do the bindings), but otherwise it looks good 👍.

Btw, I hope you didn't decide to stray away from using `CFDictionary` for the attributes just because [we don't provide a nice binding for creating those yet](https://github.com/madsmtm/objc2/issues/692)? If so, that's definitely unfortunate, then I'd rather that you wait with using `objc2-core-graphics` until we provide feature-parity.

waywardmonkeys:
> I cannot say anything about the specifics of the API (I kinda just do the bindings), but otherwise it looks good 👍.

I wanted your input on if I was using them correctly and maybe whether or not there is something that could be better (now or in the future).

> Btw, I hope you didn't decide to stray away from using `CFDictionary` for the attributes just because [we don't provide a nice binding for creating those yet](https://github.com/madsmtm/objc2/issues/692)? If so, that's definitely unfortunate, then I'd rather that you wait with using `objc2-core-graphics` until we provide feature-parity (I don't want people to use the crates when their usage ends up less flexible ;) ).

Nope! We weren't passing in any actual attributes in the old code either, so there's no functional change there really.

fontique/src/backend/coretext.rs
waywardmonkeys: I will adjust this to remove the `unwrap`.

DJMcNab:
I can't meaningfully review this code. I'm happy to formalise the approval from Mads, if needed.

waywardmonkeys:
Updated to remove the unwrap. Moved some stuff back to the top level to clean it up. The diff is terrible to read, sorry.

fontique/src/backend/coretext.rs
```diff
@@ -75,97 +71,45 @@ impl SystemFonts {
...
-impl SystemFonts {
-    fn fallback_for_text(
-        &mut self,
-        text: &str,
-        locale: Option<&str>,
-        prefer_ui: bool,
-    ) -> Option<FamilyId> {
-        let font = fallback_for_text(text, locale, prefer_ui)?;
-        self.name_map.get(&font.family_name()).map(|n| n.id())
+fn get_base_font(prefer_ui_font: bool) -> CFRetained<CTFont> {
```
DJMcNab, line 80: I don't massively like this `get_` prefix, but I don't think just `base_font` is great either. The methods we're calling all use 'Create'...

This isn't a blocking concern

DJMcNab (approved):
Largely formalising @madsmtm's previous approval

fontique/src/backend/coretext.rs
```diff
@@ -75,97 +71,45 @@ impl SystemFonts {
...
-        font::new_from_descriptor(&desc, 0.0)
+fn fallback_font_for_text(
+    text: &str,
+    locale: Option<&str>,
+    prefer_ui_font: bool,
+) -> Option<CFRetained<CTFont>> {
+    let text = CFString::from_str(text);
+    let text_range = CFRange {
+        location: 0,
+        length: unsafe { CFStringGetLength(&text) },
```
tomcur, line 103: Out of interest, is this the same as (before shadowing) `text.chars().count()`?

tomcur (approved):
I believe this is correct. I left one allocation nit as inline comment.

fontique/src/backend/coretext.rs
waywardmonkeys: I honesty don't know.

DJMcNab: I *think* it would be: `text.chars().map(char::len_utf16).sum()` ([docs](https://doc.rust-lang.org/stable/std/primitive.char.html#method.len_utf16))

fontique/src/backend/coretext.rs
waywardmonkeys: Renamed `get_base_font` and `fallback_font_for_text` to both have `create_` prefixes.

PR #253, opened 2025-01-24
Store snapshots using `git lfs`

tomcur:
This is based on [Vello's LFS setup](https://github.com/linebender/vello/blob/2f08fe770a6341857eca0754ec9efe2037325686/.github/workflows/ci.yml), including the mechanism that allows easily disabling the tests that rely on LFS, just in case bandwidth runs out.


tomcur:
Hmm... seems `git lfs` is having trouble actually pushing the files to GitHub.

wfdewith:
Interesting, GitHub doesn't show the images in the "Files changed" tab anymore, but if you click "View file", like [this](https://github.com/linebender/parley/blob/260456f51f169931670ab419cdc140e7dc398c39/parley/tests/snapshots/plain_multiline_text-0.png), it does show them.

tomcur:
Daniel and I think that's because the file moved from the repository to LFS, which GitHub's UI might not know how to deal with. In Vello, the LFS-stored files do have functioning diffs in GitHub's UI.

wfdewith:
Ah, yeah, it might not be able to show a diff between a non-LFS and LFS file, but that's only a problem for this specific PR.

.github/workflows/ci.yml
```diff
@@ -178,6 +215,11 @@ jobs:
 
       - name: cargo nextest
         run: cargo nextest run --workspace --locked --all-features --no-fail-fast
+        env:
+          # We are experimenting with git lfs, and we don't expect to run out of bandwidth.
```
DJMcNab, line 219: Maybe we should reword this slightly. Something like:
```suggestion
          # Although GitHub's LFS limits are very low, we don't expect to run out of bandwidth.
```

DJMcNab (approved):
This looks correct to me

tomcur:
CI is stuck again.

PR #254, opened 2025-01-24
Strip whitespace following a newline when whitespace collapsing is enabled

nicoburns:
This follows https://www.w3.org/TR/CSS2/text.html#white-space-model. Further work will be required to completely match that spec. In particular, whitespace preceding a newline also ought to be stripped but:

1. This is harder to implement in the current whitespace collapsing code
2. This is less consequential as trailing whitespace isn't usually visible.

This fixes a visible 1-space indent at the start of lines following a `<br />` tag in Blitz.


xorgy (approved):
This makes sense, as long as the documented purpose of `WhiteSpaceCollapse::Collapse` is to follow the CSS whitespace processing model.

tomcur (approved):
LGTM.

PR #256, opened 2025-01-26
Correctly calculate trailing whitespace for all lines

### LINKED ISSUE #251: Trailing whitespace is not marked correctly
In my attempts to implement `min_content_width` and `max_content_width`, I ran into another issue.

I've written the following test:
```rust
#[test]
fn trailing_whitespace() {
    let mut env = testenv!();

    let text = "AAA   BBB";
    layout.break_all_lines(Some(45.));
    layout.align(None, Alignment::Start, false);
    println!("width = {}", layout.width());
    println!("full_width = {}", layout.full_width());
    for (i, line) in layout.lines().enumerate() {
        println!("line {}, trailing_whitespace = {}", i, line.metrics().trailing_whitespace);
    }
}
```

This prints and renders the following:
```
width = 43.195313
full_width = 43.195313
line 0, trailing_whitespace = 0
line 1, trailing_whitespace = 0
```

Image

(I added the grey block to make the actual layout width and height visible with the patch below.)

```patch
diff --git a/parley/src/tests/utils/renderer.rs b/parley/src/tests/utils/renderer.rs
index 2169eb7..a5d6736 100644
--- a/parley/src/tests/utils/renderer.rs
+++ b/parley/src/tests/utils/renderer.rs
@@ -61,6 +61,15 @@ pub(crate) fn render_layout(
 
     let mut pen = TinySkiaPen::new(img.as_mut());
 
+    draw_rect(
+        &mut pen,
+        fpadding,
+        fpadding,
+        width as f32,
+        height as f32,
+        Color::from_rgba8(220, 220, 220, 255),
+    );
+
     for rect in selection_rects {
         draw_rect(
             &mut pen,
```

I would expect that the `full_width` of this layout includes the whitespace, while the `width` does not. Additionally, I would expect the first line to have `trailing_whitespace > 0` in its metrics.

wfdewith:
This should resolve the issue identified in #251. The current version of the code incorrectly checks the last of all line items, instead of the last line item of the current line.

I've also reduced the nested control flow blocks somewhat.

Fixes #251

### DIFF
--- a/parley/src/layout/line/greedy.rs
+++ b/parley/src/layout/line/greedy.rs
@@ -526,19 +526,17 @@ impl<'a, B: Brush> BreakLines<'a, B> {
             }
 
             // Compute size of line's trailing whitespace
-            line.metrics.trailing_whitespace = 0.0;
-            if !line.item_range.is_empty() {
-                // Note: there may not be a "last run" if there are no runs in the line
-                let last_item = &self.lines.line_items.last();
-                if let Some(last_item) = last_item {
-                    if last_item.is_text_run() && !last_item.cluster_range.is_empty() {
-                        let cluster = &self.layout.data.clusters[last_item.cluster_range.end - 1];
-                        if cluster.info.whitespace().is_space_or_nbsp() {
-                            line.metrics.trailing_whitespace = cluster.advance;
-                        }
-                    }
-                }
-            }
+            let last_run = &self.lines.line_items[line.item_range.clone()]
+                .last()
+                .filter(|item| item.is_text_run());
+            line.metrics.trailing_whitespace = last_run
+                .and_then(|run| {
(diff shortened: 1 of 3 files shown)


wfdewith:
I've updated the PR in line with @dfrg's definition of the current white space 'model' of Parley in [Zulip](https://xi.zulipchat.com/#narrow/channel/205635-text/topic/Parley.20white.20space.20model/near/496006047).

tomcur (approved):
Looks good, thanks! The bug was that the last line item in the layout was taken, instead of the last line item of this line?

wfdewith:
> Looks good, thanks! The bug was that the last line item in the layout was taken, instead of the last line item of this line?

Correct, I removed the part where I hang all trailing white space, it just hangs one cluster at most.

PR #265, opened 2025-02-02
Add the default for fontconfig cachedir

declantsien:
I have a fontconfig configuration file without `cachedir`.
Without this change `Collection.family_names()` is empty.
According to [FontConfig User Documation](https://fontconfig.pages.freedesktop.org/fontconfig/fontconfig-user.html), the default value for `cachedir` is `$XDG_CACHE_HOME/fontconfig`

~~FYI, This PR add an extra dep (xdg) for `fontique` on linux.~~

### DIFF
--- a/fontique/src/backend/fontconfig/config.rs
+++ b/fontique/src/backend/fontconfig/config.rs
@@ -30,6 +30,7 @@ pub fn parse_config(path: &Path, sink: &mut impl ParserSink) {
     if root.tag_name().name() != "fontconfig" {
         return;
     }
+    let mut has_cachedir = false;
     let mut prefer = vec![];
     'outer: for child in root.children() {
         match child.tag_name().name() {
@@ -65,6 +66,9 @@ pub fn parse_config(path: &Path, sink: &mut impl ParserSink) {
             "cachedir" => {
                 if let Some(path) = resolve_dir(child, path) {
                     sink.cache_path(&path);
+                    if !has_cachedir {
+                        has_cachedir = true;
+                    }
                 }
             }
             "include" => {
@@ -119,6 +123,16 @@ pub fn parse_config(path: &Path, sink: &mut impl ParserSink) {
                 }
             }
             _ => {}
+        }
+    }
+
+    if !has_cachedir {
+        if let Ok(home) = config_home() {
+            let default_cache_dir = std::env::var("XDG_CACHE_HOME")
+                .map(PathBuf::from)
+                .unwrap_or(Path::new(&home).join(".cache"))
+                .join("fontconfig");
+            sink.cache_path(&default_cache_dir);
         }
     }
 }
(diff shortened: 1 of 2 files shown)


fontique/src/backend/fontconfig/mod.rs
```diff
@@ -90,26 +90,13 @@ impl SystemFonts {
...
-            );
-        } else if let Ok(user_home) = std::env::var("HOME") {
-            config::parse_config(
-                PathBuf::from(user_home)
-                    .as_path()
-                    .join(".config/fontconfig/fonts.conf")
-                    .as_path(),
-                &mut config,
-            );
+        let xdg_dirs = xdg::BaseDirectories::with_prefix("fontconfig").unwrap();
```
DJMcNab, line 93: Is unwrapping the right behaviour here? In case we couldn't determine `$HOME` previously, it looks like we just didn't parse this file?

fontique/src/backend/fontconfig/mod.rs
```diff
@@ -314,6 +301,17 @@ struct Config {
     lang_maps: Vec<(String, StyleClass, String)>,
 }
 
+impl Config {
+    pub fn cache_dirs(&self) -> Vec<PathBuf> {
+        if self.cache_dirs.is_empty() {
+            let xdg_dirs = xdg::BaseDirectories::with_prefix("fontconfig").unwrap();
```
DJMcNab, line 307: This does an awful lot of allocation, given that we only use one item from it.

declantsien:
Not sure what's going on here https://github.com/linebender/parley/actions/runs/13114951204/job/36586895966?pr=265
@DJMcNab

fontique/src/backend/fontconfig/mod.rs
declantsien: Removed

fontique/src/backend/fontconfig/mod.rs
declantsien: Removed the xdg dep

fontique/src/backend/fontconfig/config.rs
```diff
@@ -30,6 +30,7 @@ pub fn parse_config(path: &Path, sink: &mut impl ParserSink) {
     if root.tag_name().name() != "fontconfig" {
         return;
     }
+    let mut cachedir_count = 0;
```
DJMcNab, line 33: I'd *probably* just use a boolean here, since the actual count doesn't matter.

tomcur:
It looks like the test failure is hitting against the TODO in `parley/src/layout/cluster.rs` line 477.

fontique/src/backend/fontconfig/config.rs
declantsien: Updated

tomcur:
I believe this PR isn't the cause of the test failure, but simply uncovered an issue where the test isn't explicitly using a known font. I'll likely have time to look at it, but perhaps not today.

declantsien:
Rebased on latest master branch

PR #267, opened 2025-02-06
Update dependencies

nicoburns:
This is just the result of running `cargo update`

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -20,15 +20,15 @@ checksum = "c71b1793ee61086797f5c80b6efa2b8ffa6d5dd703f118545808a7f2e27f7046"
 
 [[package]]
 name = "accesskit"
-version = "0.17.0"
+version = "0.17.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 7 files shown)


DJMcNab (commented):
If this PR is in service of a release, it also needs to update the Cargo.toml files.
We do the `cargo upgrade` and `cargo update` PRs as one PR.

nicoburns:
@DJMcNab Ok, this now upgrades everything except skrifa/read-fonts which I think ought to be handled separately because I'm not sure how that interacts with upgrading Swash's skrifa dep. @waywardmonkeys probably knows? Is `skrifa` a public dependency of Swash.

DJMcNab (approved):
Once CI is fixed, this seems reasonable.

I haven't carefully reviewed the FFI code. The macOS stuff is a trivially correct translation, I'm less sure about the windows change.

nicoburns:
I'm also not sure about the windows change. I've stripped it out for now so it can be reviewed separately.

waywardmonkeys:
I guess I'll have to rebase #252 now.

waywardmonkeys:
For future reference, since this is a library and not an application, you should get in the habit of doing `default-features = false` on objc2 crates as that changed in the update to 0.6 ... but I'll just get my #252 landed which already does that correctly.

PR #269, opened 2025-02-06
Use `Layout::width` instead of calculating for alignment

tomcur:
This does require line breaking to have happened, but that is already a requirement, as otherwise `Layout::lines` is empty.

This is a slight change in behavior, as trailing white space is no longer included in the alignment width.


nicoburns (approved):
Makes sense to use the already-computed value. But should we be using `Layout::width` here instead?

tomcur:
> But should we be using Layout::width here instead?

I wondered the same thing, but opened the PR keeping the same behavior. I changed it to `Layout::width` now – and fixed some more bugs that popped up with overflowing RTL-text.

PR #271, opened 2025-02-08
Fix and simplify `unjustify`, remove `LineData::alignment`

tomcur:
`unjustify` doesn't mirror `align`, so it performs the wrong calculations, and re-line-breaking or re-aligning currently lead to the wrong result after a justified alignment.

As the alignment function requires further iteration (e.g. for floated boxes), this patch removes the specialized `unjustify` implementation and instead uses the same code for aligning and unjustifying. This is a trade-off between efficiency and correctness. If this is merged in, we can always split the two again in the future once alignment has stabilized.

This stores the width the layout was aligned to, which is necessary for unjustifying.


nicoburns (approved):
This is a bit awkward, but it seems reasonable. If it fixes `unjustify` then I'm happy for it to be merged.

wfdewith (approved):
Since it fixes and simplifies things, it makes sense to merge this, but the whole idea of 'unjustifying' text feels somewhat wrong to me.

tomcur:
Agreed with both comments. Doing it a different way has some tradeoffs: I think it would either require more memory even if justified alignment is not used (e.g., adding another field like `ClusterData::adjusted_advance` that would be set by alignment and used for positioning, keeping `ClusterData::advance` as the original value), or it would require more operations for positioning (e.g., some place where we store cluster adjustments, which are added to the cluster advances as they are being iterated for drawing).

PR #275, opened 2025-02-16
Color padding in snapshot testing images

### LINKED ISSUE #274: Draw border to differentiate between layout and padding in snapshot tests
When I'm working on fixing bugs around trailing white space, I'd like it to be more clear whether Parley is rendering trailing white space or not. Similarly, it would be nice to immediately see anything that might be rendered out of bounds. To achieve this, I'd like to add some indication to differentiate padding from the actual layout in the snapshot test environment.

I currently add that indication manually, which I then remove for the PR, but it would be easier if it is the default.

Below are a couple options I experimented with.

Image

Image

Image
(Probably with a different color for the border.)

I personally prefer the second or third option, but I'm open to any ideas.

wfdewith:
As discussed in #274.

Fixes #274.


tomcur:
This is great. What is needed to move this out of draft?

tomcur (commented):
This is great. What is needed to move this out of draft?

wfdewith:
It requires some coordination around merging because everything that touches tests needs to be rebased, and this needs to be based on the most recent tip of main.

wfdewith:
It can be merged now, but be careful with merging other PRs after that.

DJMcNab (approved):

PR #276, opened 2025-02-16
Hang trailing whitespace preceding explicit newline

### LINKED ISSUE #263: Trailing whitespace with forced line breaks is not handled correctly
After #251, I noticed that trailing whitespace still shows up at the end of lines that end on a forced line break.

```rust
#[test]
fn trailing_whitespace() {
    let mut env = testenv!();

    let text = "AAA   \nBBB";
    layout.break_all_lines(None);
    layout.align(None, Alignment::Start, false);
}
```

With my patch that shows the padding, this renders the following image.

Image

<details><summary>Patch</summary>
<p>

```patch
diff --git a/parley/src/tests/utils/renderer.rs b/parley/src/tests/utils/renderer.rs
index 2169eb7..a5d6736 100644
--- a/parley/src/tests/utils/renderer.rs
+++ b/parley/src/tests/utils/renderer.rs
@@ -61,6 +61,15 @@ pub(crate) fn render_layout(
 
     let mut pen = TinySkiaPen::new(img.as_mut());
 
+    draw_rect(
+        &mut pen,
+        fpadding,
+        fpadding,
+        width as f32,
+        height as f32,
+        Color::from_rgba8(220, 220, 220, 255),
+    );
+
     for rect in selection_rects {
         draw_rect(
             &mut pen,
```
</p>
</details>

wfdewith:
Fixes #263


parley/src/layout/line/greedy.rs
```diff
@@ -535,14 +535,29 @@ impl<'a, B: Brush> BreakLines<'a, B> {
             line.metrics.trailing_whitespace = run
                 .filter(|item| item.is_text_run())
                 .and_then(|run| {
-                    let cluster = if self.layout.is_rtl() {
-                        self.layout.data.clusters[run.cluster_range.clone()].first()
+                    if self.layout.is_rtl() {
+                        // An RTL layout that has a line that ends with a space followed by a
+                        // forced line break will always have a separate run that looks like:
+                        // [" ", "\n"]. This means that the first cluster of a run with trailing
```
nicoburns, line 541: Could you explain why this is? Perhaps with an example? I believe you, but in terms of providing a meaningful review, I'm not currently fully following the logic here.

nicoburns:
We perhaps ought to land https://github.com/linebender/parley/pull/275 before this PR. As I think the tests for this PR would be much more useful with that change.

parley/src/layout/line/greedy.rs
wfdewith: So, I have to be honest here, and tell you that I don't _exactly_ know why RTL works like this.  I _assume_ that the space can be that the explicit newline forces mixed direction runs in the BiDi algorithm, but from my first skimming of https://www.unicode.org/reports/tr9/, line separators should be treated the same as any white space.

In any case, the current shaper gives you three runs like this: `[RTL("ااااا"), LTR(" \n"), RTL("بببب")]`, with the explicit newline. Without it you get just one run like this: `[RTL("بببب ااااا"]`.

I'll make this PR a draft until I understand it better and can explain it.

wfdewith:
@nicoburns I thought to revisit this PR now that ICU4X has been merged. The shaper now no longer mixes LTR and RTL runs in RTL text with hard line breaks, so there is no longer any difference between handling LTR and RTL text.

parley/src/layout/line_break.rs
```diff
@@ -576,16 +577,21 @@ impl<'a, B: Brush> BreakLines<'a, B> {
...
         line.metrics.trailing_whitespace = run
-            .filter(|item| item.is_text_run())
-            .and_then(|run| {
-                let cluster = if self.layout.is_rtl() {
-                    self.layout.data.clusters[run.cluster_range.clone()].first()
+            .filter(|item| item.is_text_run() && item.has_trailing_whitespace)
+            .map(|run| {
+                fn whitespace_advance<'c, I: Iterator<Item = &'c ClusterData>>(clusters: I) -> f32 {
+                    clusters
+                        .take_while(|cluster| cluster.info.whitespace() != Whitespace::None)
```
wfdewith, line 584: This could also be `cluster.info.whitespace() == Whitespace::Newline || cluster.info.whitespace().is_space_or_nbsp()`. I'm not sure how we should deal with trailing horizontal tabs and other whitespace.

nicoburns (approved):
Thanks for changing this to be a more intuitive implementation. I was not following the logic of the other implementation, but this one makes complete sense.

For others: the reason this is needed is because Parley appends newline clusters to the end of the previous line. So although they take up no horizontal space themselves, they were previously blocking us from detecting trailing whitespace.

parley/src/layout/line_break.rs
taj-p: I think this might need to be something like the below to prevent including newline characters counting towards trailing whitespace:

```rs
                        // Skip newlines (which have letter spacing but aren't hangable whitespace)
                        .skip_while(|cluster| cluster.info.whitespace() == Whitespace::Newline)
                        // Then count consecutive space/nbsp clusters
                        .take_while(|cluster| cluster.info.whitespace().is_space_or_nbsp())
```

The reason I picked up on this: This change seems to have caused a visual regression in letter spacing.

cc @nicoburns

wfdewith: Unfortunately, it's not that simple because of how the shaper shapes RTL text with explicit newlines. The original version of this PR correctly handled that, but was way too complex to understand. The simplification I made seemed to still be correct, but apparently it is not.

I'm inclined to say that this might actually be a shaper bug, and that your suggestion _should_ work, and if it doesn't, the shaper produces incorrect runs.

Do you have a test case for this regression?

wfdewith: @taj-p  Actually, now that I have some time to take a closer look, I think your solution is correct.

When I was initially working on this PR, Swash was still used to shape text. Swash produced RTL and LTR runs when an explicit linebreak was added in an RTL-only script, like `[RTL("ااااا"), LTR(" \n"), RTL("بببب")]` (see https://github.com/linebender/parley/pull/276#discussion_r1958805160). This behavior is what I meant with 'shaper bug' in my previous comment. Swash also reversed the cluster order within an RTL run, so the cluster at index 0 in the run would be the _left-most_ cluster rendered for that run.

Since then, shaping has been migrated to HarfRust, which works differently. It generates RTL runs for explicit linebreaks within RTL-only script, like `[RTL("ااااا"), RTL(" \n"), RTL("بببب")]`. Furthermore, it also doesn't reverse the clusters within a run, so the cluster at index 0 in a run is the _right-most_ cluster for that run.

That means that the last cluster in the cluster range is always going to be the end of the line, regardless of direction, so we no longer need to check for RTL within a run. The solution below should be correct.
```rust
line.metrics.trailing_whitespace = run
    .filter(|item| item.is_text_run() && item.has_trailing_whitespace)
    .map(|run| {
        self.layout.data.clusters[run.cluster_range.clone()]
            .iter()
            .rev()
            .skip_while(|cluster| cluster.info.whitespace() == Whitespace::Newline)
            .take_while(|cluster| cluster.info.whitespace().is_space_or_nbsp())
            .map(|cluster| cluster.advance)
            .sum()
    })
    .unwrap_or(0.0);

```

PR #278, opened 2025-02-17
Replace boolean parameter to `Layout::align` with `AlignmentOptions` struct

### LINKED ISSUE #247: Replace boolean option in `Layout::align` with `AlignmentOptions` Struct
See https://github.com/linebender/parley/pull/241#discussion_r1924001182

nicoburns:
- Fixes #247
- Builds on top of https://github.com/linebender/parley/pull/271 as otherwise these would conflict

A small but nice improvement to the public API.

### DIFF
--- a/parley/src/layout/alignment.rs
+++ b/parley/src/layout/alignment.rs
@@ -7,6 +7,27 @@ use super::{
 };
 use crate::style::Brush;
 
+/// Additional options to fine tune alignment
+#[derive(Debug, Clone)]
+pub struct AlignmentOptions {
+    /// If set to `true`, "end" and "center" alignment will apply even if the line contents are
+    /// wider than the alignment width. If it is set to `false`, all overflowing lines will be
+    /// [`Alignment::Start`] aligned.
+    pub align_when_overflowing: bool,
+}
+
+#[expect(
+    clippy::derivable_impls,
+    reason = "Make default values explicit rather than relying on the implicit default value of bool"
+)]
+impl Default for AlignmentOptions {
+    fn default() -> Self {
+        Self {
+            align_when_overflowing: false,
+        }
+    }
+}
+
 /// Align the layout.
 ///
 /// If [`Alignment::Justified`] is requested, clusters' [`ClusterData::advance`] will be adjusted.
@@ -15,12 +36,12 @@ pub(crate) fn align<B: Brush>(
     layout: &mut LayoutData<B>,
     alignment_width: Option<f32>,
     alignment: Alignment,
-    align_when_overflowing: bool,
+    options: AlignmentOptions,
 ) {
     layout.alignment_width = alignment_width.unwrap_or(layout.width);
     layout.is_aligned_justified = alignment == Alignment::Justified;
 
-    align_impl::<_, false>(layout, alignment, align_when_overflowing);
+    align_impl::<_, false>(layout, alignment, options);
 }
 
 /// Removes previous justification applied to clusters.
@@ -29,7 +50,7 @@ pub(crate) fn align<B: Brush>(
 /// layout.
 pub(crate) fn unjustify<B: Brush>(layout: &mut LayoutData<B>) {
     if layout.is_aligned_justified {
-        align_impl::<_, true>(layout, Alignment::Justified, false);
+        align_impl::<_, true>(layout, Alignment::Justified, Default::default());
         layout.is_aligned_justified = false;
     }
 }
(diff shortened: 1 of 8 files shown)


parley/src/layout/alignment.rs
```diff
@@ -7,31 +7,84 @@ use super::{
 };
 use crate::style::Brush;
 
+/// Additional options to fine tune alignment
+#[derive(Debug, Clone)]
+pub struct AlignmentOptions {
+    /// If set to `true`, "end" and "center" alignment will apply even if the line contents are
+    /// wider than the alignment width. If it is set to `false`, all overflowing lines will be
+    /// [`Alignment::Start`] aligned.
+    align_when_overflowing: bool,
```
wfdewith, line 16: This field should be `pub`.

nicoburns: Good point!

parley/src/layout/alignment.rs
```diff
@@ -7,31 +7,84 @@ use super::{
 };
 use crate::style::Brush;
 
+/// Additional options to fine tune alignment
+#[derive(Debug, Clone)]
+pub struct AlignmentOptions {
+    /// If set to `true`, "end" and "center" alignment will apply even if the line contents are
+    /// wider than the alignment width. If it is set to `false`, all overflowing lines will be
+    /// [`Alignment::Start`] aligned.
+    pub align_when_overflowing: bool,
```
tomcur, line 16: A nit: it may read more naturally by flipping the meaning of the bool and calling the field `start_alignment_on_overflow`.

tomcur (approved):
Nice!

parley/src/layout/mod.rs
```diff
@@ -181,23 +182,14 @@ impl<B: Brush> Layout<B> {
...
-    ///
-    /// If `align_when_overflowing` is set to `true`, alignment will apply even if the line
-    /// contents are wider than the alignment width. If it is set to `false`, all overflowing lines
-    /// will be [`Alignment::Start`] aligned.
     pub fn align(
         &mut self,
         container_width: Option<f32>,
         alignment: Alignment,
-        align_when_overflowing: bool,
+        options: AlignmentOptions,
```
DJMcNab, line 189: I'm not convinced by this signature. Having a pile of arguments, and then an options struct doesn't quite seem right to me:
```rust
fn align(&mut self, alignment: Alignment, options: AlignmentOptions)
```
seems more natural, given that `container_width` is already optional.

Or even:
```rust
fn align(&mut self, options: AlignmentOptions)
```

DJMcNab (commented):
Want to get discussion on the signature of `align`.

parley/src/layout/editor.rs
```diff
@@ -25,6 +25,8 @@ use crate::layout::LayoutAccessibility;
 #[cfg(feature = "accesskit")]
 use accesskit::{Node, NodeId, TreeUpdate};
 
+use super::AlignmentOptions;
```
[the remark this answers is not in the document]
nicoburns: Why?

parley/src/layout/alignment.rs
```diff
@@ -7,6 +7,27 @@ use super::{
...
+}
+
+#[expect(
+    clippy::derivable_impls,
+    reason = "Make default values explicit rather than relying on the implicit default value of bool"
+)]
+impl Default for AlignmentOptions {
+    fn default() -> Self {
+        Self {
+            align_when_overflowing: false,
```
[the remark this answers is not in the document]
nicoburns: I don't have particularly strong feelings about the default.

parley/src/layout/mod.rs
tomcur: One issue with moving `container_width` into the options is that it cannot account for floated boxes yet, and I don't think we know what that's going to look like yet.

I certainly see the logic of moving `alignment` into `AlignmentOptions`.

parley/src/layout/alignment.rs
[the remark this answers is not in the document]
tomcur: I believe the current default follows what the web does (i.e., when a line overflows, you always see the start of it, in both LTR and RTL text).

parley/src/layout/editor.rs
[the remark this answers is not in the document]
DJMcNab: For consistency with the rest of the project (and imports within this file).

It's also slightly nicer for refactoring to not use it, as moving this file means the imports wouldn't need to change (in theory). That's pretty minor, in the grand scheme of things though

parley/src/layout/mod.rs
nicoburns: > Just another consideration: perhaps container_width shouldn't be optional. Aligning to the layout's calculated width is not necessarily a sensible default, as that width changes depending on how exactly lines were broken given the container width (max_advance) used for line breaking.

This could also make sense.

nicoburns:
I think I'm going to merge this as-is, as there seems to be agreement that this is an improvement over what we have. Further API changes and/or default changes can happen in followups.

waywardmonkeys:
I don't know why there was a rush to land this rather than just get it right.

I won't include this in my update to masonry since apparently it has to change again (although now that this landed, who will drive actually fixing it?).

PR #280, opened 2025-02-17
Account for inline boxes when collapsing after newlines

nicoburns:
Followup (small bug fix) to https://github.com/linebender/parley/pull/254 which didn't take inline boxes into account. That PR was checking whether "the last character is whitespace" without checking whether the last thing pushed into the layout was text.

So if the sequence of items pushed to a layout was something like:

- inline box
- whitespace
- inline box
- whitespace
- inline box
- whitespace
- inline box

(an example which would be "status badges" in Github readme's which are typically images separated by whitespace characters)

Then all but the first "whitespace" would get collapsed as it would be incorrectly detected as directly adjacent to the preceding whitespace (even though it shouldn't because there is an inline box in between).

This PR fixes this issue by tracking the kind of the last item pushed.


nicoburns:
Hmm.. the windows "restore lfs cache" task is generating a different cache key to linux/mac and is then not finding the cached LFS files.

parley/src/tests/test_basic.rs
```diff
@@ -112,6 +112,46 @@ fn full_width_inbox() {
     }
 }
 
+#[test]
+fn inboxes_separated_by_whitespace() {
+    let mut env = testenv!();
+
+    let text = "   ";
+    let mut builder = env.ranged_builder(text);
```
wfdewith, line 120: Shouldn't this test use `TreeBuilder`? It currently has the exact same behavior on main, so it doesn't really demonstrate the bug this PR is fixing.

nicoburns: Yes, yes it should. The new snapshot has some grey rectangles around the inline boxes. What do they represent?

DJMcNab:
(For reference, the thread is https://xi.zulipchat.com/#narrow/channel/419691-linebender/topic/Parley.20CI.20failing.20.28due.20to.20LFS.20failure.3F.29)

wfdewith (approved):
It makes sense to me, and is similar in structure to the line breaker. Should we derive `Debug` on those internal structs?

PR #282, opened 2025-02-18
Fix changelog link

tomcur:


wfdewith (approved):
You'd start thinking we should automate these things.

tomcur:
Yes... markdown linting could be done in CI, but requires actually taking time to set it up.

PR #293, opened 2025-03-07
[WIP] Vertical alignment

valadaptive:
This takes one vertical position as the origin, and aligns everything to that, e.g. if you select text-top, the glyph's ascender will be at the origin. This works well if every glyph has the same vertical alignment, but if not, it falls apart (e.g. if one glyph is aligned to text-top, and the next is aligned to text-bottom, they'll be extremely vertically separated). What CSS does is align everything using the metrics of the "parent style".

I didn't get as far as implementing the alignment for actual glyph positioning, only line height calculation.


nicoburns:
Superceded by https://github.com/linebender/parley/pull/766

PR #294, opened 2025-03-08
Update to `accesskit` 0.18

waywardmonkeys:


waywardmonkeys:
I guess this will need a breaking revision published.

nicoburns:
> I guess this will need a breaking revision published.

I really think we should be more liberal with this. If we published a 0.4 version of Parley today with only this version, would anyone be inconvenienced by that? My take is that no, they would not.

Having said that, we could also consider splitting out Parley's accesskit support to live in separate crate, which would alleviate that issue entirely.

waywardmonkeys:
I'm not against a 0.4,  but I'll also be suggesting some other things that would be part of a 0.4. I think we'll want to do that 0.4 release sooner rather than later.

nicoburns (approved):

tomcur (approved):

PR #299, opened 2025-03-15
Fix text editing for layouts which contain inline boxes

valadaptive:
This PR includes https://github.com/linebender/parley/pull/296, since they both touch the selection-drawing code.

There were a couple issues preventing text cursor positioning and selection from working on lines with inline boxes:
- `Cluster::from_point` and `Cluster::from_byte_index` were setting the `run_index` in their cluster paths to the index of the nth line item *that is a run*. It should be set to the nth line item, regardless of whether the previous items are inline boxes.
- `Cluster::from_point` and `Selection::geometry_with` were ignoring inline boxes when accumulating the lines' advance widths.

It's not perfect yet. Cursor affinity should take inline boxes into account, but since we could have multiple inline boxes in a row, or inline boxes next to soft line breaks, the cursor affinity will need to be an index and not just an upstream/downstream state (how do other APIs handle this?) That's a much bigger undertaking for a later time.

There was also a discrepancy between the documentation for inline boxes and the implementation--`InlineBox::index` is documented as the byte index to insert the box at, whereas the shaping code treated it as a character index. I've changed the shaping code to treat it as a byte index, but I could change the documentation instead.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -37,6 +37,12 @@ This release has an [MSRV] of 1.82.
 - Display selected newlines as whitespace in the selection highlight. ([#296][] by [@valadaptive][])
 - Make `BreakReason` public. ([#300][] by [@valadaptive][])
 
+### Fixed
+
+#### Parley
+
+- Fix text editing for layouts which contain inline boxes ([#299][] by [@valadaptive][])
+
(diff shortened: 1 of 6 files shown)


xorgy:
#296 is merged so this can be rebased.

xorgy (approved):

PR #300, opened 2025-03-15
Make BreakReason public

valadaptive:
The `break_reason` accessor was public, but `BreakReason` was not, so nothing could be done with the accessor.


xorgy (approved):
I don't see any issue with this, since it is a plain copy enum.

Issue #298, opened 2025-03-15
Backwards cursor navigation gets stuck in a loop with RTL text + newline


valadaptive:
If there's a newline directly after an RTL segment, trying to navigate backwards/leftwards using the left arrow key (corresponding to `Selection::previous_visual`) will loop the cursor back to the beginning of the RTL segment instead of breaking out of it:

https://github.com/user-attachments/assets/efa1cde9-4dc1-497d-a356-3cbbe1911383

You can reproduce this with the following text in the `vello_editor` example:
```rs
pub const LOREM: &str = r" Lorem ipsum dolor sit amet, consectetur adipiscing elit. Morbi cursus mi sed euismod euismod. Orci varius natoque penatibus et magnis dis parturient montes, nascetur ridiculus mus. Nullam placerat efficitur tellus at semper. Morbi ac risus magna. Donec ut cursus ex. Etiam quis posuere tellus. Mauris posuere dui et turpis mollis, vitae luctus tellus consectetur. Lorem ipsum dolor sit amet, consectetur adipiscing elit. Curabitur eu facilisis nisl.

Phasellus in viverra dolor, vitae facilisis est. Maecenas malesuada massa vel ultricies feugiat. Vivamus venenatis et gהתעשייה בנושא האינטרנטa nibh nec pharetra. Phasellus vestibulum elit enim, nec scelerisque orci faucibus id. Vivamus consequat purus sit amet orci egestas, non iaculis massa porttitor. Vestibulum ut eros leo. In fermentum convallis magna in finibus. Donec justo leo, maximus ac laoreet id, volutpat ut elit. Mauris sed leo non neque laoreet faucibus. Aliquam orci arcu, faucibus in molestie eget, ornare non dui. Donec volutpat nulla in fringilla elementum. Aliquam vitae ante egestas ligula tempus vestibulum sit amet sed ante.
aהתעשייה בנושא האינטרנט
";
```

xorgy:
I've run into this but was too lazy to report it properly. Thanks! 
Let's start a branch/draft pr with a test for this, I can do it later if you don't want to. :+ )

valadaptive:
The issue goes away if I comment out [these lines](https://github.com/linebender/parley/blob/4bb6859668815e717491ce55e6fd9390e70b0e9a/parley/src/layout/cursor.rs#L29-L33):

```rust
let affinity = if cluster.is_line_break() == Some(BreakReason::Explicit) {
    Affinity::Downstream
} else {
    affinity
[...]

PR #305, opened 2025-03-19
Add cursor tests for RTL text followed by newline

valadaptive:
As requested in https://github.com/linebender/parley/issues/298.

It turns out to be a bit worse than I originally thought--when I used ctrl+left arrow in the vello_editor example, it just put me back at the start of the word. But when I call `previous_visual_word` here, it actually hangs and loops infinitely!

I'm not confident that the assertions for the expected behavior are correct--the assertion helpers seem to be a bit weird. As such, anyone fixing this should not necessarily expect the tests to pass once they've done so.


xorgy:
Thank you!

Issue #101, opened 2024-08-09
fontique: Loading fonts from fs


waywardmonkeys:
We should be able to load a font from the file system or scan a directory for fonts, rather than only having FS access via `system`. (We should be able to let the text stack memory map arbitrary font files without having to load them into a blob.)

In `fontdb`, this is easy to do by using `Database::load_font_file` or `Database::load_font_dir`.

In `font-kit`, this can be done with the `sources::fs::FsSource`.

waywardmonkeys:
I was wrong about part of the above. `Blob` works with `memmap2` so that you aren't loading the whole thing. But the other stuff is still true, I think.

dfrg:
Having `register_fonts` take a `Blob` instead of `Vec<u8>` would be one step in this direction. We could add `register_fonts_from_path` to solve the rest, but we’d want to consider whether these should be fully loaded or just parsed and added to a source path map to be lazily loaded later.

Also, since memory mapping is considered generally unsafe in Rust, we should offer the option to disable it, even for system collections.

PR #306, opened 2025-03-19
Change `register_fonts` to take a `Blob`

valadaptive:
As mentioned in https://github.com/linebender/parley/issues/101. This is nice because it lets us pass in static (or potentially static) font data that would otherwise take up twice the memory (one embedded in the executable, and one cloned from it).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,7 +19,7 @@ This release has an [MSRV] of 1.85.
 
 #### Parley
 
-- Breaking change: `Collection::register_fonts` now takes a `Blob<u8>` instead of a `Vec<u8>` ([#306][] by [@valadaptive])
+- Breaking change: `Collection::register_fonts` now takes a `Blob<u8>` instead of a `Vec<u8>` ([#306][] by [@valadaptive][])
 
 ### Internals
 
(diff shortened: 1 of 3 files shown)


DJMcNab (approved):
LGTM. Thanks!

valadaptive:
Added a changelog entry.

CHANGELOG.md
```diff
@@ -15,6 +15,12 @@ You can find its changes [documented below](#030---2025-02-27).
 
 This release has an [MSRV] of 1.85.
 
+### Changed
+
+#### Parley
+
+- Breaking change: `Collection::register_fonts` now takes a `Blob<u8>` instead of a `Vec<u8>` ([#306][] by [@valadaptive])
```
waywardmonkeys, line 22: You'll need to add the link stuff below for both `[#306]` and `[@valadaptive]`. You can use the `[]` after your name as well, as you did for the PR link.

Just keep the names below alphabetically sorted. :)

waywardmonkeys (approved):
Nit with the changelog, but other than that, LGTM.

CHANGELOG.md
valadaptive: Ah, sneaky! I've added those now as well.

xorgy (approved):
Thank you!

waywardmonkeys:
@valadaptive Since you don't (yet) have commit access here, I'll merge this one for now.

PR #307, opened 2025-03-19
Switch back to MSRV 1.82 & 2021 edition.

waywardmonkeys:
We keep the Rust Style Edition 2024, so the formatting changes previously done remain.

This reverts parts of commit 7e4aa0bab95369a62f0e1001b5f0fbc907f5599e.

This is in response to some feedback from potential community users who were concerned about an aggressive MSRV.

### DIFF
--- a/parley/src/layout/run.rs
+++ b/parley/src/layout/run.rs
@@ -112,7 +112,7 @@ impl<'a, B: Brush> Run<'a, B> {
     }
 
     /// Returns an iterator over the clusters in logical order.
-    pub fn clusters(&'a self) -> impl Iterator<Item = Cluster<'a, B>> + Clone {
+    pub fn clusters(&'a self) -> impl Iterator<Item = Cluster<'a, B>> + 'a + Clone {
         let range = self.cluster_range();
         Clusters {
             run: self,
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -8,7 +8,7 @@ env:
   # If the compilation fails, then the version specified here needs to be bumped up to reality.
   # Be sure to also update the rust-version property in the workspace Cargo.toml file,
   # plus all the README.md files of the affected packages.
-  RUST_MIN_VER: "1.85"
+  RUST_MIN_VER: "1.82"
   # List of packages that will be checked with the minimum supported Rust version.
   # This should be limited to packages that are intended for publishing.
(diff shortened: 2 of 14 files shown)


fontique/src/generic.rs
```diff
@@ -214,7 +214,7 @@ mod tests {
         let mut value = 0;
         while value <= GenericFamily::MAX_VALUE {
             // Safety: In a const context, therefore if this makes an invalid GenericFamily, that will be detected.
-            // TODO: use `&raw const value` instead of the addr_of!
+            // When updating the MSRV to 1.82 or later, this can use `&raw const value` instead of the addr_of!
```
waywardmonkeys, line 217: Once this lands, I'll submit another PR that just fixes this and removes the comment.

fontique/src/generic.rs
```diff
@@ -239,7 +239,7 @@ mod doctests {
     ///     let value = GenericFamily::MAX_VALUE + 1;
     ///     // Safety: In a const context, therefore if this makes an invalid GenericFamily, that will be detected.
     ///     // (Indeed, we rely upon that)
-    ///     // TODO: use `&raw const value` instead of the addr_of!
+    ///     // When updating the MSRV to 1.82 or later, this can use `&raw const value` instead of the addr_of!
```
waywardmonkeys, line 242: As above.

nicoburns (approved):

xorgy (approved):

fontique/src/generic.rs
waywardmonkeys: This was fixed in PR #308 separately.

fontique/src/generic.rs
waywardmonkeys: This was fixed in PR #308 separately.

parley/src/layout/run.rs
```diff
@@ -112,7 +112,7 @@ impl<'a, B: Brush> Run<'a, B> {
     }
 
     /// Returns an iterator over the clusters in logical order.
-    pub fn clusters(&'a self) -> impl Iterator<Item = Cluster<'a, B>> + Clone {
+    pub fn clusters(&'a self) -> impl Iterator<Item = Cluster<'a, B>> + 'a + Clone {
```
DJMcNab, line 115: Something something this should probably use the captures trick, but it's also underdocumented and everyone gets this wrong, and we'll be moving to 1.85 in the reasonably near future anyway.

PR #311, opened 2025-03-20
Handle the rename of `#text` to `#parley`

DJMcNab:


DJMcNab:
It is as far as I know! I did a search for `205635` and for `#text`

Issue #117, opened 2024-09-17
Fontique: Support registering a font with custom family name and `FontInfo`


nicoburns:
The use case for this is supporting CSS web fonts (`@font-face`)

Requirements:
- Register a font, overriding the family name
- It should be possible to register multiple fonts with the same family name
- It should be possible to specify custom [FontInfo](https://github.com/linebender/parley/blob/main/fontique/src/font.rs#L17) for each font (overriding the data from the font file).

Non-requirements:
- Supporting font collections (.ttc files). Web fonts do not support font collections, so for the purposes of this API we can assume that 

---

Ideally it should also be possible to register custom fonts in a separate "layer" to the system fonts such that:

- A different set of custom fonts can be used for (e.g. each page/tab), while still sharing system fonts (not entirely duplicating the database)
- It is possible to easily "roll back"/"unregister" the custom fonts (e.g. when navigating between pages / documents)

(but perhaps this is a separate feature)

PR #312, opened 2025-03-20
Add APIs for overriding fonts' metadata at registration time and unregistering fonts

valadaptive:
Resolves https://github.com/linebender/parley/issues/117, implementing the APIs requested there.

The first API added is `Collection::clear`, which resets the list of registered fonts and associated data (custom generic families and fallbacks, etc).

The second API addition is a helper struct which can be passed as a parameter to `Collection::register_fonts`, which allows the caller to override the font's family name and other metadata. I originally thought about passing a callback which could be called once per font in the provided data, but I think that's overkill and makes the most common use case a lot more cumbersome.

This is necessary for the egui work, since egui lets users specify the names of registered fonts.

### DIFF
--- a/fontique/src/collection/mod.rs
+++ b/fontique/src/collection/mod.rs
@@ -7,6 +7,8 @@ mod query;
 
 pub use query::{Query, QueryFamily, QueryFont, QueryStatus};
 
+use crate::font::FontInfoOverride;
+
 use super::SourceCache;
 
 use super::{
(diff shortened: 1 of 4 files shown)


waywardmonkeys:
@nicoburns Please make sure this meets your needs as well.

nicoburns (approved):
I'll need a bit more than this:
- Ranges for things like font-weight
- The ability to override the unicode range that the font is matched for

See: https://developer.mozilla.org/en-US/docs/Web/CSS/@font-face
But this seems like an excellent start that probably solves my 80% use case, and could be built upon.

valadaptive:
Can the ranges be done as part of the font axis stuff? Not sure how fontique resolves those in the backend.

I agree on punting on Unicode ranges for now.

waywardmonkeys:
This should have had a changelog entry before merging.

nicoburns:
> This should have had a changelog entry before merging.

https://github.com/linebender/parley/pull/313

PR #315, opened 2025-03-25
Implement word-break and overflow-wrap style properties

valadaptive:
Requires upstream support in Swash (https://github.com/dfrg/swash/pull/89).

This implements the CSS `word-break` and `overflow-wrap` style properties (including the latter's influence on the layout's min-content size).

This replaces the "emergency break" logic with something that should be simpler.

Different browsers seem to disagree on exactly which ranges of text the properties apply to. The behavior implemented here should be reasonable.

In the table below, red is the default wrapping settings, blue is `overflow-wrap: anywhere`, and green is `word-break: break-all`:

| Gecko | Blink | WebKit | Parley |
| - | - | - | - |
image | image | image | image
image | image | image | image
image | image | image | image
image| image | image | image
image | image | image | image
image | image | image | image \*
image | image | image | word_break_break_all_second_half-0

\* *This doesn't match browsers, but the behavior is officially unspecified per https://github.com/w3c/csswg-drafts/issues/3897*

### DIFF
--- a/parley/src/context.rs
+++ b/parley/src/context.rs
@@ -120,7 +120,7 @@ impl<B: Brush> LayoutContext<B> {
                 }
                 style_idx += 1;
             }
-            a.set_break_strength(word_break.as_swash());
+            a.set_break_strength(word_break);
 
             let Some((properties, boundary)) = a.next() else {
                 break;
--- a/parley/src/context.rs
+++ b/parley/src/context.rs
@@ -107,9 +107,7 @@ impl<B: Brush> LayoutContext<B> {
 
         let mut char_indices = text.char_indices();
         loop {
-            let (Some((properties, boundary)), Some((char_idx, _))) =
-                (a.next(), char_indices.next())
-            else {
+            let Some((char_idx, _)) = char_indices.next() else {
                 break;
             };
 
@@ -122,8 +120,12 @@ impl<B: Brush> LayoutContext<B> {
                 }
                 style_idx += 1;
             }
-            // Set the word break strength for the *next* character, which seems to be what Chrome does.
             a.set_break_strength(word_break.as_swash());
+
+            let Some((properties, boundary)) = a.next() else {
+                break;
+            };
+
             self.info.push((CharInfo::new(properties, boundary), 0));
         }
         if a.needs_bidi_resolution() {
--- a/parley/src/context.rs
+++ b/parley/src/context.rs
@@ -67,7 +67,6 @@ impl<B: Brush> LayoutContext<B> {
         scale: f32,
     ) -> RangedBuilder<'a, B> {
         self.begin();
-        self.analyze_text(text);
         self.ranged_style_builder.begin(text.len());
 
         fcx.source_cache.prune(128, false);
@@ -102,8 +101,30 @@ impl<B: Brush> LayoutContext<B> {
     pub(crate) fn analyze_text(&mut self, text: &str) {
         let text = if text.is_empty() { " " } else { text };
         let mut a = swash::text::analyze(text.chars());
-        for x in a.by_ref() {
-            self.info.push((CharInfo::new(x.0, x.1), 0));
+
+        let mut word_break = Default::default();
+        let mut style_idx = 0;
+
+        let mut char_indices = text.char_indices();
+        loop {
+            let (Some((properties, boundary)), Some((char_idx, _))) =
+                (a.next(), char_indices.next())
+            else {
+                break;
+            };
+
+            // Find the style for this character. If the text is empty, we may not have any styles. Otherwise,
+            // self.styles should span the entire range of the text.
+            while let Some(style) = self.styles.get(style_idx) {
+                if style.range.end > char_idx {
+                    word_break = style.style.word_break;
+                    break;
+                }
+                style_idx += 1;
+            }
+            // Set the word break strength for the *next* character, which seems to be what Chrome does.
+            a.set_break_strength(word_break.as_swash());
+            self.info.push((CharInfo::new(properties, boundary), 0));
         }
         if a.needs_bidi_resolution() {
             self.bidi.resolve(
--- a/parley/src/layout/line/greedy.rs
+++ b/parley/src/layout/line/greedy.rs
@@ -10,6 +10,7 @@ use swash::text::cluster::Whitespace;
 #[allow(unused_imports)]
 use core_maths::CoreFloat;
 
+use crate::OverflowWrap;
 use crate::layout::{
     Boundary, BreakReason, Layout, LayoutData, LayoutItem, LayoutItemKind, LineData, LineItemData,
     LineMetrics, Run,
@@ -63,6 +64,7 @@ struct BreakerState {
 
     line: LineState,
     prev_boundary: Option<PrevBoundaryState>,
+    emergency_boundary: Option<PrevBoundaryState>,
 }
 
 impl BreakerState {
@@ -88,6 +90,17 @@ impl BreakerState {
     /// the line breaking opportunity at this point.
     fn mark_line_break_opportunity(&mut self) {
         self.prev_boundary = Some(PrevBoundaryState {
+            item_idx: self.item_idx,
+            run_idx: self.run_idx,
+            cluster_idx: self.cluster_idx,
+            state: self.line.clone(),
+        });
+    }
+
+    /// Store the current iteration state so that we can revert to it if we later want to take
+    /// an *emergency* line breaking opportunity at this point.
+    fn mark_emergency_break_opportunity(&mut self) {
+        self.emergency_boundary = Some(PrevBoundaryState {
             item_idx: self.item_idx,
             run_idx: self.run_idx,
             cluster_idx: self.cluster_idx,
@@ -128,6 +141,7 @@ impl<'a, B: Brush> BreakLines<'a, B> {
         self.state.lines = self.lines.lines.len();
         self.state.line.x = 0.;
         self.state.prev_boundary = None; // Added by Nico
+        self.state.emergency_boundary = None;
         self.last_line_data()
     }
 
@@ -246,6 +260,7 @@ impl<'a, B: Brush> BreakLines<'a, B> {
                         let is_newline = whitespace == Whitespace::Newline;
                         let is_space = whitespace.is_space_or_nbsp();
                         let boundary = cluster.info().boundary();
+                        let style = &self.layout.data.styles[cluster.data.style_index as usize];
 
                         if boundary == Boundary::Line {
                             // We do not currently handle breaking within a ligature, so we ignore boundaries in such a position.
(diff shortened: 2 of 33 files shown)


DJMcNab:
Wow, this is fantastic. I'm not really best placed to review this, but I do have a few initial thoughts.

That table is excellent! One thing which immediately jumps out at me is that using red/green for significant information is potentially challenging for colourblind users. I don't know of anyone on the team which this would impact, and I'm not suggesting that you should re-make the table (it looks like a lot of work...), but I'd recommend it as something to bear in mind in the future. (My understanding is that the general advice for having "side-channel" important information of this kind is to use e.g. patterns for differentiation; I'm not sure how practical that would be in this case...)

I'm not sure what to do about the new Korean tests. I guess showing that we lay things out correctly is quite cool, even without the actual glyphs being visible. If we can find a "sufficiently" small font with a suitable license, adding it to the repo is probably reasonable.

Are the tests standard? I see you link to [wpt.fyi](https://wpt.fyi/results/css/css-text/word-break/word-break-keep-all-003.html?label=experimental&label=master&aligned) for one of them. Should the rest have a similar source provided/linked?

valadaptive:
I've added links to the WPT page for the `word-break: keep-all` tests. Perhaps I should try implementing the other tests here as well? I think it would involve a lot of tedious manual porting.

As far as I can tell, the other tests (regarding which spans of text the CSS property "attaches" to) are original. I did just glance over at some of the WPT tests, and it looks like there's one (http://wpt.live/css/css-text/word-break/word-break-break-all-inline-007.tentative.html) that fails in all the browsers, and I *think* corresponds to this case:
image

I had to implement some weird behavior to replicate the browsers in that case (see the comment in `layout/context.rs` about applying each character's line breaking property to the *next* character), but actually just reverted it to the behavior that seems more sensible since WPT says it's the browsers' fault.

Regarding the Korean/Japanese tests, I decided not to include any CJK fonts since they're huge (Noto S

xorgy:
Thank you! I'm wondering how this will work with hyphen insertion (particularly with explicit hyphenation opportunities through `U+00AD SOFT HYPHEN`).

parley/src/style/mod.rs
```diff
@@ -22,6 +22,48 @@ pub enum WhiteSpaceCollapse {
...
+
+impl WordBreak {
+    pub(crate) fn as_swash(&self) -> swash::text::WordBreakStrength {
+        match self {
+            Self::Normal => swash::text::WordBreakStrength::Normal,
+            Self::BreakAll => swash::text::WordBreakStrength::BreakAll,
+            Self::KeepAll => swash::text::WordBreakStrength::KeepAll,
+        }
+    }
+}
```
xorgy, line 48: I think it makes sense to add the `default` to `WordBreakStrength` in Swash and reuse/reexport that enum here.

parley/src/tests/test_wrap.rs
```diff
@@ -0,0 +1,273 @@
...
+    test_text("Latin latin latin latin", "latin", 120.0);
+    // These will all show up as boxes because CJK fonts are quite large (several megabytes per language) and could
+    // bloat the repository. Line break analysis should work the same regardless of font, however.
+    test_text("日本語 日本語 日本語", "japanese", 60.0);
+    test_text("한글이 한글이 한글이", "korean", 60.0);
+    // TODO: we fail this test; so does Safari
+    // https://wpt.fyi/results/css/css-text/word-break/word-break-keep-all-003.html
+    // test_text("และ และและ", "thai", 65.0);
+    test_text("フォ フォ", "ID_and_CJ", 30.0);
+    test_text("애기판다 애기판다", "korean_hangul_jamos", 90.0);
```
xorgy, line 272: Might be good to add a little comment above this string to clarify it is decomposed, some people may not be familiar with jamo decomposition, and may accidentally normalize this string.

xorgy (changes requested):
I think it makes sense to consolidate the definitions of `WordBreakStrength` and just reuse/reexport that here.
Thank you!

parley/src/style/mod.rs
valadaptive: I'm a bit apprehensive to pass through more data types directly from Swash since the long-term plan is to move away from it*, but I suppose this enum isn't doing much right now, so I've done so.

*Also, I would like to add serde support to this crate at some point, and adding it to Swash as well would be a bit of an ordeal

parley/src/tests/test_wrap.rs
valadaptive: I've added a comment about it.

parley/src/style/mod.rs
xorgy: I see, that is a valid reason.

nicoburns: It would definitely be nice if we could get to the point where Swash wasn't in Parley's public API such that we could update Swash without it being a semver-breaking change for Parley. Not sure how far away from that we are atm.

nicoburns:
Regarding testing, it's worth noting that we can run some WPT tests via Blitz (which is using Parley for text layout). We can't run everything though, because some tests require JavaScript (and some are designed for manual verification), so it may well be worth porting some of them.

---
For running word-break WPT tests with Blitz (basically anything that's screenshot based):
- Clone https://github.com/DioxusLabs/blitz
- Run `cargo run --release --package wpt -- css/css-text/word-break`
- (if you have [just](https://github.com/casey/just) installed then you can use `just wpt css/css-text/word-break`)
- You can patch in a local version of Parley using the top-level (workspace) Cargo.toml

Currently we can run 57/103 `css/css-text/word-break` (of which 23 pass)
And 1079/1774 of all `css/css-text` tests (of which 314 pass)

parley/src/layout/line/greedy.rs
```diff
@@ -322,34 +344,21 @@ impl<'a, B: Brush> BreakLines<'a, B> {
...
-                                        // Else start a new line
-                                        if try_commit_line!(BreakReason::Emergency) {
-                                            self.state.cluster_idx += 1;
-                                            return self.start_new_line();
-                                        }
-                                    }
-                                } else {
-                                    self.state.append_cluster_to_line(next_x);
-                                    self.state.cluster_idx += 1;
+                            else if let Some(prev_emergency) =
```
nicoburns, line 347: Is it always correct to try "prev_boundary" first and then fallback to "prev_emergency"? Are there cases (`OverflowWrap::Anywhere`?) where we ought to the later "emergency" breakpoint even if an earlier "regular" breakpoint is available?

I wonder whether we ought to just have one "previous break point" state that just gets set in different conditions depending on the line-breaking styles?

parley/src/context.rs
```diff
@@ -102,8 +101,32 @@ impl<B: Brush> LayoutContext<B> {
...
+                style_idx += 1;
+            }
+            a.set_break_strength(word_break);
+
+            let Some((properties, boundary)) = a.next() else {
+                break;
+            };
+
+            self.info.push((CharInfo::new(properties, boundary), 0));
         }
```
nicoburns, line 130: This seems pretty awkward, but I guess fine...

parley/src/layout/line/greedy.rs
valadaptive: > Are there cases (`OverflowWrap::Anywhere`?) where we ought to use the later "emergency" breakpoint even if an earlier "regular" breakpoint is available?

Text with `OverflowWrap::Anywhere` should only use an "emergency" breakpoint if there are no other opportunities to break the line.

The name "anywhere" is a bit confusing since it may give the impression that it behaves like `word-break: break-all`. [but it's just `break-word` with the difference that it affects the min content size](https://developer.mozilla.org/en-US/docs/Web/CSS/overflow-wrap#anywhere).

I don't think there's any way to get around having two different previous break points since we need that fallback behavior.

parley/src/context.rs
valadaptive: Yeah, it's a bit unfortunate. I thought about other ways to do it in Swash, e.g. passing in the word break strength with each call to `next`, but that would be a breaking change there and make it no longer an iterator.

parley/src/layout/line/greedy.rs
nicoburns: > The name "anywhere" is a bit confusing since it may give the impression that it behaves like word-break: break-all

I see. Part of me feels like we ought not to stick to the CSS property names here and come up with something more sensible (but using the CSS properties is also very convenient for Blitz which can translate them 1:1 from Stylo)

nicoburns (approved):

parley/src/context.rs
nicoburns: I think you could make the analysis take an iterator over `(char, WordBreakStrength)` rather than just `char`? And you could have a different constructor for backwards compat.

nicoburns:
Blitz branch pointed at this PR: https://github.com/DioxusLabs/blitz/tree/word-break-overflow-wrap

This PR is currently only making 1 extra test pass. But I wouldn't read too much into that. I can see some tests failing like https://wpt.live/css/css-text/word-break/word-break-break-all-010.html. But that test uses `ch` units which Blitz doesn't currently support (need to get font metrics out of fontique for that!).

---
Notes on debugging WPT tests using Blitz:

To open a specific test in Blitz: `cargo run --release --package readme -- https://wpt.live/css/css-text/word-break/word-break-break-all-010.html`
To inspect:
- Press `alt-h` to enable inspector mode
- Click on a node to have some debug info logged to console.

For inline formatting contexts you'll get something like the following:

```
Text content: " XXXXX"
Inline Boxes:

Lines:
Line 0:
  RUN (x: 0, w: 80)
Line 1:
  RUN (x: 0, w: 80)
Line 2:
  RUN (x: 0, w: 80)
  ```

**Comment by valadaptive** [comment]
Now that swash 0.2.2 has been released, I don't think anything is blocking this other than another review.

File: parley/src/tests/utils/renderer.rs [hunk 5]
```diff
@@ -19,7 +19,15 @@ use tiny_skia::{Color, FillRule, Paint, PathBuilder, Pixmap, PixmapMut, Rect, Tr
 
 #[derive(Clone, Copy, Debug, PartialEq)]
 pub(crate) struct ColorBrush {
-    pub(crate) color: Color,
+    pub(super) color: Color,
```
waywardmonkeys, line 22: Why make this `super` (and why as part of this PR?)

parley/src/tests/utils/renderer.rs
```diff
@@ -19,7 +19,15 @@ use tiny_skia::{Color, FillRule, Paint, PathBuilder, Pixmap, PixmapMut, Rect, Tr
...
+}
+
+impl ColorBrush {
+    pub(crate) fn new(r: u8, g: u8, b: u8, a: u8) -> Self {
```
waywardmonkeys, line 26: Rather just see this take a `Color` argument and then you can pass in things from `color::palette::css` instead which also gives them better names than a series of constants.

parley/src/tests/utils/renderer.rs
valadaptive: I think this was a change that I forgot to undo

parley/src/tests/utils/renderer.rs
valadaptive: Didn't know those were available--done

xorgy (approved):
Thank you for your diligence!

xorgy:
> Thank you for making this more accessible, specifically to germans <3 (/s)

Patches welcome for real German word breaking rules!

PR #316, opened 2025-03-25
Add a `PlainEditor` function for accessing the current selection

mwcampbell:
I'm pretty sure I need this in order to implement Android's `InputConnection` interface.


DJMcNab (approved):
I haven't reasoned about this fully, but I can't see this having any negative repercussions. And as it's needed for the Android work, then I'm happy to unblock

PR #317, opened 2025-03-25
Add `raw_text` function to `PlainEditor`

mwcampbell:
I'm not sure it's always a good idea to exclude the IME preedit region from the externally exposed `PlainEditor` text buffer. In any case, having access to the raw buffer will make it easier to implement `InputConnection` on Android. @DJMcNab if you'd rather wait and review that code when it's ready, to see if it suggests other changes that should be made to the `PlainEditor` API, that's fine with me; I'll go ahead and use this branch in the android-view demo/test app.

### DIFF
--- a/parley/src/layout/editor.rs
+++ b/parley/src/layout/editor.rs
@@ -742,8 +742,11 @@ where
         }
     }
 
-    /// Returns the current selection.
-    pub fn selection(&self) -> &Selection {
+    /// Borrow the current selection. The indices returned by functions
+    /// such as [`Selection::text_range`] refer to the raw text buffer,
+    /// including the IME preedit region, which can be accessed via
+    /// [`PlainEditor::raw_text`].
+    pub fn raw_selection(&self) -> &Selection {
         &self.selection
     }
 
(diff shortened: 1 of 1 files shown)


xorgy (approved):
This makes sense, I forgot to include this.

parley/src/layout/editor.rs
```diff
@@ -846,6 +846,12 @@ where
         }
     }
 
+    /// Borrow the text content of the buffer, including the IME preedit
+    /// region if any.
```
DJMcNab, line 850:
```suggestion
    /// region if any.
    ///
    /// Application authors should generally prefer [`text`](Self::text). That method excludes the
    /// IME preedit contents, which are not meaningful for applications to access; the
    /// in-progress IME content is not itself what the user intends to write.
```

DJMcNab (approved):
It's entirely reasonable to leave escape hatches here. By the way, #316 is potentially broken if there is an active IME; maybe we should rename the method to `raw_selection` to indicate that we haven't thought about this.

PR #319, opened 2025-03-26
Add `is_done` method to `BreakLines`

valadaptive:
With the text wrapping styles implemented, I can implement ellipsis truncation in egui by simply creating another layout for the ellipsis. However, to know whether to show the ellipsis, I need to know if there's any text left to wrap. It would be nice to be able to do that without having to actually break another line and then revert it.


xorgy (approved):

DJMcNab:
@xorgy just FYI, `valadaptive` is now a Linebender member, and so is now able to merge her own PRs (so she has chance to do any last minute changes, etc.)

xorgy:
Thanks @DJMcNab I see that now.

Issue #325, opened 2025-03-30
Better support for custom line positioning


valadaptive:
There have been a few scenarios in integrating Parley with egui where I've thought "this would be easy to implement if only I could manually adjust lines' positions". For instance:
- Adjusting the starting X offset of a line because there's previous text on that same line. This can sort of be faked with an inline box, but there are a lot of complications around the fact that they can't be vertically aligned right now.
- Disabling line breaks entirely. egui has a single-line textbox widget that doesn't display newlines. If I could adjust the X and Y positions of lines, it would be pretty easy to just manually place one line after the other horizontally instead of vertically. If not, the alternatives would be to go over every code path that lets you set a textbox's contents and filter out newlines there, or to implement this type of layout in Parley itself.
- Not egui-related, but flowing text [into shapes](https://www.w3.org/TR/2004/WD-SVG12-20041027/flow.html) or around floated boxes. This requires manually positioning each span of text, since each one can start at an arbitrary x-coordinate and you can have multiple spans in a single row.

Right now, you can't just adjust the position of each line when rendering it because there are places in Parley that assume the lines will be laid out in a given way (the "cursor to point" and "point to cursor" methods, selection rectangle drawing, and AccessKit bounding boxes come to mind). There are often good reasons for Parley to make these assumptions--for instance, "cursor to point" can do a binary search over the Y-positions of each line, something which wouldn't work if they could be arbitrarily repositioned.

While the higher-level layout code is great for most use cases, I think an "escape hatch" to gain access to line-level granularity would be very useful and decrease the number of layout-related features that Parley itself would have to implement.

I think a good place to start would be moving some of the logic of the layout-dependent methods mentioned above into methods on `Line`. People who are fine with using the regular line layout could call into the existing methods on `Layout`, but those who need more control could call the `Line` methods and provide or apply the necessary position transformations themselves beforehand.

nicoburns:
> Adjusting the starting X offset of a line because there's previous text on that same line. 

Can you give an example of this? Why are these separate "lines" rather than two spans of text on the same line?

> Disabling line breaks entirely. egui has a single-line textbox widget that doesn't display newlines. If I could adjust the X and Y positions of lines, it would be pretty easy to just manually place one line after the other horizontally instead of vertically. If not, the alternatives would be to go over every code path that lets you set a textbox's contents and filter out newlines there, or to implement this type of layout in Parley itself.

[...]

PR #326, opened 2025-03-30
Allow specifying absolute line height

### LINKED ISSUE #320: Allow specifying line-height in absolute units more conveniently
Sort of a follow-up to https://github.com/linebender/parley/pull/297#issuecomment-2749219613:

> I think a better option would be to add support for absolute line height to the style settings--that way, if one wants to avoid the "jitter", they can provide a round number for the line height themselves.

But this seems like it would also be useful for other applications, either because they allow *their* API consumers to specify line height in absolute units or because they resolve the units themselves (e.g. Blitz).

Currently you can work around this by just dividing the resolved absolute line height by the font size, so Parley can multiply it again ([which is what Blitz does](https://github.com/DioxusLabs/blitz/blob/5a670fb7c3ef26454bebaed79ad3c39fe430994b/packages/blitz-dom/src/stylo_to_parley.rs#L49-L50)) but has the potential to introduce floating-point error and feels somewhat kludgey.

valadaptive:
Resolves #320.

Not sure if it's too annoying to always have to specify whether you want the line height to be relative or absolute, but I think this is nicer than the roundtrip of absolute -> relative -> absolute that happens currently if you need to specify absolute line height.

### DIFF
--- a/parley/src/style/mod.rs
+++ b/parley/src/style/mod.rs
@@ -22,6 +22,14 @@ pub enum WhiteSpaceCollapse {
     Preserve,
 }
 
+#[derive(Debug, Clone, Copy, PartialEq)]
+pub enum LineHeight {
+    /// Line height specified as a multiple of the font size.
+    Relative(f32),
+    /// Line height specified in absolute units.
+    Absolute(f32),
+}
+
 /// Properties that define a style.
 #[derive(Clone, PartialEq, Debug)]
 pub enum StyleProperty<'a, B: Brush> {
@@ -59,8 +67,8 @@ pub enum StyleProperty<'a, B: Brush> {
     StrikethroughSize(Option<f32>),
     /// Brush for rendering the strikethrough decoration.
     StrikethroughBrush(Option<B>),
-    /// Line height multiplier.
-    LineHeight(f32),
+    /// Line height.
+    LineHeight(LineHeight),
     /// Extra spacing between words.
     WordSpacing(f32),
     /// Extra spacing between letters.
@@ -104,8 +112,8 @@ pub struct TextStyle<'a, B: Brush> {
     pub strikethrough_size: Option<f32>,
     /// Brush for rendering the strikethrough decoration.
     pub strikethrough_brush: Option<B>,
-    /// Line height multiplier.
-    pub line_height: f32,
+    /// Line height.
+    pub line_height: LineHeight,
     /// Extra spacing between words.
     pub word_spacing: f32,
     /// Extra spacing between letters.
@@ -132,7 +140,7 @@ impl<B: Brush> Default for TextStyle<'_, B> {
             strikethrough_offset: Default::default(),
             strikethrough_size: Default::default(),
             strikethrough_brush: Default::default(),
-            line_height: 1.2,
+            line_height: LineHeight::Relative(1.2),
             word_spacing: Default::default(),
             letter_spacing: Default::default(),
         }
--- a/parley/src/style/mod.rs
+++ b/parley/src/style/mod.rs
@@ -170,3 +170,9 @@ impl<B: Brush> From<GenericFamily> for StyleProperty<'_, B> {
         StyleProperty::FontStack(f.into())
     }
 }
+
+impl<B: Brush> From<LineHeight> for StyleProperty<'_, B> {
+    fn from(value: LineHeight) -> Self {
+        Self::LineHeight(value)
+    }
+}
--- a/examples/swash_render/src/main.rs
+++ b/examples/swash_render/src/main.rs
@@ -79,7 +79,7 @@ fn main() {
         let root_style = TextStyle {
             brush: text_brush,
             font_stack,
-            line_height: 1.3,
+            line_height: parley::LineHeight::Relative(1.3),
             font_size: 16.0,
             ..Default::default()
         };
@@ -137,7 +137,7 @@ fn main() {
 
         // Set default font family
         builder.push_default(font_stack);
-        builder.push_default(StyleProperty::LineHeight(1.3));
+        builder.push_default(StyleProperty::LineHeight(parley::LineHeight::Relative(1.3)));
         builder.push_default(StyleProperty::FontSize(16.0));
 
         // Set the first 4 characters to bold
--- a/examples/tiny_skia_render/src/main.rs
+++ b/examples/tiny_skia_render/src/main.rs
@@ -76,7 +76,7 @@ fn main() {
 
     // Set default font family
     builder.push_default(GenericFamily::SystemUi);
-    builder.push_default(StyleProperty::LineHeight(1.3));
+    builder.push_default(StyleProperty::LineHeight(parley::LineHeight::Relative(1.3)));
     builder.push_default(StyleProperty::FontSize(16.0));
 
     // Set the first 4 characters to bold
--- a/examples/vello_editor/src/text.rs
+++ b/examples/vello_editor/src/text.rs
@@ -43,7 +43,7 @@ impl Editor {
         editor.set_text(text);
         editor.set_scale(1.0);
         let styles = editor.edit_styles();
-        styles.insert(StyleProperty::LineHeight(1.2));
+        styles.insert(StyleProperty::LineHeight(parley::LineHeight::Relative(1.2)));
         styles.insert(GenericFamily::SystemUi.into());
         styles.insert(StyleProperty::Brush(palette::css::WHITE.into()));
         Self {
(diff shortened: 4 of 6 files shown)


nicoburns (approved):
Seems reasonable to me. I'm not too worried about API ergonomics, as I suspect it will be largely other libraries/frameworks consuming Parley so I doubt this will need to be typed manually that much.

waywardmonkeys:
I would like another review or two on this one… timebix it to within 48 hours.

xorgy (commented):
This is a reasonable way to do this, but I would like an `impl From` like we have for font families.

waywardmonkeys (changes requested):
I am concerned about the ergonomics. I'm also not sure about having something where almost everything in Parley is absolute, but this one was relative, but now we want to make it do both.

I also don't like a `From` impl that returns a relative one since looking at the code, it won't be at all obvious which one it is, without reading docs or looking at the implementation.

This might be different if we had a `Percentage` or similar wrapper class or something.

Another option might be to make this always be absolute and require that the user of the API do the multiplication / scaling themselves.

I'd like to hear what @dfrg thinks about all of this.

(Marking as request changes to be sure we address this before something is merged. I'm not against some sort of change, just want us to get it right with consensus.)

xorgy:
For clarity, `From<f32>` was not what I had in mind, but rather `From<LineHeight> for StyleProperty` (like `From<GenericFamily> for StyleProperty`.

parley/src/style/mod.rs
```diff
@@ -22,6 +22,23 @@ pub enum WhiteSpaceCollapse {
...
+    /// Line height specified in absolute units.
+    Absolute(f32),
+}
+
+impl From<f32> for LineHeight {
+    fn from(value: f32) -> Self {
+        Self::Relative(value)
+    }
+}
+
```
xorgy, line 41: I was thinking `From<LineHeight> for StyleProperty`; `From<f32> for LineHeight` is ambiguous.

xorgy (changes requested):

valadaptive:
> For clarity, `From<f32>` was not what I had in mind, but rather `From<LineHeight> for StyleProperty` (like `From<GenericFamily> for StyleProperty`, where there is no ambiguity). The builders accept `impl Into<StyleProperty>`, which is why we are able to push a `GenericFamily` directly.

That would make much more sense :facepalm:

> I am concerned about the ergonomics. I'm also not sure about having something where almost everything in Parley is absolute, but this one was relative, but now we want to make it do both.

My main concern is that going from relative to absolute units while keeping the same type for both would result in existing code silently doing the wrong thing, and API consumers would have to track down all the places they're passing in either a `StyleProperty::LineHeight` or a `TextStyle`.

parley/src/style/mod.rs
```diff
@@ -22,6 +22,14 @@ pub enum WhiteSpaceCollapse {
     Preserve,
 }
 
+#[derive(Debug, Clone, Copy, PartialEq)]
+pub enum LineHeight {
+    /// Line height specified as a multiple of the font size.
```
DJMcNab, line 27: Is a relative line height ever affected by the selected font (not just now, but is that something we might theoretically ever want?). That to me would be the only reasonable argument for keeping a relative line height around long-term. See for example [`line-height: normal`](https://developer.mozilla.org/en-US/docs/Web/CSS/line-height#normal), for a case where this does change based on the font family.
The other arguments which I think can be dismissed:
- CSS overloads the `line-height` property to mean both
- It's a pretty bad breaking change

There is another argument, which is one of defaults: any default absolute line height is stupid.

DJMcNab (commented):
I don't really like where this PR takes us. I think it is a better state than before this PR, but I think that to move it where I'd want us to be would require a second breaking change.

But I'm also not really sure where I want us to be. My gut reaction is that we should only support absolute line heights, and that relative line heights are out-of-scope.
The problem with not having a reasonable default line height is that people will evaluate Parley, get an extremely broken rendering, and not know how to debug it.

To clarify, this PR isn't blocking forward progress in egui, right?
If that's the case, my gut reaction is to ask you to continue to use the (*awful*) workaround of pre-dividing by the font size for now.

I think this probably needs Chad's input.

parley/src/style/mod.rs
nicoburns: > Is a relative line height ever affected by the selected font (not just now, but is that something we might theoretically ever want?). That to me would be the only reasonable argument for keeping a relative line height around long-term.

It's not now in Parley, but it used to be (and I suspect this is why Parley currently uses relative line height). It was changed in https://github.com/linebender/parley/pull/84/files to match CSS (which defines line height relative to the nominal font size). Previously it took into account the metrics from the font.

Whether this is ever desirable I'm not sure.

dfrg:
If we’re going to make this configurable then it would be nice to offer an option to fully respect font metrics as the code originally did. 

That is, I’d like to see something like:

```rust
enum LineHeight {
    MetricsRelative(f32),
    SizeRelative(f32),
    Absolute(f32),
}
```

where the first is the original behavior, the second is the current behavior and the last does what this PR proposes. For the first two variants, the associated value is a multiplier. The default should be `MetricsRelative(1.0)` because this is generally what you want unless you’re dealing with CSS. 

In a better world, the line breaker would take a callback to determine line height so we don’t have to be opinionated but we’re not there yet.

valadaptive:
> The other arguments which I think can be dismissed:
> 
>   * CSS overloads the `line-height` property to mean both
> 
>   * It's a pretty bad breaking change
> 
> 
> There is another argument, which is one of defaults: any default absolute line height is stupid.

Is that third argument one that you're putting forward or dismissing?

> If we’re going to make this configurable then it would be nice to offer an option to fully respect font metrics as the code originally did.

Agreed. If others are fine with your proposed API, I could try implementing that. I think egui's old font layout code may also be metrics-relative?

> In a better world, the line breaker would take a callback to determine line height so we don’t have to be opinionated but we’re not there yet.

This ties into [some discussion](https://github.com/linebender/parley/issues/325#issuecomment-2764478652) in https://github.com/linebender/parley/issues/325 (from @nicoburns):

> My main observation beyond what you've written is that the [`BreakLines` struct](https://github.com/linebender/parley/blob/main/parley/src/layout/line/greedy.rs#L100) is already well setup for incremental layout where Parley lays content out up to a point, yields control flow back to the caller, and then the caller run it's own logic and calls back into Parley when it's ready.

This would definitely be very helpful. Right now, the line breaking code doesn't seem to even calculate line height until the `finish` method after all the lines have been broken (right now, unless I'm doing something wrong, the vertical advance returned by `BreakLines::break_next` is always 0.0 as a consequence). I'm not sure if this is intrinsic to the architecture or just how it happens to be implemented now, but it'd be nice to change.

Some CSS vertical alignment properties require two-pass layout vertically (they're relative to the bounds of the line, but the bounds of the line depend on vertical alignment, so items with those alignments are aligned in a second pass). Not sure how that would interact with the line breaking code if we want to support custom callbacks.

DJMcNab:
> > There is another argument, which is one of defaults: any default absolute line height is stupid.
>
> Is that third argument one that you're putting forward or dismissing?

I'm putting it forward; that is to say, I don't dismiss it out of hand. I don't class it as a "definite" blocker for an "absolute line height only" world, but it's an argument against it.

I think in that proposal, we can actually remove `SizeRelative` entirely. If we want a mode based on the font's metrics (which is pretty reasonable), then I agree that a default based on metrics is what we want. Although that does still have the issue of jumpy (/potentially overlapping(?)) rounded line heights.
I agree, if `BreakLines` would let you handle this, say by post-hoc specifying the "real height" of a line once it is broken, that would make the most sense. That way, you can implement whatever rounding you like even when you're metrics aware. But that needn't happen in this PR; it's orthogonal.

So concretely, my suggested way forward is to change this to:
```rust
enum LineHeight {
    MetricsRelative(f32),
    Absolute(f32),
}
```

nicoburns:
My take on `SizeRelative` is that I think it's nice from an API perspective (esp. if we're going to have an enum anyway) as it effectively documents how to achieve that effect. And it's a minimal cost to us from a maintenance perspective (a single multiply). But I don't feel super-strongly about this as it is indeed easy to build this on top of absolute line height.

DJMcNab (approved):
Adding `MetricsRelative` needn't block this PR, and it seems pretty settled that we want `LineRelative`

xStrom:
I agree with Nico that having `SizeRelative` is an easy ergonomics win if we're already going enum. The implementation burden is trivial and it doesn't really make API consumers' life harder, because resulting line heights are always just absolute `f32`, so no matching required.

Thus, I support Chad's API proposal of having all three variants.

valadaptive:
I'll let someone else redo this PR with whatever consensus exists. This seems like a quagmire and the current hack/workaround is good enough for my egui work.

PR #331, opened 2025-04-04
Don't reset cursor affinity on newlines

valadaptive:
Resolves https://github.com/linebender/parley/issues/298.

This bit of code was apparently here to ensure the cursor moved down a line when you typed a newline, but I've tested it (with both cursor affinities, and around wrapped lines) and didn't observe any change in behavior with it removed.

Since having this code here breaks RTL text navigation, and doesn't seem to do anything else, I think it can be removed now.

~~There is some weird (probably incorrect) behavior around inserting newlines within an RTL span--as far as I can tell, all of that behavior is pre-existing, and occurs with or without this code present. I'll file a new issue about that.~~ (After testing how browsers behave in this scenario, this appears to just be how it works. Bidirectional text is weird.)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -43,6 +43,7 @@ This release has an [MSRV] of 1.82.
 #### Parley
 
 - Fix text editing for layouts which contain inline boxes ([#299][] by [@valadaptive][])
+- Fix cursor navigation in RTL text sometimes getting stuck within a line ([#331][] by [@valadaptive][])
 
 ## [0.3.0] - 2025-02-27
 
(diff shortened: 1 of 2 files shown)


DJMcNab (approved):
I agree; I can understand both:
- Why this code made sense in the model we used before #170
- Why this code is no longer needed - it's the code around https://github.com/linebender/parley/blob/f9fc967b88f906c75de73222f51e010b2a50f1fc/parley/src/layout/cursor.rs#L281-L291

xorgy (approved):
I'm satisfied this is an improvement and not a regression. :+1:

DJMcNab:
We're not really good with testing discipline, but it would be good to land a test for #298.
It can probably be one of the usual cursor movement tests.

PR #335, opened 2025-04-12
fontique: Fix panic on macOS in debug mode.

NoahR02:
This commit fixes a panic on macOS by enabling the "relax-sign-encoding" feature for the "objc2" crate.

Relevant issues in the objc2 crate:
https://github.com/madsmtm/objc2/issues/566
https://github.com/madsmtm/objc2/pull/567

I made sure to use version of objc2 that the current objc2-*  variations use, so that we don't duplicate the crate.
Screenshot 2025-04-11 at 8 51 30 PM

---
Additional Info:

System Info: Mac Mini M4, 16 GB of RAM.

Before the fix, when running the vello-editor or any example:
Screenshot 2025-04-11 at 9 12 02 PM

After the fix:
Screenshot 2025-04-11 at 9 14 29 PM

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -907,6 +907,7 @@ dependencies = [
  "icu_locid",
  "icu_properties",
  "memmap2",
+ "objc2 0.6.0",
  "objc2-core-foundation",
  "objc2-core-text",
  "objc2-foundation 0.3.0",
--- a/fontique/Cargo.toml
+++ b/fontique/Cargo.toml
@@ -26,6 +26,7 @@ system = [
     "std",
     "dep:windows",
     "dep:windows-core",
+    "dep:objc2",
     "dep:objc2-core-foundation",
     "dep:objc2-core-text",
     "dep:objc2-foundation",
@@ -53,6 +54,10 @@ windows = { version = "0.58.0", features = [
 windows-core = { version = "0.58", optional = true }
 
 [target.'cfg(target_vendor = "apple")'.dependencies]
+# FIX: Enable relax-sign-encoding to prevent the bug described in this issue: https://github.com/madsmtm/objc2/issues/566
+objc2 = { version = "0.6.0", optional = true, features = ["std", "relax-sign-encoding"] }
+# NOTE: When updating objc2-foundation, objc2-core-foundation, or objc2-core-text make sure to use the version of objc2
+# that they reference to prevent crate duplication.
 objc2-foundation = { version = "0.3.0", optional = true, default-features = false, features = [
     "alloc",
     "NSArray",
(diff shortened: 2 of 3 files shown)


xorgy:
Could you document the issue a bit more? I can't reproduce (`vello_editor` in main just runs fine for me on an M1 Mac Mini running Sequoia).

NoahR02:
> Could you document the issue a bit more? I can't reproduce (`vello_editor` in main just runs fine for me on an M1 Mac Mini running Sequoia).

From what I understand after reading this issue: [https://github.com/madsmtm/objc2/issues/566](https://github.com/madsmtm/objc2/issues/566), Swift tries to use NSInteger instead of NSUInteger  for `countByEnumeratingWithState_objects_count`'s return type.

Here is a link to the obj-c docs https://developer.apple.com/documentation/foundation/nsfastenumeration/countbyenumerating(with:objects:count:)?language=objc and the FFI binding function is listed below:
image

The panic is happening because of this line which likely interprets the function's return type as a signed integer:
image

You can ultimately track down this panic to this verification error, which is getting back a LongLong instead of the expected ULongLong.
image
        
image

---

I'm not too sure why this error doesn't persist on your side as I'm also running Sequoia 15.4. 

They likely can't work around this if I had to guess since NSUInteger is valid in obj-c, but Swift maps it to `Int` which gets mapped to a `__swift_intptr_t`. https://github.com/swiftlang/swift/blob/24814c5e9da1552e3886bf380c53da1f10d8335a/lib/ClangImporter/MappedTypes.def#L154-L156 . 

Sidenote:
The q and Q in the panic message are referring to their Encoding enum:
image

NoahR02:
One more important piece of information that I just discovered is that this only happens in debug mode, release mode works fine. I'm guessing that the objc2 library is not checking the method signature in release mode and that is why it works in the release build.

jamesthurley:
FWIW, I just ran into this as well. I started getting the error:
```
expected return to have type code 'q', but found 'Q'
```
from `objc2` via `[fontique](fontique::scan::scan_paths)`

I'm not sure what changed to trigger it, but searching for the error lead me to this 13 hour old PR, which seems like quite a coincidence. I'm on an M3 MacBook Air.

xorgy:
Ah, that'd do it. I rarely run in debug mode.

xorgy (approved):

xorgy (commented):
This works, but we want to check that this is still the best known solution.

NoahR02:
> This works, but we want to check that this is still the best known solution.


This first image below is one that I posted last week, that shows that the code that panics goes into the objc2_foundation library when we loop through the `paths` variable which calls its `objc2` iterator. There may be a better solution, but that solution would have to be made in the objc2_* libraries themselves from what I know.  I searched their issues and pull requests and it seems like the `relax-sign-encoding` feature is currently still the best solution.
image

image

xorgy (approved):
Thank you for your submission, and your research on the topic. 
Feel free to merge.

PR #338, opened 2025-04-21
fontique: Update the objc2 crates

waywardmonkeys:
The updated versions have a nicer API and one less unsafe.

### DIFF
--- a/fontique/src/backend/coretext.rs
+++ b/fontique/src/backend/coretext.rs
@@ -7,12 +7,8 @@ use super::{
 use alloc::sync::Arc;
 use core::ptr::{null, null_mut};
 use hashbrown::HashMap;
-use objc2_core_foundation::{CFDictionaryCreate, CFRange, CFRetained, CFString, CFStringGetLength};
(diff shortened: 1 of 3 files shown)


fontique/src/backend/coretext.rs
```diff
@@ -72,23 +68,23 @@ impl SystemFonts {
         let key = key.into();
         let sample = key.script().sample()?;
         let font = create_fallback_font_for_text(sample, key.locale(), false)?;
-        let family_name = unsafe { CTFontCopyFamilyName(&font) };
+        let family_name = unsafe { font.family_name() };
```
DJMcNab, line 71: Is this unsafe only because the safety invariants haven't been checked on the objc2 side?

DJMcNab (approved):
Seems reasonable

fontique/src/backend/coretext.rs
waywardmonkeys: They have to be analyzed on the objc2 side by a human and marked as being able to be safe.

PR #346, opened 2025-05-05
Enable Parley/Fontique to compile to wasm with default features enabled

nicoburns:
## Motivation 

Enable Parley and Fontique to compile to WASM with default features enabled. This improves discoverability of the fact that it is even possible to compile to WASM.

## Changes made

- On platforms which do not have a real system font backend, enable the dummy backend even when the `system` feature is enabled.
- Add a couple of missing method stubs to the dummy backend

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -14,6 +14,10 @@ env:
   RUST_MIN_VER_PKGS: "-p parley -p fontique"
   # List of features that depend on the standard library and will be excluded from no_std checks.
   FEATURES_DEPENDING_ON_STD: "std,default,system"
+  # List of packages that can not target Wasm.
+  NO_WASM_PKGS: "--exclude xtask"
+  # List of packages that can not target Android.
+  NO_ANDROID_PKGS: "--exclude xtask"
 
 
 # Rationale
@@ -98,7 +102,7 @@ jobs:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        os: [ windows-latest, macos-latest, ubuntu-latest ]
+        os: [windows-latest, macos-latest, ubuntu-latest]
     steps:
       - uses: actions/checkout@v4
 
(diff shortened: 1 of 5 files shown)


waywardmonkeys:
Please ping me in 16 hours if I don’t respond further.

waywardmonkeys (commented):
I am roughly in favor of this general approach. Need to think about it a little more.

xorgy:
Yeah, I think on wasm there will never be a true `System’ but it could to some extent be handled through globals.

DJMcNab (commented):
Thanks - I think making the fact that we support BYOF on WASM is supported clear is worthwhile.

Could we warn if you use this backend and the system feature is enabled?
I worry that it's not discoverable why fonts are not being selected at all. A naïve consumer wouldn't be wrong for expecting this to work, as our documentation suggests that if you have the system feature enabled, you will get system fonts. And the web platform does clearly support fonts (as evidenced by you being able to read this comment on GitHub's page), but it just so happens that functionality isn't exposed to the web.

Should there be a corresponding CI change? The WASM + system is I think a new configuration which didn't work before.

nicoburns:
> Yeah, I think on wasm there will never be a true `System’ but it could to some extent be handled through globals.

Well there is https://developer.mozilla.org/en-US/docs/Web/API/Local_Font_Access_API. Although it's looking like this will never be supported anywhere except Chrome. So perhaps not worth bothering with. I suspect the long-term "system font experience" for wasm might involve loading fonts from a web server that does on-demand font-subsetting using something like https://github.com/googlefonts/fontations/tree/main/incremental-font-transfer

> Could we warn if you use this backend and the system feature is enabled?

I would prefer not to warn as I don't think there will be an easy way to turn the warning off if you want to accept it (disabling a feature selectively by platform isn't well supported by Cargo). Could we just document it? Taffy uses a docs-rs-only feature flag to automatically pull comments on feature flags from `Cargo.toml` into `lib.rs` docs (looks like this: https://docs.rs/taffy/latest/taffy/#feature-flags) Perhaps we could do the same for Parley and then this would be a good place to document this?

DJMcNab:
I expect that turning the warning off would involve disabling `system_fonts` in the `CollectionOptions`, which indicates that you're aware that there won't be system fonts and so will be providing them yourselves.

valadaptive (approved):
This is also going to be necessary for egui (which supports WASM).

FWIW, I'd rather not have any runtime warning if system fonts are enabled in WASM. The API consumer would have to insert their own check to disable the option if they're running in WASM...but the option does nothing in WASM, so the sole purpose of the check would be enshrining "I read the warning" in the code itself. Maybe just note the currently-supported platforms in the `Collection::new` documentation?

waywardmonkeys (commented):
I'd like to see this go in, but it'll need CI before approving it.

xStrom (approved):
I added CI changes to this PR and can confirm that it passes Wasm CI checks.

waywardmonkeys (approved):

DJMcNab:
> FWIW, I'd rather not have any runtime warning if system fonts are enabled in WASM. The API consumer would have to insert their own check to disable the option if they're running in WASM...but the option does nothing in WASM, so the sole purpose of the check would be enshrining "I read the warning" in the code itself

My point is more than the user manually loading fonts would also be behind that same check. So in effect, it's a check which they already have to do. In my mind, giving a warning and explanation of why "I've tried to use this, and I just get no rendering at all" happens, and a path towards fixing it, is good practise.
But I see this has been overruled, which is fine - that's how the process works.

waywardmonkeys:
@DJMcNab What you just described with it only logging when the bad scenario happens sounds reasonable and I would welcome that as a follow up PR.

PR #359, opened 2025-05-19
Change reprs of Font{Width, Weight, Style}; support Eq, Ord and Hash

dhardy:
All inputs to a query now support `Eq` and `Hash`. (Motivation: I hash [`FontSelector`](https://docs.rs/kas-text/latest/kas_text/fonts/struct.FontSelector.html) and am looking at switching this to Fontique.)

The changes to `FontWidth` are entirely internal aside from the more limited precision. I would be surprised if the result does not have sufficient precision anywhere. I used fixed-point using a denominator of 256 since the predefined values factor nicely, though it could use a denominator of 100 as fontconfig does.

The changes to `FontWeight` are **not** only internal, and a couple of examples were changed. I think this is still preferable however.

The changes to `FontStyle` are also not internal due to the way the enum works... on second thoughts, it may be preferable to use a new-type over the `i16`. Otherwise this is similar to `FontWidth`.

### DIFF
--- a/examples/swash_render/src/main.rs
+++ b/examples/swash_render/src/main.rs
@@ -69,7 +69,7 @@ fn main() {
     let text_brush = ColorBrush { color: text_color };
     let brush_style = StyleProperty::Brush(text_brush);
     let font_stack = FontStack::from("system-ui");
-    let bold_style = StyleProperty::FontWeight(FontWeight::new(600.0));
+    let bold_style = StyleProperty::FontWeight(FontWeight::new(600));
     let underline_style = StyleProperty::Underline(true);
     let strikethrough_style = StyleProperty::Strikethrough(true);
 
--- a/examples/tiny_skia_render/src/main.rs
+++ b/examples/tiny_skia_render/src/main.rs
@@ -83,7 +83,7 @@ fn main() {
     builder.push_default(StyleProperty::FontSize(16.0));
 
     // Set the first 4 characters to bold
-    let bold = FontWeight::new(600.0);
+    let bold = FontWeight::new(600);
     builder.push(StyleProperty::FontWeight(bold), 0..4);
 
     // Set the underline & strikethrough style
--- a/parley/src/lib.rs
+++ b/parley/src/lib.rs
@@ -42,7 +42,7 @@
 //! builder.push_default(StyleProperty::FontSize(16.0));
 //!
 //! // Set a style that applies to the first 4 characters
-//! builder.push(StyleProperty::FontWeight(FontWeight::new(600.0)), 0..4);
+//! builder.push(StyleProperty::FontWeight(FontWeight::new(600)), 0..4);
 //!
 //! // Add a box to be laid out inline with the text
 //! builder.push_inline_box(InlineBox { id: 0, index: 5, width: 50.0, height: 50.0 });


xorgy:
While I personally don't mind integer weight (even when it's driving a variable property), I think that opinions differ on this. As an alternative, you could implement Ord manually using `total_cmp` and it'd probably be fine.

dhardy:
I fully expect opinions to differ on this. Mine is that the differences are mostly unimportant, but slightly preferable in that using FP values like `500.0` before seemed messy, and in that styles are better quantized.

I care more about `Hash` than `Ord`.

nicoburns:
I don't know whether an integer or float representation for the styles is better, but you should be able to implement Hash for float with a manual impl too.

valadaptive (changes requested):
If I'm reading the OpenType spec correctly, then [font variation axes use 16.16 fixed-point](https://learn.microsoft.com/en-us/typography/opentype/otspec183/otvaroverview#coordinate-scales-and-normalization).

If we *are* going to change the internal representation, I think using something like [`font_types::Fixed`](https://docs.rs/font-types/latest/font_types/struct.Fixed.html) would be ideal (we already depend on `font_types` via `skrifa` and `read-fonts`), since it matches the OpenType spec.

Swash, the current shaping engine, uses floats for the public API and converts to fixed-point internally, so this would result in a few back-and-forth conversions, but I believe the plan is to replace Swash eventually anyway.

Alternatively, the [`ordered-float` crate](https://lib.rs/crates/ordered-float) could provide `Hash + Ord` floating-point types to use internally.

I'll leave it up to others what the best course of action is, but I don't think we should define an ad-hoc fixed-point type as part of this project.

dhardy:
> but you should be able to implement `Hash` for (a wrapper of a) float with a manual impl too

Possible, but there are footguns with that — e.g. -0.0 and +0.0 compare equal so need to have equal hash despite different binary repr, besides the NaN thing. So avoiding FP types is preferred.

In general, I think fixed-point types make more sense than FP for fonts. I haven't looked too closely at exactly how much precision is required in these cases, but the 8.8 bit precision used here is probably sufficient?

dhardy:
> If I'm reading the OpenType spec correctly, then [font variation axes use 16.16 fixed-point](https://learn.microsoft.com/en-us/typography/opentype/otspec183/otvaroverview#coordinate-scales-and-normalization).

See also <https://learn.microsoft.com/en-us/typography/opentype/otspec183/fvar>. It looks like `Fixed` is 16.16 here, with only one example (62.5) using any non-integral values. Hence `Fixed` is a generally-applicable format for variations, not an optimal type for any specific variation or `Tag`.

Using `Fixed` in place of all the `f32` values this PR tries to replace seems acceptable, though not optimal. Specifically:

1. The normalised width is typically between 0.5 and 2. [usWidthClass](https://learn.microsoft.com/en-us/typography/opentype/spec/os2#uswidthclass) values are typically from 50, 62.5, ..., 200. The standard normalized values (0.5, 0.625, ...) only use 2.3 bits of precision, so a type like `font_types::F6Dot10` provides plenty of precision.
2. Standard weight values are 100, 200, ... 900, with [usWeightClass](https://learn.microsoft.com/en-us/typography/opentype/spec/os2#usweightclass) denoting the range `1..=1000` as valid. We can just use `u16` here.
3. [Slant](https://learn.microsoft.com/en-us/typography/opentype/otspec183/dvaraxistag_slnt) values must be greater than -90 and less than +90. No indication that fractional precision is required. So i8 is sufficient. There's no appropriate `font_types` type here, unless we change units to e.g. radians.

But *is it preferable to size-optimize these types*?

- This makes it cheaper to cache a number of font faces with variations (likely insignificant).
- It provides a degree of quantization, but often explicit quantization is still needed.

Also fine to just use `font_types::Fixed` for all of the above then. (`font_types` is an existing dependency of `read-fonts` of `fontique`.)

Final point: why is `FontStyle::Oblique`'s payload optional? Why not make it required but with a `const FontStyle::OBLIQUE: Self = FontStyle::Oblique(14.into())` value?

valadaptive:
> [Slant](https://learn.microsoft.com/en-us/typography/opentype/otspec183/dvaraxistag_slnt) values must be greater than -90 and less than +90. No indication that fractional precision is required. So i8 is sufficient. There's no appropriate font_types type here, unless we change units to e.g. radians.

At least in Firefox, fractional slant values do have an effect on variable fonts with the `slnt` axis (e.g. Roboto Flex). [The Google Fonts tester](https://fonts.google.com/specimen/Roboto+Flex/tester) has a slider that limits you to integer values, but if you remove this restriction using "inspect element", you can observe fractional slant values.

> But is it preferable to size-optimize these types?

I'd say it isn't. It's better for them to map 1:1 to an OpenType variation axis, even if their "expected" ranges are narrower or your specific use case requires less precision. This also avoids the risk of misjudging their expected ranges or precision (see my previous comment about fractional slant). I also just dislike the idea of removing functionality as part of a micro-optimization.

I'd wait for more opinions before making any changes to the PR; I'm not speaking from any position of authority and other people might have different opinions.

nicoburns:
We should probably get @dfrg's opinion on this.

A note that we should try to avoid using `font-types` types directly in our public API as we want to be able to bump `skrifa`/`read-fonts`/`font-types` in non-semver-breaking releases (as these crates currently bump their version very frequently). Our own versions of the same types would be fine.

dfrg:
In general, I’d prefer not to quantize these values. As mentioned, variations enable extended precision and we should support that.

I’d be happy to accept a PR adding custom Hash and Eq implementations for these though. We can do the simple thing and avoid handling denormals as we’re not likely to encounter them in these types.

Also, +1 to avoiding fontations types in the public API.

dhardy:
> I’d be happy to accept a PR adding custom Hash and Eq implementations for these though. We can do the simple thing and avoid handling denormals as we’re not likely to encounter them in these types.

Sub-normals aren't a problem; they're ordered and equality works as expected. There are two special FP cases that are potentially a problem: +/- 0.0 compares equal despite differing binary reprs, while NaN compares unequal (and has many binary representations). You end up needing something like [`OrderedFloat`](https://docs.rs/ordered-float/latest/ordered_float/struct.OrderedFloat.html) to do this properly (if violating IEEE754 can be considered doing it properly).

Personally I consider fixed-point far superior for this use-case. If you don't wish to re-export `font_types::Fixed`, we can copy it (reducing the API surface to only actually-used fns).

dhardy:
I had a look at existing fixed-point crates:

- [fixed](https://crates.io/crates/fixed) is somewhat large, actively developed and looks a very high quality. It contains a lot, so if used it might make sense to directly re-export just [`fixed::types::I16F16`](https://docs.rs/fixed/1.29.0/fixed/types/type.I16F16.html) and the [prelude traits](https://docs.rs/fixed/1.29.0/fixed/prelude/index.html). There is an alpha 2.0 version dependent on [a nightly feature](https://github.com/rust-lang/rust/issues/76560).
- [ufix](https://crates.io/crates/ufix) is smaller, heavily uses generics, and has some interesting design choices (automatically increasing type precision on arithmetic operations). I don't think it's a good fit.
- [`font_types::Fixed`](https://docs.rs/font-types/latest/font_types/struct.Fixed.html) is more-or-less what's required. I disagree with a few design choices (e.g. `Fixed::from_i32` silently discards the high 16 bits).
- I considered writing my own fixed `16.16` format. It wouldn't be too hard (eschewing all generics to focus on a fixed `16.16` format). But it would essentially be copying a lot of `fixed::types::I16F16`.

nicoburns:
> Personally I consider fixed-point far superior for this use-case. If you don't wish to re-export font_types::Fixed, we can copy it (reducing the API surface to only actually-used fns).

This would probably be my preferred approach (copy-pasting `font_types::Fixed`).

I think avoiding "large" (and nightly) is going to be important here, and I suspect the maintenance involved with our own `Fixed` type will be minimal in practice.

DJMcNab:
If this hashing is just used for optimisation, then I'd suggest using the same approach we did in Color - https://docs.rs/color/latest/color/cache_key/index.html is pretty generic (but unfortunately is locked in the Color crate)

dhardy:
I could still do as proposed here, but I think I am better off using my own types with mappings into `fontique::FontWidth` etc, making my original motivation moot.

[This issue](https://github.com/googlefonts/fontations/issues/693) indicates that fixed precision may not always be appropriate for fonts.

So, do you still want this PR?

DJMcNab:
I think having some support for using these in cache key use cases would be good, but I don't think we have the bandwidth to work through the tradeoffs, if it's not blocking you.

DJMcNab:
@dhardy are you familiar with the [office hours](https://xi.zulipchat.com/#narrow/channel/359642-office-hours). It would be good to have you join so that we can discuss what you need from Parley.

DJMcNab:
Yes, that's right. It was a suggestion in the meeting to make sure that you're aware of them.
Hopefully we'll see you in a future edition

PR #362, opened 2025-05-20
Add absolute and metrics-relative line height styles

valadaptive:
This implements @dfrg's [proposal](https://github.com/linebender/parley/pull/326#issuecomment-2766033007), since most people agreed it was the way forward.

If you don't want `FontSizeRelative`, I can remove it--`MetricsRelative(1.0)` is a sensible default that works at all font sizes. But some people leaned towards keeping `FontSizeRelative`.

The default line height has been changed from `FontSizeRelative(1.2)` (what I believe CSS uses) to `MetricsRelative(1.0)`. This means a lot of the snapshot tests will change as well.


examples/swash_render/src/main.rs
```diff
@@ -141,7 +141,9 @@ fn main() {
 
         // Set default font family
         builder.push_default(font_stack);
-        builder.push_default(StyleProperty::LineHeight(1.3));
+        builder.push_default(StyleProperty::LineHeight(parley::LineHeight::SizeRelative(
+            1.3,
+        )));
```
nicoburns, line 146: Could we add an `Into` from `LineHeight` to `StyleProperty::LineHeight` to make this a little less verbose?

nicoburns (commented):
This generally looks good to me. A few minor points in comments, none of which I'd consider blocking.

My main question would be why so many (all?) of the snapshot tests have changed. Has the default line-height changed? In theory it seems like nothing here should change existing snapshots as we are only adding additional line-height options.

valadaptive:
> Has the default line-height changed?

Yes, Chad suggested that in the original proposal. I can revert that if it touches too many tests.

parley/src/style/mod.rs
```diff
@@ -41,6 +41,16 @@ pub enum OverflowWrap {
     BreakWord,
 }
 
+#[derive(Debug, Clone, Copy, PartialEq)]
+pub enum LineHeight {
+    /// Line height specified relative to what's defined in the font's metrics.
+    MetricsRelative(f32),
+    /// Line height specified as a multiple of the font size.
+    SizeRelative(f32),
```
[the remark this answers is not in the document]
valadaptive: Makes sense.

examples/swash_render/src/main.rs
valadaptive: I think that impl exists, I just did a quick search-and-replace over the examples and forgot to update them to actually use it.

nicoburns:
> > Has the default line-height changed?
> 
> Yes, Chad suggested that in the original proposal. I can revert that if it touches too many tests.

I see. I don't feel particularly strongly about what the default is (I'm not planning to use it anyway!). But could you document what the new default is and the previous default was (in the PR description), and perhaps add a note that this is why the snapshots have changed?

valadaptive:
Done.

parley/src/resolve/mod.rs
```diff
@@ -486,7 +489,17 @@ impl<B: Brush> ResolvedStyle<B> {
...
+            LineHeight(value) => match (self.line_height, *value) {
+                (crate::LineHeight::MetricsRelative(a), crate::LineHeight::MetricsRelative(b))
+                | (
+                    crate::LineHeight::FontSizeRelative(a),
+                    crate::LineHeight::FontSizeRelative(b),
+                )
+                | (crate::LineHeight::Absolute(a), crate::LineHeight::Absolute(b)) => {
+                    nearly_eq(a, b)
+                }
+                _ => false,
```
nicoburns, line 501: A couple of notes on this one:
- Consider importing `crate::LineHeight` at the top of the file (partly for consistency with surrounding code, partly because I think the `match` would format better like that)
- Consider making this a method on `LineHeight`.

parley/src/resolve/mod.rs
```diff
@@ -499,7 +512,15 @@ impl<B: Brush> ResolvedStyle<B> {
...
-            line_height: self.line_height * self.font_size,
+            line_height: match self.line_height {
+                crate::LineHeight::MetricsRelative(value) => {
+                    crate::LayoutLineHeight::MetricsRelative(value)
+                }
+                crate::LineHeight::FontSizeRelative(value) => {
+                    crate::LayoutLineHeight::Absolute(value * self.font_size)
+                }
+                crate::LineHeight::Absolute(value) => crate::LayoutLineHeight::Absolute(value),
+            },
```
nicoburns, line 523: Same as `nearly_eq` above (considering `use crate::LineHeight` + make this a method).

nicoburns (changes requested):
Apologies, I have a couple more changes to request:
1. A changelog entry:
   - Should note the introduction of the `Lineheight` enum
   - Should note the change of default line-height
2. Suggestions in comments are optional but recommended.

nicoburns (approved):

parley/src/style/mod.rs
```diff
@@ -41,6 +43,42 @@ pub enum OverflowWrap {
...
+        }
+    }
+
+    pub(crate) fn resolve(&self, scale: f32) -> LayoutLineHeight {
+        match self {
+            Self::MetricsRelative(value) => LayoutLineHeight::MetricsRelative(*value),
+            Self::FontSizeRelative(value) => LayoutLineHeight::Absolute(*value * scale),
+            Self::Absolute(value) => LayoutLineHeight::Absolute(*value),
+        }
+    }
```
xStrom, line 79: The input seems to be font size, not scale.

```suggestion
    pub(crate) fn resolve(&self, font_size: f32) -> LayoutLineHeight {
        match self {
            Self::MetricsRelative(value) => LayoutLineHeight::MetricsRelative(*value),
            Self::FontSizeRelative(value) => LayoutLineHeight::Absolute(*value * font_size),
            Self::Absolute(value) => LayoutLineHeight::Absolute(*value),
        }
    }
```

parley/src/layout/line/greedy.rs
```diff
@@ -523,7 +523,6 @@ impl<'a, B: Brush> BreakLines<'a, B> {
                         // Compute the run's vertical metrics
                         line.metrics.ascent = line.metrics.ascent.max(run.metrics.ascent);
                         line.metrics.descent = line.metrics.descent.max(run.metrics.descent);
-                        line.metrics.leading = line.metrics.leading.max(run.metrics.leading);
```
xStrom, line 526: Why aren't we calculating the line's leading any more? Unclear to me.

xStrom (commented):
Great to see absolute line height support moving forward!

I have a few inline comments.

On a more general note, this PR really shows that using default line height for snapshot tests is a bad idea. This type of PR would benefit greatly from zero changed snapshots. Anyway, the actual line height tests did manually specify a line height and still work, so the validity is just hidden in noise. I would prefer setting the old default line height as the default choice for tests to reduce the snapshot churn. However it's not a blocker.

parley/src/layout/line/greedy.rs
valadaptive: Ah, I probably should've mentioned this in the PR itself so more people could take a look at it. I believe that this is dead code now, since leading is always calculated [here](https://github.com/linebender/parley/blob/5be9994032e446bb9fbb893324631329e3e1f1f3/parley/src/layout/line/greedy.rs#L607-L608), and both of the removed leading-calculation lines should flow into that code unconditionally.

xStrom: Thanks and I think you're correct. Good catch!

xStrom (approved):
This moves us forward nicely, thanks for your continued efforts!

dfrg:
> This moves us forward nicely, thanks for your continued efforts!

+1, much appreciated!

valadaptive:
Updated the tests to use a default line height of `FontSizeRelative(1.0)`.

One last (semi-related) thing: it seems the default line height used to be 1.2 if you used the `TreeStyleBuilder`, but 1.0 if you used the `RangedStyleBuilder`. This is seemingly due to a discrepancy between [the `Default` impl for `TextStyle`](https://github.com/linebender/parley/blob/5be9994032e446bb9fbb893324631329e3e1f1f3/parley/src/style/mod.rs#L162) and [the one for `ResolvedStyle`](https://github.com/linebender/parley/blob/5be9994032e446bb9fbb893324631329e3e1f1f3/parley/src/resolve/mod.rs#L430).

In the future, we should probably find a way to avoid introducing further discrepancies of that sort, but for now I've listed it in the changelog. Only a couple tests used `TreeStyleBuilder`, and their snapshots now differ.

nicoburns:
> Only a couple tests used TreeStyleBuilder, and their snapshots now differ.

Perhaps we ought to set the line-height explicitly in those specific tests?

parley/src/layout/mod.rs
```diff
@@ -280,6 +280,23 @@ pub struct Line<'a, B: Brush> {
...
+pub(crate) enum LayoutLineHeight {
+    MetricsRelative(f32),
+    Absolute(f32),
+}
+
+impl LayoutLineHeight {
+    pub(crate) fn resolve(&self, run: &RunData) -> f32 {
+        match self {
+            Self::MetricsRelative(value) => {
+                (run.metrics.ascent + run.metrics.descent + run.metrics.leading) * value
```
DJMcNab, line 293: I only just found out what `leading` is - it seems to be made for exactly this use case!

I wonder if `value` should actually be a multiplier on `leading`. It's probably too cute. Landing this as-is would be fine

parley/src/style/mod.rs
```diff
@@ -41,6 +43,42 @@ pub enum OverflowWrap {
     BreakWord,
 }
 
+#[derive(Debug, Clone, Copy, PartialEq)]
+pub enum LineHeight {
+    /// Line height specified relative to what's defined in the font's metrics.
```
DJMcNab, line 48:
```suggestion
    /// The line's height is a multiple of the ["line height"](https://developer.apple.com/library/archive/documentation/StringsTextFonts/Conceptual/TextAndWebiPhoneOS/TypoFeatures/TextSystemFeatures.html) defined by the font's metrics.
```
There might be a better link for Line height available. The picture in https://tex.stackexchange.com/questions/614909/pk-height-ascend-descend-and-cap-height-of-a-pk-font-face is pretty good.

DJMcNab (approved):
Code changes look really good. The only actionable changes are some docs/examples suggestions, which should be straightforward to implement.

examples/swash_render/src/main.rs
```diff
@@ -82,7 +82,7 @@ fn main() {
         let root_style = TextStyle {
             brush: text_brush,
             font_stack,
-            line_height: 1.3,
+            line_height: LineHeight::FontSizeRelative(1.3),
```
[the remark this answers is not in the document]
xStrom: I think that the example showing how to set a line height for the root style has educational value. I would keep it specified.

parley/src/style/mod.rs
```diff
@@ -41,6 +43,42 @@ pub enum OverflowWrap {
...
+    MetricsRelative(f32),
+    /// Line height specified as a multiple of the font size.
+    FontSizeRelative(f32),
+    /// Line height specified in absolute units.
```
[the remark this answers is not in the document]
xStrom: If we were to give more insight then I think the concept of distributing the lines to fit container height will be the more useful piece of knowledge.

```suggestion
    /// Line height specified in absolute units.
    /// Useful for precisely matching layout container height.
```

parley/src/tests/test_lines.rs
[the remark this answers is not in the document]
xStrom: That's right, the test code was temporary until the arrival of absolute line height capability. However, I'm willing to do the test change PR as follow-up, so I wouldn't block on it here.

parley/tests/snapshots/lines_line_height_size_relative-0.png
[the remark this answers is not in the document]
xStrom: Yeah the new font size tests don't apply styles, so it is expected.

parley/src/layout/mod.rs
xStrom: For what it's worth, this summing is how the metrics based line height was calculated in [the original Chad code](https://github.com/linebender/parley/blob/3356fd4cf0bda07b95dcd697ee8327e5cd95e4e0/src/layout.rs#L550-L553).

CHANGELOG.md
```diff
@@ -26,6 +26,8 @@ This release has an [MSRV] of 1.82.
 #### Parley
 
 - Breaking change: `Layout::min_content_width`, `Layout::max_content_width`, and `Layout::content_widths` have been replaced with `Layout::calculate_content_widths`, which does not internally cache the widths. This means that `Layout` is now `Sync` again, but callers will have to cache the min and max content widths themselves. ([#353][] by [@valadaptive][])
+- Breaking change: the line height style property (`StyleProperty::LineHeight` and the `line_height` field on `TextStyle`) is now a `LineHeight` enum that allows you to specify absolute, font-size-relative, and font-metrics-relative line heights. Previously, it was always font-size-relative. ([#362][] by [@valadaptive][])
+  - The default line height was previously `LineHeight::FontSizeRelative(1.0)` if you used `RangedStyleBuilder`, or `LineHeight::FontSizeRelative(1.2)` if you used `TreeStyleBuilder`. It is now `LineHeight::MetricsRelative(1.0)` in both cases. This will affect layout if you don't specify your own line height.
```
xStrom, line 30: Let's do [one line per sentence](https://linebender.org/wiki/formatting-scheme/#markdown) in markdown files.

```suggestion
- Breaking change: the line height style property (`StyleProperty::LineHeight` and the `line_height` field on `TextStyle`) is now a `LineHeight` enum that allows you to specify absolute, font-size-relative, and font-metrics-relative line heights.
  Previously, it was always font-size-relative. ([#362][] by [@valadaptive][])
  - The default line height was previously `LineHeight::FontSizeRelative(1.0)` if you used `RangedStyleBuilder`, or `LineHeight::FontSizeRelative(1.2)` if you used `TreeStyleBuilder`.
    It is now `LineHeight::MetricsRelative(1.0)` in both cases.
    This will affect layout if you don't specify your own line height.
```

parley/src/layout/mod.rs
```diff
@@ -280,6 +280,23 @@ pub struct Line<'a, B: Brush> {
     data: &'a LineData,
 }
 
+#[derive(Clone, Debug)]
```
xStrom, line 283: Let's get this back.

```suggestion
#[derive(Copy, Clone, Debug)]
```

parley/src/style/mod.rs
xStrom, line 79: Let's get this back too.

```suggestion
    pub(crate) fn resolve(&self, font_size: f32) -> LayoutLineHeight {
        match self {
            Self::MetricsRelative(value) => LayoutLineHeight::MetricsRelative(*value),
            Self::FontSizeRelative(value) => LayoutLineHeight::Absolute(*value * font_size),
            Self::Absolute(value) => LayoutLineHeight::Absolute(*value),
        }
    }
```

xStrom (commented):
It looks like some commits got lost when you force pushed without pulling.

The previous situation of the different builders having different line height defaults is crazy, glad that's no longer the case.

Great to also see less snapshot churn here. We might want to eventually change the test default, but then it can happen independently of any code changes.

examples/swash_render/src/main.rs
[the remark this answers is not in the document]
DJMcNab: I can understand that for `swash_render`. I still think it stands for Vello Editor.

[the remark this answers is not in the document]
valadaptive: Left this here but removed the vello_editor line height setting.

parley/src/style/mod.rs
```diff
@@ -41,6 +43,42 @@ pub enum OverflowWrap {
     BreakWord,
 }
 
+#[derive(Debug, Clone, Copy, PartialEq)]
+pub enum LineHeight {
+    /// Line height specified relative to what's defined in the font's metrics.
+    MetricsRelative(f32),
+    /// Line height specified as a multiple of the font size.
```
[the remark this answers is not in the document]
valadaptive: Updated the docs here.

parley/src/style/mod.rs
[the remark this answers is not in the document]
valadaptive: Expanded on this a bit in the doc comments.

parley/src/tests/test_lines.rs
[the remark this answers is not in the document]
valadaptive: Updated the test code.

parley/src/layout/mod.rs
valadaptive: `leading` is just the font-specified extra gap between lines, beyond the ascender and descender height. It's quite small compared to other metrics, and could vary a lot between fonts.

Say one font has an ascent+descent of 700 and a leading of 300, and another has an ascent+descent of 900 and a leading of 100. Multiplying just the leading would result in `LineHeight::MetricsRelative` being three times as sensitive on the former font than the latter, despite both having the same line height. Theoretically it could also be zero or negative (the OpenType property is a signed value) which would result in more weirdness.

parley/src/style/mod.rs
valadaptive: Not sure how I feel about linking to a mostly-unrelated StackOverflow post about Go, and couldn't find any other diagrams that mentioned the line gap / leading metric. I just explained it in the doc comment.

valadaptive:
Whoops--I need to get into the habit of using `--force-with-lease` instead of `--force`.

I've implemented most of the suggestions, and expanded the doc comments for the `LineHeight` type quite a bit. I've also implemented `Default` for `LineHeight`.

I'd rather update the few tests whose snapshots changed; I can visually confirm they still pass, and it's not a big deal since there's not as much snapshot churn as updating the entire test suite would be.

parley/tests/snapshots/leading_whitespace-preserve.png
xorgy, line 1: I feel like since this changes the default, it's okay to update the test to its prior equivalent rather than just letting it change arbitrarily for this.

parley/src/layout/mod.rs
DJMcNab: This was largely idle musing - I agree it shouldn't be actioned.

xStrom (approved):
Excellent work, thanks!

parley/src/resolve/mod.rs
```diff
@@ -152,7 +152,7 @@ impl ResolveContext {
             }
             StyleProperty::StrikethroughSize(value) => StrikethroughSize(value.map(|x| x * scale)),
             StyleProperty::StrikethroughBrush(value) => StrikethroughBrush(value.clone()),
-            StyleProperty::LineHeight(value) => LineHeight(*value),
+            StyleProperty::LineHeight(value) => LineHeight(value.scale(scale)),
```
xStrom, line 155: This is correct and is in the `resolve_property` method.

However, the next method, `resolve_entire_style_set`, needs the following edit:

```diff
- line_height: raw_style.line_height,
+ line_height: raw_style.line_height.scale(scale),
```

parley/src/resolve/mod.rs
```diff
@@ -427,7 +427,7 @@ impl<B: Brush> Default for ResolvedStyle<B> {
             brush: Default::default(),
             underline: Default::default(),
             strikethrough: Default::default(),
-            line_height: 1.,
+            line_height: Default::default(),
```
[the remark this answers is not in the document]
xStrom: I'll note that there are 90 instances of `Default::default()` in this repo.

Also, if the thinking is to reduce contributor friction then I would not bring these up during review at all. The better action would be to just open a PR that changes them. 👍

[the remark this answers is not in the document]
valadaptive: Every other field here is initialized with `Default::default()`. I'll change them in a follow-up PR rather than introduce an inconsistency here.

valadaptive:
I think I've addressed all the review comments (fingers crossed). Apologies if I've missed any.

parley/src/resolve/mod.rs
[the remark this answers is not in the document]
DJMcNab: Apologies. That is the correct procedure. Thank you for the correction Kaur.

PR #367, opened 2025-05-24
Add `TextWrapMode` style

### LINKED ISSUE #328: Add style to disable soft wrapping (`text-wrap-mode: nowrap`)
Parley currently always wraps text is it overflows the max-advance. We should add a style that disables soft-wrapping for a span of text.

CSS calls this `text-wrap-mode` (https://developer.mozilla.org/en-US/docs/Web/CSS/text-wrap-mode). Parley currently always behaves as if `text-wrap-mode: wrap`. We should implement `text-wrap-mode: nowrap` or equivalent functionality.

nicoburns:
# Objective

Allow non-emergency soft-wrapping to be disabled for a span of text
- Fixes #328

## Changes made

- Add new `TextWrapMode` style
- Update line breaking to not register non-emergency wrapping opportunities within a `NoWrap` span
- Update `calculate_content_widths` to account 

## Notes
Web specifies a `nowrap` keyword (all lowercase) which would be `Nowrap` with Rust's casing conventions. I've gone with `NoWrap` (capital W) as I see little reason to follow the web quirk here.

## Todo

- [ ] Apply `TextWrapMode` to soft-wrap opportunities generated by inline boxes
- [ ] Work out what should happen when a soft-wrap opportunity is on the boundary of a `TextWrapMode` style change

## Screenshots

See the "tags" ("London", "UK", etc) at the bottom of the cards.

`main`:

Screenshot 2025-05-24 at 20 26 05

This PR:

Screenshot 2025-05-24 at 20 27 22


parley/src/style/mod.rs
```diff
@@ -25,6 +25,19 @@ pub enum WhiteSpaceCollapse {
...
+/// Control over non-"emergency" line-breaking.
+///
+/// See <https://drafts.csswg.org/css-text-4/#text-wrap-mode> for more information.
+#[derive(Copy, Clone, Default, PartialEq, Eq, Debug)]
+#[repr(u8)]
+pub enum TextWrapMode {
+    /// Wrap at non-emergency soft-wrap opportunities when necessary to prevent overflow.
+    #[default]
+    Wrap,
+    /// Do not wrap as non-emergy soft-wrap opportunities.
```
xStrom, line 37:
```suggestion
    /// Do not wrap at non-emergy soft-wrap opportunities.
```

parley/src/style/mod.rs
```diff
@@ -25,6 +25,19 @@ pub enum WhiteSpaceCollapse {
...
+///
+/// See <https://drafts.csswg.org/css-text-4/#text-wrap-mode> for more information.
+#[derive(Copy, Clone, Default, PartialEq, Eq, Debug)]
+#[repr(u8)]
+pub enum TextWrapMode {
+    /// Wrap at non-emergency soft-wrap opportunities when necessary to prevent overflow.
+    #[default]
+    Wrap,
+    /// Do not wrap at non-emergency soft-wrap opportunities.
+    NoWrap,
```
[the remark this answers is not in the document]
waywardmonkeys: I think it depends. I think in this case this is like a Boolean where false is don’t wrap, more like zero. But I guess you came the counter argument that if you were to memset, you want the default…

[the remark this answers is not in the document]
nicoburns: I can, but manually setting enum values sounds like an antipattern, even in C. Surely any FFI bindings would have enums with named variants on the foreign side anyway?

parley/src/layout/line_break.rs
```diff
@@ -290,7 +295,11 @@ impl<'a, B: Brush> BreakLines<'a, B> {
                         let boundary = cluster.info().boundary();
                         let style = &self.layout.data.styles[cluster.data.style_index as usize];
```
taj-p, line 296: We needn't lookup the style on a per cluster basis IIUC. We [push runs for a given style](https://github.com/linebender/parley/blob/50f480a15bd7bec539e44b945fd6110fdbdc5cfe/parley/src/shape/mod.rs#L398).

No action required - just thought I'd highlight this for potential discussion.

parley/src/layout/line_break.rs
```diff
@@ -353,7 +363,7 @@ impl<'a, B: Brush> BreakLines<'a, B> {
                         // Else we line break:
                         else {
```
taj-p, line 366: Would it be possible to somewhere add a comment stating that when `TextWrapMode` is `NoWrap`, we expect to fall into the final `else` block because previous and emergency boundaries are never recorded

parley/src/layout/line_break.rs
taj-p, line 1: Could you please add these tests to `test_wrap.rs`?

```rs

#[test]
fn text_wrap_mode_nowrap_disables_soft_wraps() {

    let text = "Most words are short. But Antidisestablishmentarianism is long and needs to wrap.";
    let wrap_width = 120.0;

    let mut baseline_layout = env.ranged_builder(text).build(text);
    baseline_layout.break_all_lines(Some(wrap_width));
    assert!(
        baseline_layout.len() > 1,
        "Expected baseline layout to wrap with width {wrap_width}"
    );

    let mut builder = env.ranged_builder(text);
    builder.push_default(StyleProperty::TextWrapMode(TextWrapMode::NoWrap));
    layout.break_all_lines(Some(wrap_width));

    assert_eq!(
        layout.len(),
        1,
        "Applying TextWrapMode::NoWrap should prevent soft wrapping"
    );

    let line_advance = layout
        .lines()
        .next()
        .expect("layout should have one line")
        .metrics()
        .advance;
    assert!(
        line_advance > wrap_width,
        "Line advance {line_advance} should overflow the requested width {wrap_width}"
    );
}

#[test]
fn text_wrap_mode_allows_break_before_nowrap_span() {

    let text = "Hello world!";
    let prefix = "Hello ";

    let prefix_width = {
        let mut layout = env.ranged_builder(prefix).build(prefix);
        layout.break_all_lines(None);
        layout.width()
    };

    let wrap_width = prefix_width + 1.0;

    let start = text.find("world!").unwrap();
    let mut builder = env.ranged_builder(text);
    builder.push(
        StyleProperty::TextWrapMode(TextWrapMode::NoWrap),
        // `start..text.len()` === "world!"
        start..text.len(),
    );

    layout.break_all_lines(Some(wrap_width));

    assert_eq!(
        layout.len(),
        2,
        "Layout should still wrap before the NoWrap span boundary"
    );

    let first_line = layout.get(0).unwrap();
    let second_line = layout.get(1).unwrap();
    assert_eq!(
        &text[first_line.text_range()],
        "Hello ",
        "First line should end before the NoWrap span"
    );
    assert_eq!(
        &text[second_line.text_range()],
        "world!",
        "Second line should contain the NoWrap span"
    );
}

#[test]
fn text_wrap_mode_updates_min_content_width() {

    let text = "Tags: London UK Paris FR";
    let span_text = "London UK";
    let span_start = text.find(span_text).unwrap();
    let span_range = span_start..span_start + span_text.len();

    let widths_wrap = {
        let layout = env.ranged_builder(text).build(text);
        layout.calculate_content_widths()
    };

    let widths_nowrap = {
        let mut builder = env.ranged_builder(text);
        builder.push(
            StyleProperty::TextWrapMode(TextWrapMode::NoWrap),
            span_range.clone(),
        );
        let layout = builder.build(text);
        layout.calculate_content_widths()
    };

    let span_width = {
        let mut layout = env.ranged_builder(span_text).build(span_text);
        layout.break_all_lines(None);
        layout.width()
    };

    assert!(
        widths_wrap.min < span_width,
        "Without NoWrap, min content width {} should be smaller than the span width {}",
        widths_wrap.min,
        span_width
    );
    assert!(
        widths_nowrap.min >= span_width - 0.5,
        "With NoWrap, min content width {} should be at least the span width {}",
        widths_nowrap.min,
        span_width
    );
}
```

parley/src/layout/line_break.rs
nicoburns: I will add a comment, but note that `text-wrap-mode` exists at the span level, so it is possible that a wrapping boundary was recorded on this line before we entered a `nowrap` section.

parley/src/layout/line_break.rs
```diff
@@ -394,7 +411,14 @@ impl<'a, B: Brush> BreakLines<'a, B> {
...
                                     return self.start_new_line();
                                 }
-                            } else {
+                            }
+                            // Case: no line-breaking opportunities available
+                            //
+                            // This can happen when wrapping is disabled (TextWrapMode::NoWrap) or when no wrapping opportunities
+                            // (according to our `OverflowWrap` and `WordBreak` styles) have yet been encountered.
+                            //
+                            // We fall back to appending the content to the line.
```
taj-p, line 420: These comments are brilliant! Thanks @nicoburns 🙏

parley/src/layout/line_break.rs
nicoburns: I agree. My impression is that may not have been the case historically, because there seem to be quite a few places which are looking up styles on clusters (do clusters even need a style index, or could we access indirectly from the run?). I think we ought to change this seperately from this PR though to avoiding mixing unrelated changes.

taj-p: Nice! I agree. I'm glad to be able to corroborate that idea with your understanding. Also agree that this shouldn't affect this PR.

PR #370, opened 2025-05-27
Add a stub `attributed_text` crate.

waywardmonkeys:
This will let the first PR with code not have this boilerplate in it.

It is marked as not being published to prevent it from being accidentally published.


attributed_text/src/lib.rs
```diff
@@ -0,0 +1,19 @@
+// Copyright 2025 the Parley Authors
+// SPDX-License-Identifier: Apache-2.0 OR MIT
+
+//! Attributed Text is a Rust crate which ...
+//!
+//! ## Features
+//!
+//! - `std` (enabled by default): Use the Rust standard library.
```
DJMcNab, line 8: Features should document why you'd use them, not just "how" they work. Can we just add something like: "this is currently unused and is provided for forward compatibility".

DJMcNab (approved):
Seems reasonable

xorgy (approved):

attributed_text/src/lib.rs
DJMcNab: Well for example [`color`](https://docs.rs/color/latest/color/#features) says why you would enable the feature.

PR #372, opened 2025-05-28
Fontique: Improve config file locating.

dhardy:
Clarify/simplify the XDG config directory fallback logic for config includes and user-level configurations.

### DIFF
--- a/fontique/src/backend/fontconfig/mod.rs
+++ b/fontique/src/backend/fontconfig/mod.rs
@@ -88,7 +88,9 @@ impl SystemFonts {
         let mut raw_families: HashMap<_, _> = Default::default();
         let mut fallback_map: HashMap<Script, FallbackFamilies> = Default::default();
 
-        // First, parse the raw config files
+        // First, parse the raw config files. We attempt to replicate fontconfig
+        // behaviour as specified in the user guide:
+        // https://www.freedesktop.org/software/fontconfig/fontconfig-user.html
         let mut config = Config::default();
         config::parse_config("/etc/fonts/fonts.conf".as_ref(), &mut config);
         if let Some(dir) = std::env::var("XDG_CONFIG_HOME")


xorgy (changes requested):
As far as I understand it, `$XDG_CONFIG_DIRS`  is not meant to substitute entirely for `$XDG_CONFIG_HOME` and may not (does not?) include `$XDG_CONFIG_HOME`.

From your source: (emphasis mine)

> `$XDG_CONFIG_DIRS` defines the preference-ordered set of base directories to search for configuration files _in addition to_ the `$XDG_CONFIG_HOME` base directory.

So the correct implementation of this would be `$XDG_CONFIG_HOME` OR `$HOME/.config`,  THEN `$XDG_CONFIG_DIRS`.

xorgy:
I think it might be worth surveying a little more closely the config file discovery behavior of FontConfig. I think the `/etc/fonts/fonts.conf` and `<include>` handling seems to be fine, but the `fonts-conf` manual lists `~/.fonts.conf` and `~/.fonts.conf.d` (though it's unclear whether, like `/etc/fonts/fonts.conf`, `~/.fonts.conf.d` would need to be manually included by something else.

fontique/src/backend/fontconfig/mod.rs
```diff
@@ -87,25 +87,39 @@ impl SystemFonts {
...
-                PathBuf::from(user_home)
-                    .as_path()
-                    .join(".config/fontconfig/fonts.conf")
-                    .as_path(),
-                &mut config,
-            );
+            config_dir = Some(PathBuf::from(&user_home).join(".config"));
+            let path = PathBuf::from(user_home).join(".config/fontconfig/fonts.conf");
+            config::parse_config(&path, &mut config);
+        }
```
xorgy: Might be worth combining these, since the `$HOME` fallback is meant specifically to substitute for `$XDG_CONFIG_HOME` (so use some sort of `or` instead of two if-lets).

xorgy:
Looks better now, maybe the slight cleanup I suggested, plus a comment that references the documentation, and it'll be ready to go. :+ )

fontique/src/backend/fontconfig/mod.rs
dhardy: Okay... tweak pushed, but it's not objectively better IMO.

dhardy:
Well, the current version of this PR is much simplified. I suggest we leave `$XDG_CONFIG_DIRS` for now.

waywardmonkeys (commented):
I think we need a doc comment somewhere to explicitly state what our intent / understanding is and maybe some of the uses of env vars should have a link to the documentation that you linked to from the PR.

xorgy:
@dhardy by the way, since this is approved, and since you are a contributor, you can go ahead and merge this if you believe it is ready and that you have addressed any requests in the reviews.

dhardy:
@xorgy this is ready for merge. I don't see a merge button, so assume I have not been given this permission.

PR #374, opened 2025-05-31
Prepare for 0.5.0 release of Parley

nicoburns:
- I have checked that the changelog is up to date (and determined that no new entries were necessary). A couple of PRs are not listed, but they're just adding derives for `Debug` or `Hash` impls for types which doesn't seem useful to log in the changelog to me.
- I have run `cargo update --ignore-rust-version` and  `cargo upgrade --ignore-rust-version`
- I have bumped the versions of Parley and Fontique to `0.5.0`

A new release was requested in https://xi.zulipchat.com/#narrow/channel/205635-parley/topic/New.20release.3F/with/521457605

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,10 +8,20 @@ Subheadings to categorize changes are `added, changed, deprecated, removed, fixe
 
 # Changelog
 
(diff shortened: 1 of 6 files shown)


xStrom (approved):
Seems reasonable, let's do it!

Issue #371, opened 2025-05-28
Certain Unicode characters not properly matched with fontconfig backend


ofluffydev:
On Arch Linux with Hyprland, and having installed all of the following fonts:
```
    ttf-noto
    ttf-linux-libertine
    ttf-unifont
    ttf-arimo-nerd
    ttf-unifont-bin
    noto-fonts
    ttf-dejavu
    ttf-liberation
    gnu-unifont
```

The characters used in Xilem's transforms example: `✓⨯↶↑↷←→↓` don't have a suitable font selected, and so show a [tofu](https://fonts.google.com/knowledge/glossary/tofu).

See also [#parley > Fontconfig backend for fontique - why not FFI?](https://xi.zulipchat.com/#narrow/channel/205635-parley/topic/Fontconfig.20backend.20for.20fontique.20-.20why.20not.20FFI.3F/with/520625025)

DJMcNab:
On my machine, I can see the four directional arrows (`↑←→↓`), but not the other symbols. Which is quite bizarre, because they're all in the unicode "arrows" block.

dhardy:
Another observation: font matching currently (if I understand correctly) matches all fonts with the specified family and attributes, followed by all fonts with the matching fallback script. I'm not sure if that's what should happen here.

Fonts should have a [unicode range](https://learn.microsoft.com/en-us/typography/opentype/spec/os2#ur) property. If the matcher were to specify a block, then the matching fonts could be limited to those which contain this block, thus reducing the number of irrelevant matches.

As for the latter problem of fullwidth latin characters not getting any suitable font matches, perhaps a second level of fallback is needed: if no fonts support the requested block, then check for all fonts which do? But that would likely cause too many matches.

I don't see a Unicode standard on font selection, though there is a mention of "OpenType font selection".

nicoburns:
> Fonts should have a [unicode range](https://learn.microsoft.com/en-us/typography/opentype/spec/os2#ur) property. If the matcher were to specify a block, then the matching fonts could be limited to those which contain this block, thus reducing the number of irrelevant matches.

Unicode range matching is also important for Blitz. In Blitz it's a correctness issue as the unicode range can be user-specified (as part of an `@font-face` directive) and may not match the range of the font (it may for example be smaller than the range actually supported by a font because a different font is preferred for some of the unicode range).

dhardy:
I was looking at implementing this, but hit a couple of blockers.

[...]

PR #377, opened 2025-06-07
Add UnicodeRange for font query

dhardy:
This is a partial solution to #371.
My testing shows improvements; e.g. that arrows now use the most appropriate font instead of the first which happens to contain a match (this makes styles much more consistent).

- [x] Add `enum UnicodeRange`. New code because I couldn't find an appropriate impl on `crates.io` (though there are some things touching on this [in ttf_parser](https://docs.rs/ttf-parser/latest/ttf_parser/struct.Face.html#method.unicode_ranges) and [in read-fonts](https://docs.rs/read-fonts/latest/read_fonts/tables/os2/static.OS2_UNICODE_RANGES.html)).
- [ ] Fill in remaining ranges
- [x] Add `struct UnicodeRanges`
- [x] Filter `Query` by matching ranges
- [ ] Extend query when no matches are available

The missing part (last item) is to add a second-stage fallback (all fonts functional over the range) for when there are no matches. This is not quite so straightforward since `Collection` doesn't have a function to list all available families; I'd like some guidance on the best way to do this within fontique. Should `Query::matches_with` automatically do this when it has no other matches, or should it be up to the caller to call something else like `Query::all_fonts_for_range(range: UnicodeRange)` in this case?

Parley changes are needed to use this. `kas-text` is updated here: https://github.com/kas-gui/kas-text/pull/97.

### DIFF
--- a/fontique/src/collection/query.rs
+++ b/fontique/src/collection/query.rs
@@ -3,13 +3,12 @@
 
 //! Query support.
 
-use super::super::{Collection, SourceCache};
-
 use alloc::vec::Vec;
 
-use super::{
-    super::{Attributes, Blob, FallbackKey, FamilyId, FamilyInfo, GenericFamily, Synthesis},
-    Inner,
+use super::Inner;
+use crate::{
+    Attributes, Blob, Collection, FallbackKey, FamilyId, FamilyInfo, GenericFamily, SourceCache,
+    Synthesis, UnicodeRange,
 };
 
 #[derive(Clone, Default)]
@@ -34,6 +33,7 @@ pub struct Query<'a> {
     source_cache: &'a mut SourceCache,
     attributes: Attributes,
     fallbacks: Option<FallbackKey>,
+    range: Option<UnicodeRange>,
 }
 
 impl<'a> Query<'a> {
@@ -45,6 +45,7 @@ impl<'a> Query<'a> {
             source_cache,
             attributes: Attributes::default(),
             fallbacks: None,
+            range: None,
         }
     }
 
@@ -99,6 +100,14 @@ impl<'a> Query<'a> {
         }
     }
 
+    /// Sets the required [`UnicodeRange`]
+    ///
+    /// If set, only fonts claiming to be functional over this range will be
+    /// returned.
+    pub fn set_range(&mut self, ur: impl Into<Option<UnicodeRange>>) {
+        self.range = ur.into()
+    }
+
     /// Invokes the given callback with all fonts that match the current
     /// settings.
     ///
@@ -129,6 +138,7 @@ impl<'a> Query<'a> {
             let mut best_index = None;
             if let Some(font) = load_font(
                 family_info,
+                self.range,
                 self.attributes,
                 &mut family.best,
                 false,
@@ -146,6 +156,7 @@ impl<'a> Query<'a> {
             }
             if let Some(font) = load_font(
                 family_info,
+                self.range,
                 self.attributes,
                 &mut family.default,
                 true,
@@ -223,6 +234,7 @@ pub struct QueryFont {
 
 fn load_font<'a>(
     family: &FamilyInfo,
+    opt_range: Option<UnicodeRange>,
     attributes: Attributes,
     font: &'a mut Entry<QueryFont>,
     is_default: bool,
@@ -241,6 +253,12 @@ fn load_font<'a>(
                 family.match_index(attributes.width, attributes.style, attributes.weight, true)?
             };
             let font_info = family.fonts().get(family_index)?;
+            if let Some(range) = opt_range {
+                if !font_info.ranges().contains(range) {
+                    return None;
+                }
+            }
+
             let blob = font_info.load(Some(source_cache))?;
             let blob_index = font_info.index();
             let synthesis =
--- a/fontique/src/font.rs
+++ b/fontique/src/font.rs
@@ -6,6 +6,7 @@
 use super::attributes::{FontStyle, FontWeight, FontWidth};
 use super::source::{SourceInfo, SourceKind};
 use super::{Blob, source_cache::SourceCache};
+use crate::UnicodeRanges;
 use core::fmt;
 use read_fonts::{FontRef, TableProvider as _, types::Tag};
 use smallvec::SmallVec;
@@ -15,6 +16,7 @@ type AxisVec = SmallVec<[AxisInfo; 1]>;
 /// Representation of a single font in a family.
 #[derive(Clone, Debug)]
 pub struct FontInfo {
+    ranges: UnicodeRanges,
     source: SourceInfo,
     index: u32,
     width: FontWidth,
@@ -40,6 +42,13 @@ impl FontInfo {
                 Self::from_font_ref(&font, source.clone(), index)
             }
         }
+    }
+
+    /// Returns the bit-map of contained ranges
+    ///
+    /// This represents Unicode characters over which the font is "functional".
+    pub fn ranges(&self) -> UnicodeRanges {
+        self.ranges
     }
 
     /// Returns an object describing how to locate the data containing this
@@ -199,7 +208,7 @@ impl FontInfo {
         source: SourceInfo,
         index: u32,
     ) -> Option<Self> {
-        let (width, style, weight) = read_attributes(font);
+        let (ranges, width, style, weight) = read_attributes(font);
         let (axes, attr_axes) = if let Ok(fvar_axes) = font.fvar().and_then(|fvar| fvar.axes()) {
             let mut axes = SmallVec::<[AxisInfo; 1]>::with_capacity(fvar_axes.len());
             let mut attrs_axes = 0_u8;
@@ -225,6 +234,7 @@ impl FontInfo {
             (Default::default(), Default::default())
         };
         Some(Self {
+            ranges,
             source,
             index,
             width,
@@ -368,7 +378,7 @@ impl fmt::Debug for Synthesis {
     }
 }
 
-fn read_attributes(font: &FontRef<'_>) -> (FontWidth, FontStyle, FontWeight) {
+fn read_attributes(font: &FontRef<'_>) -> (UnicodeRanges, FontWidth, FontStyle, FontWeight) {
     use read_fonts::{
         TableProvider,
         tables::{
@@ -392,7 +402,17 @@ fn read_attributes(font: &FontRef<'_>) -> (FontWidth, FontStyle, FontWeight) {
         })
     }
 
-    fn from_os2_post(os2: Os2<'_>, post: Option<Post<'_>>) -> (FontWidth, FontStyle, FontWeight) {
+    fn from_os2_post(
+        os2: Os2<'_>,
+        post: Option<Post<'_>>,
+    ) -> (UnicodeRanges, FontWidth, FontStyle, FontWeight) {
+        let ranges = UnicodeRanges::from([
+            os2.ul_unicode_range_1(),
+            os2.ul_unicode_range_2(),
+            os2.ul_unicode_range_3(),
+            os2.ul_unicode_range_4(),
+        ]);
+
         let width = width_from_width_class(os2.us_width_class());
         // Bits 1 and 9 of the fsSelection field signify italic and
         // oblique, respectively.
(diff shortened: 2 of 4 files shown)


dhardy:
The test input from #371; also some Hebrew characters:
image

<details>
<summary>Font matches</summary>
<pre><code>
[2025-06-07T14:11:48Z DEBUG kas_text::fonts::resolver] select: Script::Latn, Some(BasicLatin), GenericFamily::SystemUi, FontWeight(400), FontWidth(256), Normal
[2025-06-07T14:11:48Z DEBUG kas_text::fonts::library] match: Cantarell
[2025-06-07T14:11:48Z DEBUG kas_text::fonts::library] match: DejaVu Sans
[2025-06-07T14:11:48Z DEBUG kas_text::fonts::resolver] select: Script::Zyyy, Some(BasicLatin), GenericFamily::SystemUi, FontWeight(400), FontWidth(256), Normal
[2025-06-07T14:11:48Z DEBUG kas_text::fonts::library] match: Cantarell
[2025-06-07T14:11:48Z DEBUG kas_text::fonts::resolver] select: Script::Hebr, Some(Hebrew), GenericFamily::SystemUi, FontWeight(400), FontWidth(256), Normal
[2025-06-07T14:11:48Z DEBUG kas_text::fonts::library] match: Noto Sans Hebrew
[2025-06-07T14:11:48Z DEBUG kas_text::fonts::resolver] select: Script::Latn, Some(BasicLatin), GenericFamily::Serif, FontWeight(400), FontWidth(256), Normal
[2025-06-07T14:11:48Z DEBUG kas_text::fonts::library] match: Noto Serif
[2025-06-07T14:11:48Z DEBUG kas_text::fonts::library] match: Nimbus Roman
[2025-06-07T14:11:48Z DEBUG kas_text::fonts::library] match: URW Bookman
[2025-06-07T14:11:48Z DEBUG kas_text::fonts::library] match: C059
[2025-06-07T14:11:48Z DEBUG kas_text::fonts::library] match: P052
[2025-06-07T14:11:48Z DEBUG kas_text::fonts::library] match: Standard Symbols PS
[2025-06-07T14:11:48Z DEBUG kas_text::fonts::library] match: Caladea
[2025-06-07T14:11:48Z DEBUG kas_text::fonts::library] match: Symbola
[2025-06-07T14:11:48Z DEBUG kas_text::fonts::library] match: STIX
[2025-06-07T14:11:48Z DEBUG kas_text::fonts::library] match: Lohit Bengali
[2025-06-07T14:11:48Z DEBUG kas_text::fonts::library] match: Lohit Gujarati
[2025-06-07T14:11:48Z DEBUG kas_text::fonts::library] match: Lohit Marathi
[2025-06-07T14:11:48Z DEBUG kas_text::fonts::library] match: Lohit Tamil
[2025-06-07T14:11:48Z DEBUG kas_text::fonts::library] match: Lohit Kannada
[2025-06-07T14:11:48Z DEBUG kas_text::fonts::library] match: Lohit Telugu
[2025-06-07T14:11:48Z DEBUG kas_text::fonts::library] match: DejaVu Sans
[2025-06-07T14:11:48Z DEBUG kas_text::fonts::resolver] select: Script::Zyyy, Some(BasicLatin), GenericFamily::Serif, FontWeight(400), FontWidth(256), Normal
[2025-06-07T14:11:48Z DEBUG kas_text::fonts::library] match: Noto Serif
[2025-06-07T14:11:48Z DEBUG kas_text::fonts::library] match: Nimbus Roman
[2025-06-07T14:11:48Z DEBUG kas_text::fonts::library] match: URW Bookman
[2025-06-07T14:11:48Z DEBUG kas_text::fonts::library] match: C059
[2025-06-07T14:11:48Z DEBUG kas_text::fonts::library] match: P052
[2025-06-07T14:11:48Z DEBUG kas_text::fonts::library] match: Standard Symbols PS
[2025-06-07T14:11:48Z DEBUG kas_text::fonts::library] match: Caladea
[2025-06-07T14:11:48Z DEBUG kas_text::fonts::library] match: Symbola
[2025-06-07T14:11:48Z DEBUG kas_text::fonts::library] match: STIX
[2025-06-07T14:11:48Z DEBUG kas_text::fonts::library] match: Lohit Bengali
[2025-06-07T14:11:48Z DEBUG kas_text::fonts::library] match: Lohit Gujarati
[2025-06-07T14:11:48Z DEBUG kas_text::fonts::library] match: Lohit Marathi
[2025-06-07T14:11:48Z DEBUG kas_text::fonts::library] match: Lohit Tamil
[2025-06-07T14:11:48Z DEBUG kas_text::fonts::library] match: Lohit Kannada
[2025-06-07T14:11:48Z DEBUG kas_text::fonts::library] match: Lohit Telugu
[2025-06-07T14:11:56Z DEBUG kas_text::fonts::resolver] select: Script::Zyyy, Some(Dingbats), GenericFamily::Serif, FontWeight(400), FontWidth(256), Normal
[2025-06-07T14:11:56Z DEBUG kas_text::fonts::library] match: Symbola
[2025-06-07T14:11:56Z DEBUG kas_text::fonts::library] match: STIX
[2025-06-07T14:11:56Z DEBUG kas_text::fonts::resolver] select: Script::Zyyy, None, GenericFamily::Serif, FontWeight(400), FontWidth(256), Normal
[2025-06-07T14:11:56Z DEBUG kas_text::fonts::library] match: Noto Serif
[2025-06-07T14:11:56Z DEBUG kas_text::fonts::library] match: Nimbus Roman
[2025-06-07T14:11:56Z DEBUG kas_text::fonts::library] match: URW Bookman
[2025-06-07T14:11:56Z DEBUG kas_text::fonts::library] match: C059
[2025-06-07T14:11:56Z DEBUG kas_text::fonts::library] match: P052
[2025-06-07T14:11:56Z DEBUG kas_text::fonts::library] match: Standard Symbols PS
[2025-06-07T14:11:56Z DEBUG kas_text::fonts::library] match: Caladea
[2025-06-07T14:11:56Z DEBUG kas_text::fonts::library] match: Symbola
[2025-06-07T14:11:56Z DEBUG kas_text::fonts::library] match: STIX
[2025-06-07T14:11:56Z DEBUG kas_text::fonts::library] match: Lohit Bengali
[2025-06-07T14:11:56Z DEBUG kas_text::fonts::library] match: Lohit Gujarati
[2025-06-07T14:11:56Z DEBUG kas_text::fonts::library] match: Lohit Marathi
[2025-06-07T14:11:56Z DEBUG kas_text::fonts::library] match: Lohit Tamil
[2025-06-07T14:11:56Z DEBUG kas_text::fonts::library] match: Lohit Kannada
[2025-06-07T14:11:56Z DEBUG kas_text::fonts::library] match: Lohit Telugu
[2025-06-07T14:11:56Z DEBUG kas_text::fonts::resolver] select: Script::Zyyy, Some(Arrows), GenericFamily::Serif, FontWeight(400), FontWidth(256), Normal
[2025-06-07T14:11:56Z DEBUG kas_text::fonts::library] match: Symbola
[2025-06-07T14:11:56Z DEBUG kas_text::fonts::library] match: STIX
</code></pre>
</details>

nicoburns:
Hmm... I believe that for CSS matching we need support for arbitrary numeric ranges (not just "standard ranges") specified as part of the [FontInfoOverride](https://docs.rs/fontique/latest/fontique/struct.FontInfoOverride.html). See https://developer.mozilla.org/en-US/docs/Web/CSS/@font-face/unicode-range. So perhaps it would make sense to have some variant on `Vec<Range<u32>>` (`SmallVec<[Range<u32>; 4]>`?) as the core type for unicode ranges within Fontique?

Reading this PR it looks like the 128 standard ranges thing is part of an OpenType standard? So I guess we may still need that code (not my area of expertise), although presumably one can also check which codepoints a font *actually* supports?

dhardy:
> although presumably one can also check which codepoints a font actually supports?

Yes, of course. That's largely orthogonal to this PR. I had assumed that Parley would already do this, but maybe not; this may be why `kas-text` already did better than Parley in #371.

The way this works in `kas-text` is that, per glyph, I create a hash of the font selector (family & attributes), the script and (after this PR) the `unicode_range`); this is used to select a matching font *list* and cache it via the hash. Then, per glyph, I take the first matching font face from that list which contains the `char` (with the quirks that it prefers to use the last `char`'s face if possible).

[`kas-text` docs](https://docs.rs/kas-text/latest/kas_text/fonts/index.html) describe both `FontId` (the list) and `FaceId` (one font face), though that's a little dated (there's no longer a "default font").

This is per-`char` fallback, and the PR here mostly makes it more efficient (those font lists are smaller, omitting fonts that likely wouldn't cover the current `char`).

Using this PR *without* per-`char` fallback would probably mostly work, but likely not everywhere.

The second (missing) part of this PR would help by providing some likely font matches when no suitable fonts are found otherwise. E.g. if someone uses an arrow then the family won't identify a suitable font (though some contain a few common arrow glyphs anyway) and the script ("common") won't either, but the `UnicodeRange` will.

---

> I believe that for CSS matching we need support for arbitrary numeric ranges

Hmm, `UnicodeRange` is intended to mirror [ulUnicodeRange](https://learn.microsoft.com/en-us/typography/opentype/spec/os2#ur) from the OS/2 table, which most fonts provide. I guess the CSS `unicode-range` is designed to do a similar thing, but (a) is more flexible and (b) relies on a web page's CSS to specify available fonts (with the ranges they cover).

For fonts found on the system, I think we normally only have the info from the OS/2 table.

The *other* reason I chose to precisely mirror the OS/2 specification is because of how `kas-text` caches a set of fonts as a `FontId`; I want some finite small-ish number of categories here (much fewer than the number of possible `char` codes and independent of fonts for simplicity).

I guess there are other possible approaches; e.g. building a (very large) map from `(FontFamily, FontWeight, FontWidth, FontStyle, char)` to a font face directly, or caching maps from `char` to a font face for each `(FontFamily, FontWeight, FontWidth, FontStyle)` or just performing the whole look-up chain for each `char` (probably too slow for reactive UI).

I'm not sure if you want to copy the approach I've adopted for `kas-text` in Parley since it's not exactly compatible with CSS's `unicode-range` — but I don't see a good alternative that allows fast cached look-ups (and is not horribly complex / needing a huge hash map).

*I* would like this to be adopted by `fontique` *anyway*, since it's not incompatible with using another approach for Parley.

dhardy:
Combined with #378 (which effectively just massively increases the number of font matches), this PR is effectively a fast pre-filter for callbacks. E.g. with this PR, matching for a font supporting Hebrew (`Script` and `UnicodeRange` properties) yields 6 matches; without this PR (only `Script` property), it yields 61 (of which only 17 and 50 are obviously Hebrew fonts).

That is because the `Script` property is only used for fallbacks, not to prune other matches. I'm not sure whether that should change (not if we also have this PR, *possibly* otherwise though I don't know if it would be problematic), but even if so it still wouldn't work for arrows (which have a `UnicodeRange` but not a specific `Script`).

nicoburns:
> this PR is effectively a fast pre-filter for callbacks

I am understanding correctly that the invariant this PR is relying on / taking advantage of is "If a font doesn't list a 'standard unicode range' then it doesn't contain any glyphs for any codepoints within that range"? If a font *does* support a given "standard range" then we then still need to check whether the font actually contain the specific glyph we are looking for but this allows for fast-rejecting when a font doesn't cover a given "standard range"?

dhardy:
Honestly, the only specification I found for behaviour is [this one](https://learn.microsoft.com/en-us/typography/opentype/spec/os2#ulunicoderange):

> This field is used to specify the Unicode blocks or ranges encompassed by the font file in 'cmap' subtables for platform 3, encoding ID 1 (Microsoft platform, Unicode BMP) and platform 3, encoding ID 10 (Microsoft platform, Unicode full repertoire). If a bit is set (1), then the Unicode ranges assigned to that bit are considered functional. If the bit is clear (0), then the range is not considered functional. Each of the bits is treated as an independent flag and the bits can be set in any combination. The determination of “functional” is left up to the font designer, although character set selection should attempt to be functional by ranges if at all possible. 

I fully expect that there are some `char`s contained by some fonts which do not indicate that they are "functional" over the corresponding input range, and thus are rejected by the new filter in this PR.

But does that matter? Lets consider `→` ("Arrow" range), using a Sans-Serif font family:

- Without this PR, `→` would be picked from the first matching font, the Sans-Serif font. This may be more consistent with other text in that font.
- With this PR and a suitable "Arrow" font, `→` would be picked from the arrow font. This is (according to my testing) more consistent with other, less common arrows like `↷`, but maybe less consistent with ordinary Sans-Serif text.
- With this PR without a suitable "Arrow" font, `→` would not be matched. That is a problem if a system doesn't have such a font.

The latter point can be addressed by changing how glyph fallback works: instead of checking one long list of fonts, use at least two: a pre-filtered list of preferred fonts (possibly also using `Script` for filtering) and a longer (unfiltered) list. This would require further changes.

Related questions:

- Which font pick is preferable? (See `→` example above — IMO the "Arrow"-specific font is preferable in the very specific case I tested.)
- How do you ensure that a space between two Arabic glyphs does not revert to the default Sans-Serif font? (This is important for shaping runs.) The (first) answer I came up with is the afore-mentioned `kas-text` quirk, but this feels like a cheap hack. A better answer may be to ensure the `Script` and `locale` affect the font selection.

---

Conclusion: we likely still want `UnicodeRange`, but the query API needs some other changes (new issue).

Issue #323, opened 2025-03-29
On Linux, `GenericFamily::SystemUi` should fall back to `sans-serif`


valadaptive:
On my machine, there are a lot of languages/Unicode ranges that the `GenericFamily::SystemUi` family doesn't display because it doesn't have any fonts to fall back to. It wants to fall back to various "Noto Sans [script] UI" fonts, which I don't have installed:

![Image](https://github.com/user-attachments/assets/8f5c13a0-692b-4e35-bbbd-605cf3caf681)

Changing this to `GenericFamily::SansSerif` works, but this changes the preferred family as well:

![Image](https://github.com/user-attachments/assets/8bdb5052-156f-4559-b5cc-0d9426beecc4)

Fontconfig [comes with a stock config file to fall back to `sans-serif`](https://gitlab.freedesktop.org/fontconfig/fontconfig/-/blob/564a8a9e8d08694cfb3dfd43e2a9f58156283afe/conf.d/49-sansserif.conf) for all families that are not `sans-serif`, `serif`, or `monospace`. This includes the other generic families like `system-ui`. This lets other applications (e.g. Inkscape) fall back properly:

![Image](https://github.com/user-attachments/assets/1519277a-5b5c-4a29-b1a2-aad826ac7789)

In lieu of reimplementing all of fontconfig's matching code, we could probably just hardcode this fallback behavior here. I'm not sure how it interacts with language mapping or the "style classes", or about the right level to implement it at.

xorgy:
We probably should (at least eventually) implement this part of fontconfig fallbacks, it is used for family substitution in a lot of applications (for example, for metric equivalents and clones).

Issue #355, opened 2025-05-14
fontique: KDE fonts


dhardy:
Fontique does not correctly detect UI fonts on KDE 5:
```
GenericFamily::SystemUi:
Cantarell
Cantarell
Cantarell

GenericFamily::UiSerif:

GenericFamily::UiSansSerif:

GenericFamily::UiMonospace:

GenericFamily::UiRounded:

```

`XDG_CONFIG_HOME` is not set, while `XDG_CONFIG_DIRS` is, but does not point to any font configuration, so my best guess is that we must test `XDG_CURRENT_DESKTOP=KDE(:.*)?`, then parse `$HOME/.config/kdeglobals`.

Excerpt from the latter:
```
[General]
ColorSchemeHash=0be804dba87e3512aeb4be3d78ed981f59f0f2f4
XftHintStyle=hintslight
XftSubPixel=none
fixed=Noto Sans Mono,11,-1,5,50,0,0,0,0,0
font=Noto Sans,12,-1,5,50,0,0,0,0,0
menuFont=Noto Sans,12,-1,5,50,0,0,0,0,0
smallestReadableFont=Noto Sans,10,-1,5,50,0,0,0,0,0
toolBarFont=Noto Sans,12,-1,5,50,0,0,0,0,0
```
This is:
- The family name
- Size ("points", so presumably we just multiply by 96/72 = 4/3 to get dots-per-Em)
- ... requires investigation

The hinting (none/slight/medium/full) & sub-pixel rendering (`none`, `rgb`, `bgr`, `vrgb`, `vbgr`) info is likely best ignored since (a) it is beyond the scope of Fontique and (b) is rasterer dependent.

xorgy:
Thanks for having a look at this, there's a significant rework required to make Fontique and Parley handle fallback and generic families well on various platforms. FontConfig is pretty sprawling, and the information in the config files is simultaneously very detailed and missing important details.

dhardy:
I get the impression that fontique (and anything attempting to do this job in a cross-platform manner) is going to end up as a pile of hacks. I think most of the time there isn't even good documentation, besides source code (where available)?

Let me know if you want further testing on this platform.

xorgy:
I am personally interested in DE/system font settings, but I'm not 100% sure where they belong in the Fontique/Parley ecosystem. If you have some ideas or example programs for resolving these I'd love to see it.
[...]

PR #378, opened 2025-06-08
Reimplement fontconfig backend with FFI

valadaptive:
- Resolves https://github.com/linebender/parley/issues/323.
- Might solve https://github.com/linebender/parley/issues/355 as well; @dhardy, let me know if this fixes your issue.
- Might solve https://github.com/linebender/parley/issues/371; @ofluffydev, let me know if this fixes your issue.

This was discussed [on Zulip](https://xi.zulipchat.com/#narrow/channel/205635-parley/topic/Fontconfig.20backend.20for.20fontique.20-.20why.20not.20FFI.3F/with/519362683). Essentially, the current fontconfig backend only roughly approximates fontconfig's actual font matching behavior, which is very intricate and difficult to implement. This has resulted in many fonts failing to match.

This PR reimplements it entirely using FFI calls to the fontconfig library, under the assumption that any platform with fontconfig configuration files probably also has the fontconfig library installed. For bindings, I'm using `yeslogic-fontconfig-sys` with my own binding layer on top of it. Yeslogic also provides a higher-level `fontconfig` crate but it's very barebones and not useful here.

I've enabled the `dlopen` feature of `yeslogic-fontconfig-sys`; this should mean that Parley-using applications should work even if fontconfig is not installed. We could probably enable the fontconfig backend on other Unixes (various BSDs, illumos, etc). The bindings are not generated at build time, so I don't think there are any issues. I am, however, unsure if `dlopen`ing libraries runs into any sandboxing concerns with Flatpak or something.

I've implemented family and fallback matching to the best of my ability (I've done some manual testing, and `("Hani", None)` returns a Japanese font whereas `("Hani", Some("zh"))` returns a Chinese font). However, I'm still not sure I fully understand fallback matching. If, for instance, you pass a nonexistent script or locale, it'll just return whatever font ends up as the global super-duper-fallback (Bitstream Vera on my machine).

In fact, I believe Fontconfig matching will *always* return at least some font. For family matches, I've filtered out all families whose names don't match, but for fallback matches, I believe it will always return *something*. I'm not sure if there's logic somewhere else that expects nonexistent scripts/locales to return nothing and breaks if we return a bogus font instead.

All that being said, I think this is a good improvement over the current implementation.


dhardy:
I'm less keen on requiring non-Rust deps, but using the standard `fontconfig` impl *does* have merit. Testing:

<details>
<summary>examples/generic_families.rs</summary>
<pre><code>
GenericFamily::Serif:
Noto Serif
Noto Naskh Arabic
Nimbus Roman
Standard Symbols PS
Caladea
Symbola
STIX
Lohit Bengali
Lohit Gujarati
Lohit Marathi
Lohit Tamil
Lohit Kannada
Lohit Telugu
Adwaita Mono
Adwaita Sans
DejaVu Sans
Carlito
Noto Sans Mono
Noto Sans CJK JP
Noto Serif CJK JP
Jomolhari
Khmer OS System
Liberation Mono
Liberation Serif
Lohit Devanagari
RIT Rachana
Mingzat
Nuosu SIL
Waree
D050000L
Droid Sans Ethiopic
Droid Sans Fallback
Noto Sans Gothic
Noto Sans Math
Noto Sans NKo
Noto Sans Symbols 2
Noto Emoji
Noto Sans Bengali
Noto Sans Canadian Aboriginal
Noto Sans Cherokee
Noto Sans Devanagari
Noto Sans Ethiopic
Noto Sans Georgian
Noto Sans Gujarati
Noto Sans Gurmukhi
Noto Sans Hebrew
Noto Sans Kannada
Noto Sans Lao
Noto Sans Meetei Mayek
Noto Sans Ol Chiki
Noto Sans Oriya
Noto Sans Sinhala
Noto Sans Syriac
Noto Sans Tamil
Noto Sans Telugu
Noto Sans Thaana
OpenSymbol
Padauk
Noto Color Emoji

GenericFamily::SansSerif:
Noto Sans
Noto Sans Arabic
DejaVu Sans
Noto Sans Meetei Mayek
Noto Sans Ol Chiki
Adwaita Sans
Nimbus Sans
Carlito
Noto Sans Math
Mingzat
Padauk
Nuosu SIL
Droid Sans Devanagari
Droid Sans Ethiopic
Droid Sans Fallback
Droid Sans Hebrew
Droid Sans Tamil
Droid Sans Thai
Waree
Lohit Bengali
Lohit Gujarati
Lohit Marathi
Lohit Kannada
Lohit Telugu
Source Code Pro
Adwaita Mono
DejaVu Sans Mono
Symbola
Noto Sans CJK JP
Noto Serif CJK JP
Jomolhari
Khmer OS System
Liberation Serif
Lohit Devanagari
RIT Rachana
STIX
D050000L
Noto Sans Gothic
Noto Sans NKo
Noto Sans Symbols 2
Noto Emoji
Noto Sans Bengali
Noto Sans Canadian Aboriginal
Noto Sans Cherokee
Noto Sans Devanagari
Noto Sans Ethiopic
Noto Sans Georgian
Noto Sans Gujarati
Noto Sans Gurmukhi
Noto Sans Hebrew
Noto Sans Kannada
Noto Sans Lao
Noto Sans Oriya
Noto Sans Sinhala
Noto Sans Syriac
Noto Sans Tamil
Noto Sans Telugu
Noto Sans Thaana
OpenSymbol
Noto Color Emoji

GenericFamily::Monospace:
Noto Sans Mono
DejaVu Sans Mono
Adwaita Mono
Nimbus Mono PS
Source Code Pro
Lohit Bengali
Lohit Gujarati
Lohit Marathi
Lohit Tamil
Lohit Kannada
Lohit Telugu
Adwaita Sans
DejaVu Sans
Symbola
Carlito
Noto Sans CJK JP
Noto Serif CJK JP
Jomolhari
Khmer OS System
Liberation Mono
Liberation Serif
Lohit Devanagari
PakType Naskh Basic
RIT Rachana
Mingzat
Nuosu SIL
STIX
Waree
D050000L
Vazirmatn
Droid Arabic Kufi
Droid Sans Ethiopic
Droid Sans Fallback
Noto Sans Gothic
Noto Sans Math
Noto Sans NKo
Noto Sans Symbols 2
Noto Emoji
Noto Naskh Arabic
Noto Sans Bengali
Noto Sans Canadian Aboriginal
Noto Sans Cherokee
Noto Sans Devanagari
Noto Sans Ethiopic
Noto Sans Georgian
Noto Sans Gujarati
Noto Sans Gurmukhi
Noto Sans Hebrew
Noto Sans Kannada
Noto Sans Lao
Noto Sans Meetei Mayek
Noto Sans Ol Chiki
Noto Sans Oriya
Noto Sans Sinhala
Noto Sans Syriac
Noto Sans Tamil
Noto Sans Telugu
Noto Sans Thaana
OpenSymbol
Padauk
Noto Color Emoji

GenericFamily::Cursive:
Z003
Noto Sans Arabic
Noto Sans
DejaVu Sans
Noto Sans Meetei Mayek
Noto Sans Ol Chiki
Adwaita Sans
Carlito
Noto Sans Math
Mingzat
Padauk
Nuosu SIL
Droid Sans Devanagari
Droid Sans Ethiopic
Droid Sans Fallback
Droid Sans Hebrew
Droid Sans Tamil
Droid Sans Thai
Waree
Lohit Bengali
Lohit Gujarati
Lohit Marathi
Lohit Kannada
Lohit Telugu
Source Code Pro
Adwaita Mono
DejaVu Sans Mono
Symbola
Noto Sans CJK JP
Noto Serif CJK JP
Jomolhari
Khmer OS System
Liberation Serif
Lohit Devanagari
RIT Rachana
STIX
D050000L
Noto Sans Gothic
Noto Sans NKo
Noto Sans Symbols 2
Noto Emoji
Noto Sans Bengali
Noto Sans Canadian Aboriginal
Noto Sans Cherokee
Noto Sans Devanagari
Noto Sans Ethiopic
Noto Sans Georgian
Noto Sans Gujarati
Noto Sans Gurmukhi
Noto Sans Hebrew
Noto Sans Kannada
Noto Sans Lao
Noto Sans Oriya
Noto Sans Sinhala
Noto Sans Syriac
Noto Sans Tamil
Noto Sans Telugu
Noto Sans Thaana
OpenSymbol
Noto Color Emoji

GenericFamily::Fantasy:
Noto Sans
Noto Sans Symbols 2
Noto Sans Symbols
Noto Sans Arabic
DejaVu Sans
Noto Sans Meetei Mayek
Noto Sans Ol Chiki
Adwaita Sans
Nimbus Sans
Carlito
Noto Sans Math
Mingzat
Padauk
Nuosu SIL
Droid Sans Devanagari
Droid Sans Ethiopic
Droid Sans Fallback
Droid Sans Hebrew
Droid Sans Tamil
Droid Sans Thai
Waree
Lohit Bengali
Lohit Gujarati
Lohit Marathi
Lohit Kannada
Lohit Telugu
Source Code Pro
Adwaita Mono
Symbola
Noto Sans CJK JP
Noto Serif CJK JP
Jomolhari
Khmer OS System
Liberation Serif
Lohit Devanagari
RIT Rachana
STIX
D050000L
Noto Sans Gothic
Noto Sans NKo
Noto Emoji
Noto Sans Bengali
Noto Sans Canadian Aboriginal
Noto Sans Cherokee
Noto Sans Devanagari
Noto Sans Ethiopic
Noto Sans Georgian
Noto Sans Gujarati
Noto Sans Gurmukhi
Noto Sans Hebrew
Noto Sans Kannada
Noto Sans Lao
Noto Sans Oriya
Noto Sans Sinhala
Noto Sans Syriac
Noto Sans Tamil
Noto Sans Telugu
Noto Sans Thaana
OpenSymbol
Noto Color Emoji

GenericFamily::SystemUi:
Cantarell
Noto Sans Arabic
Noto Sans
DejaVu Sans
Noto Sans Meetei Mayek
Noto Sans Ol Chiki
Adwaita Sans
Nimbus Sans
Carlito
Noto Sans Math
Mingzat
Padauk
Nuosu SIL
Droid Sans Devanagari
Droid Sans Ethiopic
Droid Sans Fallback
Droid Sans Hebrew
Droid Sans Tamil
Droid Sans Thai
Waree
Lohit Bengali
Lohit Gujarati
Lohit Marathi
Lohit Kannada
Lohit Telugu
Source Code Pro
Adwaita Mono
DejaVu Sans Mono
Symbola
Noto Sans CJK JP
Noto Serif CJK JP
Jomolhari
Khmer OS System
Liberation Serif
Lohit Devanagari
RIT Rachana
STIX
D050000L
Noto Sans Gothic
Noto Sans NKo
Noto Sans Symbols 2
Noto Emoji
Noto Sans Bengali
Noto Sans Canadian Aboriginal
Noto Sans Cherokee
Noto Sans Devanagari
Noto Sans Ethiopic
Noto Sans Georgian
Noto Sans Gujarati
Noto Sans Gurmukhi
Noto Sans Hebrew
Noto Sans Kannada
Noto Sans Lao
Noto Sans Oriya
Noto Sans Sinhala
Noto Sans Syriac
Noto Sans Tamil
Noto Sans Telugu
Noto Sans Thaana
OpenSymbol
Noto Color Emoji

GenericFamily::UiSerif:

GenericFamily::UiSansSerif:

GenericFamily::UiMonospace:

GenericFamily::UiRounded:

GenericFamily::Emoji:
Noto Color Emoji
Noto Emoji
Symbola
Noto Sans Arabic
Noto Sans
DejaVu Sans
Noto Sans Meetei Mayek
Noto Sans Ol Chiki
Adwaita Sans
Nimbus Sans
Carlito
Noto Sans Math
Mingzat
Padauk
Nuosu SIL
Droid Sans Devanagari
Droid Sans Ethiopic
Droid Sans Fallback
Droid Sans Hebrew
Droid Sans Tamil
Droid Sans Thai
Waree
Lohit Bengali
Lohit Gujarati
Lohit Marathi
Lohit Kannada
Lohit Telugu
Adwaita Mono
Noto Sans CJK JP
Noto Serif CJK JP
Jomolhari
Khmer OS System
Liberation Serif
Lohit Devanagari
RIT Rachana
STIX
D050000L
Noto Sans Gothic
Noto Sans NKo
Noto Sans Symbols 2
Noto Sans Bengali
Noto Sans Canadian Aboriginal
Noto Sans Cherokee
Noto Sans Devanagari
Noto Sans Ethiopic
Noto Sans Georgian
Noto Sans Gujarati
Noto Sans Gurmukhi
Noto Sans Hebrew
Noto Sans Kannada
Noto Sans Lao
Noto Sans Oriya
Noto Sans Sinhala
Noto Sans Syriac
Noto Sans Tamil
Noto Sans Telugu
Noto Sans Thaana
OpenSymbol

GenericFamily::Math:
STIX
Symbola
Noto Sans Arabic
Noto Sans
DejaVu Sans
Noto Sans Meetei Mayek
Noto Sans Ol Chiki
Adwaita Sans
Nimbus Sans
Carlito
Noto Sans Math
Noto Sans Thai
Mingzat
Padauk
Nuosu SIL
Waree
Droid Sans Devanagari
Droid Sans Ethiopic
Droid Sans Fallback
Droid Sans Hebrew
Droid Sans Tamil
Lohit Bengali
Lohit Gujarati
Lohit Marathi
Lohit Kannada
Lohit Telugu
Adwaita Mono
Noto Sans CJK JP
Noto Serif CJK JP
Jomolhari
Khmer OS System
Liberation Serif
Lohit Devanagari
RIT Rachana
D050000L
Noto Sans Gothic
Noto Sans NKo
Noto Sans Symbols 2
Noto Emoji
Noto Sans Bengali
Noto Sans Canadian Aboriginal
Noto Sans Cherokee
Noto Sans Devanagari
Noto Sans Ethiopic
Noto Sans Georgian
Noto Sans Gujarati
Noto Sans Gurmukhi
Noto Sans Hebrew
Noto Sans Kannada
Noto Sans Lao
Noto Sans Oriya
Noto Sans Sinhala
Noto Sans Syriac
Noto Sans Tamil
Noto Sans Telugu
Noto Sans Thaana
OpenSymbol
Noto Color Emoji

GenericFamily::FangSong:
</code></pre>
</details>

That's a lot more matches than before, which means more possible fallbacks. I'm not convinced that's a good thing though; it depends on whether there are performance implications for fallback. I suspect (provided that there is some level of filtering) most of these additional fonts won't normally do anything (neither matching glyphs nor causing overhead) anyway, but the latter is contigent on having some other level of filtering (like #377).

Testing shows that #377 merges cleanly with this, yielding more possible matches but in a manner that's okay. E.g. here's the matches for arrows (shorter than the matches for Latin):
```
[2025-06-09T08:10:55Z DEBUG kas_text::fonts::resolver] select: Script::Zyyy, Some(Arrows), GenericFamily::Serif, FontWeight(400), FontWidth(256), Normal
[2025-06-09T08:10:55Z DEBUG kas_text::fonts::library] match: Symbola
[2025-06-09T08:10:55Z DEBUG kas_text::fonts::library] match: STIX
[2025-06-09T08:10:55Z DEBUG kas_text::fonts::library] match: Adwaita Mono
[2025-06-09T08:10:55Z DEBUG kas_text::fonts::library] match: Adwaita Sans
[2025-06-09T08:10:55Z DEBUG kas_text::fonts::library] match: DejaVu Sans
[2025-06-09T08:10:55Z DEBUG kas_text::fonts::library] match: Carlito
[2025-06-09T08:10:55Z DEBUG kas_text::fonts::library] match: Noto Sans Mono
[2025-06-09T08:10:55Z DEBUG kas_text::fonts::library] match: Liberation Mono
[2025-06-09T08:10:55Z DEBUG kas_text::fonts::library] match: Liberation Serif
[2025-06-09T08:10:55Z DEBUG kas_text::fonts::library] match: Droid Sans Fallback
[2025-06-09T08:10:55Z DEBUG kas_text::fonts::library] match: Noto Sans Math
[2025-06-09T08:10:55Z DEBUG kas_text::fonts::library] match: Noto Sans Symbols 2
[2025-06-09T08:10:55Z DEBUG kas_text::fonts::library] match: OpenSymbol
```

This *does* solve matching `Char 'Ａ'=0xff21` : FULLWIDTH LATIN CAPITAL LETTER A. This was problematic previously because it *is* Latin, yet isn't included in common scripts. I think we could also solve this using #377 however (using the as-yet-unimplemented part of falling back to any font supporting the given UnicodeRange) since this has its own range: 68 = Halfwidth And Fullwidth Forms.

nicoburns:
Wanted to note that this doesn't necessarily have to be *instead* of the existing linux backend. We could have both and let people choose. Not saying we necessarily want to to do that in this case, but just raising it as a possibility. We could probably even abstract the backends out behind a trait if we wanted to (making it easier for users to bring their own custom backends).

nicoburns:
> Additional note: GenericFamily::UiSerif and other Ui* generic families still don't match anything. As I said before, these should likely be removed.

`ui-*` generic families seem to be a thing on macOS (and in a draft web standard: https://drafts.csswg.org/css-fonts-4/#ui-serif-def). Them currently not matching anything on Linux seems to be expected behaviour. If they are part of a font-stack then presumably that's no different to attempting to find a named font that isn't installed on the system. Perhaps we ought to just hardcode the non-macOS backends to return a non-match for these families for the time being?

valadaptive:
> I'm less keen on requiring non-Rust deps

The fontconfig bindings themselves are entirely Rust; they don't require integrating C into the build process. You also don't need to ship fontconfig with whatever application uses Parley; it's a platform library that we can expect to be available on all graphical Linux platforms. It's similar to DirectWrite or CoreText in that regard.

> most of these additional fonts won't normally do anything (neither matching glyphs nor causing overhead)

Fontconfig has an option when sorting fonts by Unicode coverage to remove any fonts later in the list that don't provide any more coverage than earlier fonts. That option is already enabled in this PR, so I think all the matched fonts do genuinely provide at least a bit more Unicode coverage.



> Wanted to note that this doesn't necessarily have to be _instead_ of the existing linux backend. We could have both and let people choose. Not saying we necessarily want to to do that in this case, but just raising it as a possibility. We could probably even abstract the backends out behind a trait if we wanted to (making it easier for users to bring their own custom backends).

I don't want to do that in this particular case; the previous backend was very half-baked. One possibility is using the existing folder-scanning implementation (used for the Android and macOS backends) to implement a much simpler Linux backend. I don't think there is any benefit to an implementation that parses fontconfig files but deliberately avoids calling into fontconfig; AFAIK, its ABI has been stable since its introduction in the 90s and there aren't really going to be any systems that have fontconfig files but no fontconfig library.

I do have a 20%-finished Rust reimplementation of fontconfig on the backburner, but even then I think dynamically linking to the platform's fontconfig is better, because there's no risk of fontconfig adding a new config option that the application's bundled implementation can't parse, or updating their cache file format. Of particular note is that the version of `fontconfig-cache-parser` we're currently using supports [cache file versions 7 and 8](https://github.com/jneem/fontconfig-cache-parser/blob/d38b30e2d902be9aa72f171ff2275ad97345b2d2/src/data.rs#L27), and upstream is [on version 9](https://gitlab.freedesktop.org/fontconfig/fontconfig/-/blob/043f3e60fa40b3e63e42382cada5e0c33ecab2e1/configure.ac#L72).

xorgy:
I think that this makes sense, as lame as it is aesthetically. Thank you for providing the patch. I'll give it my own review but since it is significant I'd like to see more than one reviewer. :+ )

fontique/src/backend/fontconfig.rs
```diff
@@ -0,0 +1,686 @@
...
+        unsafe { (LIB.FcConfigSubstitute)(self.inner.as_ptr(), pattern.inner.as_ptr(), kind) };
+    }
+
+    fn font_sort<'me, 'ret, 'pat: 'ret>(
+        &'me self,
+        pattern: &'pat Pattern,
+        trim: bool,
+    ) -> MatchResult<FontSet<'ret>> {
+        let mut result = 0;
+        // The returned FcFontSet is for us to free
```
xorgy: nit: missing period.

fontique/src/backend/fontconfig.rs
```diff
@@ -0,0 +1,686 @@
...
+        let Some(font_set) = (unsafe {
+            FontSet::from_raw(
+                (LIB.FcConfigGetFonts)(config.inner.as_ptr(), FcSetSystem),
+                Ownership::Fontconfig,
+            )
+        }) else {
+            return Default::default();
+        };
+
+        // Populate the family name map
```
xorgy: nit: missing period.

fontique/src/backend/fontconfig.rs
```diff
@@ -0,0 +1,686 @@
...
+        let mut name_map = FamilyNameMap::default();
+        for pattern in font_set.iter() {
+            let mut i = 0;
+
+            let mut first_name_id = None;
+            // For fonts with more than one family name, the second one is
+            // *often* (but not always) an RBIZ name
```
xorgy: nit: missing period.

fontique/src/backend/fontconfig.rs
```diff
@@ -0,0 +1,686 @@
...
+
+            // There are a lot of duplicate font names in the substituted
+            // pattern. Keep track of which ones have already been added to the
+            // list.
+            let mut added_names = HashSet::new();
+
+            for font in font_set.iter() {
+                // Not sure if FcFontRenderPrepare performs any substitutions
+                // relevant to fallback family name matching, but it's a good
+                // idea to call it just in case
```
xorgy: nit: missing period.

fontique/src/backend/fontconfig.rs
```diff
@@ -0,0 +1,686 @@
...
+                let Some(font) = config.font_render_prepare(&pattern, &font) else {
+                    continue;
+                };
+                // Generic families can have more than one name, but the only
+                // one we care about is the first one
```
xorgy: nit: missing period.

fontique/src/backend/fontconfig.rs
```diff
@@ -0,0 +1,686 @@
...
+        if let Some(set) = locale_lang_set {
+            pattern.add_langset(FC_LANG, &set);
+        }
+        if let Some(set) = script_char_set {
+            pattern.add_charset(FC_CHARSET, set);
+        }
+
+        config.substitute(&mut pattern, FcMatchPattern);
+
+        // This calls FcFontRenderPrepare for us
```
xorgy: nit: missing period.

fontique/src/backend/fontconfig.rs
```diff
@@ -0,0 +1,686 @@
...
+        self.name_map.get(&family_name).map(FamilyName::id)
+    }
+}
+
+impl SystemFonts {
+    fn family_uncached(&mut self, id: FamilyId) -> Option<FamilyInfo> {
+        let config = self.config.as_ref()?;
+        let name = self.name_map.get_by_id(id).cloned()?;
+
+        // Match by family name
```
xorgy: nit: missing period.

fontique/src/backend/fontconfig.rs
```diff
@@ -0,0 +1,686 @@
...
+    (GenericFamily::Emoji, c"emoji"),
+    (GenericFamily::Math, c"math"),
+];
+
+/// Fontconfig seems to force RBIZ (regular, bold, italic, bold italic) when
+/// categorizing fonts. This removes those suffixes from family names so that
+/// we can match on all attributes.
+fn strip_rbiz(name: &str) -> &str {
+    // TODO(valadaptive): this seems incomplete. check fcname.c for their
+    // constants
```
xorgy: Regarding this, I think they are also matched case-insensitive in `fcname.c`

xorgy: If it was the weights, slants, and widths, then it would be something like
```rust
    const SUFFIXES: &[&str] = &[
        // Weights.
        " Thin",
        " ExtraLight",
        " UltraLight",
        " Light",
        " Book",
        " Regular",
        " Medium",
        " DemiBold",
        " SemiBold",
        " Bold",
        " ExtraBold",
        " UltraBold",
        " Black",
        " Heavy",
        // Slants.
        " Roman",
        " Italic",
        " Oblique",
        // Widths.
        " UltraCondensed",
        " ExtraCondensed",
        " Condensed",
        " SemiCondensed",
        " Normal",
        " SemiExpanded",
        " Expanded",
        " ExtraExpanded",
        " UltraExpanded",
    ];
```

xorgy: And `name.strip_suffix` would probably end up being `name.trim_end_matches`.

fontique/src/backend/fontconfig.rs
dhardy: This isn't a sentence so it doesn't deserve a period.

fontique/src/backend/fontconfig.rs
valadaptive: Now I'm starting to question if this is really the right approach. For instance, Monaspace Krypton's extra-light weight is named "ExLight", which isn't anywhere in fontconfig's list. [Sinkin Sans](https://www.fontsquirrel.com/fonts/sinkin-sans) includes the numeric weight in the RBIZ suffix. Lato has a "Hairline" weight.

Fontconfig provides the `FC_STYLE` property on fonts which *should* provide the RBIZ suffix, but for e.g. Lato Hairline Italic, the family names returned from fontconfig are "Lato" and "Lato Hairline", but the `FC_STYLE`s are "Hairline Italic" and "Italic"--there is nothing there that lets us strip just the "Hairline" suffix.

Looking at Lato Hairline Italic in otexplorer, its name table seems to have multiple `FAMILY_NAME`s  that differ by platform/encoding/language ID:

```
│   platform_id: 1                                    1       │ 00 01       │
│   encoding_id: 0                                            │ 00 00       │
│   language_id: 0                                            │ 00 00       │
│   name_id: FAMILY_NAME
│   length: 4                                                 │ 00 04       │
│   string_offset: +195                                       │ 00 C3       │
│     Lato                                                    │             │
```

```
│   platform_id: 3                                    18      │ 00 03       │
│   encoding_id: 1                                            │ 00 01       │
│   language_id: 1033                                         │ 04 09       │
│   name_id: FAMILY_NAME
│   length: 26                                                │ 00 1A       │
│   string_offset: +2300                                      │ 08 FC       │
│     Lato Hairline                                           │             │
```

That level of detail doesn't help us with fontconfig, which just returns both names with no other information, but does explain *why* it does so. I am not well-versed in OpenType but I think what we really want is the "typographic family name", which fontconfig doesn't give us.

Any ideas or preferences for what we should do here? I'm leaning towards just writing up this state of affairs in the TODO comment and punting on it for now.

xorgy: I think writing up is the correct move, especially if there's no regression. FontConfig is slowly being rewritten in Rust with Fontations, so this roll cake is being eaten two slices at a time. Eventually we'll meet in the middle I guess.

DJMcNab: > FontConfig is slowly being rewritten in Rust with Fontations

Well that's interesting, surprising and good news! Is there an announcement/plan of this anywhere, or is it just something that's quietly happening?

valadaptive: Testing out CSS font matching in Chrome and Firefox, `font-family: "Lato Heavy"` works but `font-family: "Lato Bold"` doesn't match any fonts. Maybe we should *just* strip "Bold" and "Italic"? I can't find a list of which suffixes are actually stripped or which code does so.

dfrg: > > FontConfig is slowly being rewritten in Rust with Fontations
> 
> Well that's interesting, surprising and good news! Is there an announcement/plan of this anywhere, or is it just something that's quietly happening?

This isn’t a complete replacement. We’re simply providing an alternative to the FreeType component that is used to build the font database. The motivation is that we’d like to completely remove the FreeType dependency from Chromium.

dfrg: > Testing out CSS font matching in Chrome and Firefox, `font-family: "Lato Heavy"` works but `font-family: "Lato Bold"` doesn't match any fonts. Maybe we should _just_ strip "Bold" and "Italic"? I can't find a list of which suffixes are actually stripped or which code does so.

I don’t think it’s necessary to strip Bold and Italic since those already fit into the RBIZ model. 

Since browsers seem to use whatever fontconfig gives them perhaps we should do the same and just remove this code.

valadaptive: Removed `strip_rbiz`, but noticed that `FamilyNameMap::add_alias` doesn't actually alias the given family name to the given ID--it [creates a new `FamilyId`](https://github.com/linebender/parley/blob/375ba1c64ecf0694f265ff4242f177cfc155c4f8/fontique/src/family_name.rs#L85) for the map entry. Is this intentional?

dfrg: Good catch. That looks like a bug to me.

valadaptive: I've put up a separate PR for that.

fontique/src/backend/fontconfig.rs
```diff
@@ -656,48 +655,3 @@ const GENERIC_FAMILY_NAMES: &[(GenericFamily, &CStr)] = &[
     (GenericFamily::Emoji, c"emoji"),
     (GenericFamily::Math, c"math"),
 ];
-
-/// Fontconfig seems to force RBIZ (regular, bold, italic, bold italic) when
-/// categorizing fonts. This removes those suffixes from family names so that
-/// we can match on all attributes.
```
khaledhosny: FWIW, I don’t think this comment is true. The first family name in both font files below is the typographic family name (name ID 16) and second one is the RBIZ family name (name ID 1), and similarly the first style is the typographic style name and the second one is the RBIZ style name:
```
$ fc-query Lato-Hairline.ttf
Pattern has 28 elts (size 32)
	family: "Lato"(s) "Lato Hairline"(s)
	style: "Hairline"(s) "Regular"(s)
	fullname: "Lato Hairline"(s)

$ fc-query Lato-HairlineItalic.ttf
Pattern has 28 elts (size 32)
	family: "Lato"(s) "Lato Hairline"(s)
	style: "Hairline Italic"(s) "Italic"(s)
	fullname: "Lato Hairline Italic"(s)
```

valadaptive: This makes sense actually. I guess I'd assumed fontconfig added all the family names with name ID 1 (see the comment thread above). But looking into fontconfig's source, [it seems it adds the typographic family first and then the RBIZ family](https://gitlab.freedesktop.org/fontconfig/fontconfig/-/blob/41f2ae7df48d192b938ab78217bece1fab09dd1b/src/fcfreetype.c#L1057-1068). I *think* in the case of multiple names under a given ID, it adds them all and deduplicates them?

khaledhosny: > I _think_ in the case of multiple names under a given ID, it adds them all and deduplicates them?

This usually happens with localized names, but I haven’t seen duplicated names with the same language.

DJMcNab (approved):
I don't think waiting for a second thorough review here is viable.

I've given this a test with Xilem, and it doesn't appear to have caused any regressions. Thanks!

xorgy:
Yup, this should be merged.

PR #380, opened 2025-06-13
Use the correct family ID when adding font aliases

valadaptive:
Discovered this as part of #378; splitting it off because it affects all the fontique backends.

When adding font aliases, the code was not actually using the family ID we passed in, but creating a fresh family ID each time. This meant that font aliases were all mapped to nonexistent family IDs, and all queries for a font by one of its aliases would fail.

I'm a bit concerned that this has the potential to cause regressions or other general weirdness, since all font aliases went unused due to this bug. Not sure how to test that other than to just give it a spin.

### DIFF
--- a/fontique/src/family_name.rs
+++ b/fontique/src/family_name.rs
@@ -82,7 +82,7 @@ impl FamilyNameMap {
             }
             let new_name = FamilyName {
                 name: name.into(),
-                id: FamilyId::new(),
+                id,
             };
             self.name_map.insert(key.as_bytes().into(), new_name);
         }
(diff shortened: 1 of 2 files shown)


fontique/src/family_name.rs
```diff
@@ -82,7 +82,7 @@ impl FamilyNameMap {
             }
             let new_name = FamilyName {
                 name: name.into(),
-                id: FamilyId::new(),
+                id,
             };
             self.name_map.insert(key.as_bytes().into(), new_name);
```
valadaptive, line 87: We can actually use [`entry_ref`](https://docs.rs/hashbrown/latest/hashbrown/hash_map/struct.HashMap.html#method.entry_ref) here; not sure if I should do that cleanup or leave it for another PR

dfrg (approved):
LGTM. This was certainly the intended behavior so it seems fine to land the change and address any potential issues with other backends if they arise.

PR #381, opened 2025-06-20
Select "paragraphs" rather than words for triple click

DJMcNab:
See discussion around [#parley > Sharing selection/editing code between Blitz and Masonry @ 💬](https://xi.zulipchat.com/#narrow/channel/205635-parley/topic/Sharing.20selection.2Fediting.20code.20between.20Blitz.20and.20Masonry/near/525065337)

### DIFF
--- a/examples/vello_editor/src/text.rs
+++ b/examples/vello_editor/src/text.rs
@@ -309,7 +309,7 @@ impl Editor {
                         let mut drv = self.editor.driver(&mut self.font_cx, &mut self.layout_cx);
                         match click_count {
                             2 => drv.select_word_at_point(cursor_pos.0, cursor_pos.1),
-                            3 => drv.select_line_at_point(cursor_pos.0, cursor_pos.1),
+                            3 => drv.select_paragraph_at_point(cursor_pos.0, cursor_pos.1),
                             _ => drv.move_to_point(cursor_pos.0, cursor_pos.1),
                         }
                     }
--- a/parley/src/layout/cursor.rs
+++ b/parley/src/layout/cursor.rs
@@ -487,6 +487,19 @@ impl Selection {
         }
     }
 
+    /// Creates a new selection bounding the paragraph at the given coordinates.
+    pub fn paragraph_from_point<B: Brush>(layout: &Layout<B>, x: f32, y: f32) -> Self {
+        let Self { anchor, focus, .. } = Self::from_point(layout, x, y)
+            .paragraph_start(layout, false)
+            .paragraph_end(layout, true);
+        Self {
+            anchor,
+            focus,
+            anchor_base: AnchorBase::Line(anchor, focus),
+            h_pos: None,
+        }
+    }
+
     #[cfg(feature = "accesskit")]
     pub fn from_access_selection<B: Brush>(
         selection: &accesskit::TextSelection,
@@ -719,6 +732,42 @@ impl Selection {
         }
     }
 
+    /// Returns a new selection with the focus moved to the start of the
+    /// current paragraph.
+    ///
+    /// If `extend` is `true` then the current anchor will be retained,
+    /// otherwise the new selection will be collapsed.
+    #[must_use]
+    pub fn paragraph_start<B: Brush>(&self, layout: &Layout<B>, extend: bool) -> Self {
+        if let Some((mut paragraph_start_index, line)) = self.focus.line(layout) {
+            let mut result_byte_index = line.text_range().start;
+            loop {
+                if paragraph_start_index == 0 {
+                    break;
+                }
+                let prev_index = paragraph_start_index - 1;
+                if let Some(line) = layout.get(prev_index) {
+                    if matches!(line.break_reason(), BreakReason::Explicit) {
+                        // The start of the line 'paragraph_start_index' is the target point.
+                        break;
+                    }
+                    result_byte_index = line.text_range().start;
+                    paragraph_start_index = prev_index;
+                } else {
+                    // We think this is unreachable, but to best not panic here.
+                    // Ideally, we'd log an error, but don't seem to have `log` nor `tracing` set up here.
+                    return *self;
+                }
+            }
+            self.maybe_extend(
+                Cursor::from_byte_index(layout, result_byte_index, Affinity::Downstream),
+                extend,
+            )
+        } else {
+            *self
+        }
+    }
+
     /// Returns a new selection with the focus moved to the end of the
     /// current line.
     ///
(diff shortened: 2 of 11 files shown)


xorgy:
Sorry I didn't see this until now, I'll take a look. :+ )

xorgy:
It behaves well for me in the covered case (two consecutive line breaks). Going forward we should have some better tools for multi-paragraph layouts. The stated design intent of Parley as it is, is that users should produce a layout for each paragraph. There are also other considerations like true paragraph separators (`U+2029 PARAGRAPH SEPARATOR`), and the distinction between a paragraph break and two consecutive line breaks.

With all that having been said, I think this is a useful behavior to support, even if it is probably not how Parley should be used in ‘real’ multi-paragraph layouts.

parley/src/layout/cursor.rs
```diff
@@ -719,6 +732,42 @@ impl Selection {
...
+                if let Some(line) = layout.get(prev_index) {
+                    if matches!(line.break_reason(), BreakReason::Explicit) {
+                        // The start of the line 'paragraph_start_index' is the target point.
+                        break;
+                    }
+                    result_byte_index = line.text_range().start;
+                    paragraph_start_index = prev_index;
+                } else {
+                    // We think this is unreachable, but to best not panic here.
+                    // Ideally, we'd log an error, but don't seem to have `log` nor `tracing` set up here.
```
xorgy: Probably we should whiteboard this scenario, and get it to the point where we can confidently put `unreachable!` here. May be worth a couple tests.

parley/src/layout/cursor.rs
```diff
@@ -719,6 +732,42 @@ impl Selection {
...
+    #[must_use]
+    pub fn paragraph_start<B: Brush>(&self, layout: &Layout<B>, extend: bool) -> Self {
+        if let Some((mut paragraph_start_index, line)) = self.focus.line(layout) {
+            let mut result_byte_index = line.text_range().start;
+            loop {
+                if paragraph_start_index == 0 {
+                    break;
+                }
+                let prev_index = paragraph_start_index - 1;
+                if let Some(line) = layout.get(prev_index) {
```
xorgy: Since this is expected to be infallible, maybe this should be let-else for better readability.

examples/vello_editor/src/text.rs
```diff
@@ -309,7 +309,7 @@ impl Editor {
                         let mut drv = self.editor.driver(&mut self.font_cx, &mut self.layout_cx);
                         match click_count {
                             2 => drv.select_word_at_point(cursor_pos.0, cursor_pos.1),
-                            3 => drv.select_line_at_point(cursor_pos.0, cursor_pos.1),
+                            3 => drv.select_paragraph_at_point(cursor_pos.0, cursor_pos.1),
```
xorgy: Might be nice to exercise `move_to_paragraph_start` and `move_to_paragraph_end` with <kbd>Ctrl</kbd> + <kbd>🠝</kbd> and <kbd>Ctrl</kbd> + <kbd>🠟</kbd>, `select_to_paragraph_start` and `select_to_paragraph_end` with <kbd>Ctrl</kbd> + <kbd>Shift</kbd> + <kbd>🠝</kbd> and <kbd>Ctrl</kbd> + <kbd>Shift</kbd> + <kbd>🠟</kbd>.

parley/src/layout/editor.rs
```diff
@@ -655,12 +707,27 @@ where
     }
 
     /// Select the physical line at the point.
+    ///
+    /// Note that this metehod determines line breaks for any reason, including due to word wrapping.
+    /// To select the text between explicit newlines, use [`select_paragraph_at_point`](Self::select_paragraph_at_point).
+    /// In most text editing cases, this is the preferred behaviour.
+    // TODO: Rename to `select_soft_line_at_point`? Very few users should use this method.
```
xorgy: I think that's a fair change, maybe there should still be `select_line_at_point` for explicit lines; this is a desirable behavior in code editors.

xorgy (changes requested):

parley/src/layout/editor.rs
DJMcNab: What actually is the operation exposed by `select_line_at_point`?
I think that it has the same semantics as  `select_paragraph_at_point` as implemented in this PR.

examples/vello_editor/src/text.rs
DJMcNab: Is that standard behaviour in any apps? I'm hesitant to introduce behaviour in this file which users shouldn't reproduce, because this code is likely to be closely copied into any user code (indeed, this same code is effectively copied into Masonry)

parley/src/layout/cursor.rs
DJMcNab: I've convinced myself that this is truly unreachable; it's not rocket science.

examples/vello_editor/src/text.rs
xorgy: Ope, I didn't realize I'm only familiar with these from Emacs.... they may be less common than I thought :joy:

nicoburns: At some point we'll want to hook these up to the [apple standard keybindings](https://developer.apple.com/documentation/appkit/nsstandardkeybindingresponding?language=objc). Many of which are bound to emacs-style keybindings by default. But we probably don't want to detect the keybindings directly in that case as the system does that (keybindings are customisable as a system preference) and sends us a string (like `moveToEndOfParagraph:`) that represents a high-level editing action directly.

(I implemented this in Winit so these events should be in the next winit release)

parley/src/layout/editor.rs
xorgy: Oh, sorry I misread the code, I thought you were checking for start of text, end of text, and _two consecutive hard line breaks_. A single hard line break is just `select_line_at_point`.

xorgy (changes requested):
Now that I have had a closer look, I realize that this isn't even a commonly accepted definition of a paragraph. It would be understandable to call two consecutive hard breaks a paragraph, but a single hard break is just a line. If these are renamed to _line_ rather than _paragraph_ though, I'm happy with the changes.

DJMcNab:
Just doing `s/paragraph/line` would not be good, because almost everything else in Parley uses "line" to mean soft lines. I'm not going to make a PR to rename every use of "line" to soft line; and I don't think it would be reasonable to block this PR on such a refactor.

I've used "hard line" instead, as the natural mirror to soft line.

If you want to follow up by renaming line to soft line, so that we can use the name "line" for this case, then feel free (although I'm not planning to review that PR).

examples/vello_editor/src/text.rs
PoignardAzur: As much as I like the "ctrl+up = move to paragraph start" behavior, no, it's **very much** isn't standard.

parley/src/layout/cursor.rs
```diff
@@ -719,6 +734,42 @@ impl Selection {
...
+                if hard_line_start_index == 0 {
+                    break;
+                }
+                let prev_index = hard_line_start_index - 1;
+                let Some(line) = layout.get(prev_index) else {
+                    unreachable!(
+                        "{hard_line_start_index} is a valid line in the layout, but {prev_index} isn't, despite the latter being smaller.\n\
+                        The layout has {} lines.",
+                        layout.len()
+                    );
```
PoignardAzur, line 755: I don't think you need a message that detailed for an `unreachable!()` macro FWIW.

PoignardAzur (approved):
LGTM.

xorgy (approved):
Thanks... maybe some day we'll have the real deal. :+ )

PR #382, opened 2025-06-26
vello_editor: Use ui-events-winit for keyboard and pointer events.

xorgy:
This simplifies touch handling, drag state, and click counting.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1563,7 +1563,7 @@ dependencies = [
  "js-sys",
  "log",
  "wasm-bindgen",
- "windows-core 0.58.0",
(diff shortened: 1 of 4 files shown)


waywardmonkeys (approved):
Does this differ at all from what's in the android_view demo?

xorgy:
> Does this differ at all from what's in the android_view demo?

Ever so slightly. The android-view demo one uses raw keycodes (which are not unmapped like they are on other platforms), and it supports shift-click to extend a selection: 
https://github.com/rust-mobile/android-view/blob/07a4660ecb56c2875548ad6c4310e20bf189505a/demo/src/text.rs#L326-L329

PR #385, opened 2025-06-30
Implement shift-click

kekelp:
This is a pull request to implement shift-click selection extension.

Shift-click extends the selection starting from the anchor base, but always using cluster granularity. As far as I could tell, there's no way to do this with `Selection`'s current public API. The new function `Selection::extend_to_precise_point` allows doing this.

The function is very similar to the existing `Selection::extend_to_point`, but I created a new one instead of modifying it, since I'm not familiar with the backwards compatibility requirements.

### DIFF
--- a/parley/src/layout/cursor.rs
+++ b/parley/src/layout/cursor.rs
@@ -864,6 +864,33 @@ impl Selection {
         }
     }
 
+    /// Returns a new selection with the focus extended to the given point.
+    #[must_use]
+    pub fn extend_to_precise_point<B: Brush>(&self, layout: &Layout<B>, x: f32, y: f32) -> Self {
+        let cluster = Cursor::from_point(layout, x, y);
+        match self.anchor_base {
+            AnchorBase::Cluster => Self::new(self.anchor, cluster),
+            AnchorBase::Word(start, end) => {
+                let [anchor, focus] = cursor_min_max(layout, [cluster, cluster, start, end]);
+                Self {
+                    anchor,
+                    focus,
+                    anchor_base: self.anchor_base,
+                    h_pos: None,
+                }
+            }
+            AnchorBase::Line(start, end) => {
+                let [anchor, focus] = cursor_min_max(layout, [cluster, cluster, start, end]);
+                Self {
+                    anchor,
+                    focus,
+                    anchor_base: self.anchor_base,
+                    h_pos: None,
+                }
+            }
+        }
+    }
+
     /// Returns a new selection with the current anchor and the focus set to
     /// the given value.
     #[must_use]
--- a/parley/src/layout/cursor.rs
+++ b/parley/src/layout/cursor.rs
@@ -841,8 +841,7 @@ impl Selection {
             AnchorBase::Cluster => Self::new(self.anchor, Cursor::from_point(layout, x, y)),
             AnchorBase::Word(start, end) => {
                 let target = Self::word_from_point(layout, x, y);
-                let [anchor, focus] =
-                    cursor_min_max(layout, [target.anchor, target.focus, start, end]);
+                let [anchor, focus] = extend_selection(target, [start, end]);
                 Self {
                     anchor,
                     focus,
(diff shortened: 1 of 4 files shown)


parley/src/layout/cursor.rs
```diff
@@ -864,6 +864,33 @@ impl Selection {
...
+                let [anchor, focus] = cursor_min_max(layout, [cluster, cluster, start, end]);
+                Self {
+                    anchor,
+                    focus,
+                    anchor_base: self.anchor_base,
+                    h_pos: None,
+                }
+            }
+            AnchorBase::Line(start, end) => {
+                let [anchor, focus] = cursor_min_max(layout, [cluster, cluster, start, end]);
```
DJMcNab: I think changing `cursor_min_max` to accept a shared slice of cursors as part of this PR would be reasonable (if that isn't a breaking change). Duplicating `cluster` here just to make this work is pretty bad.

parley/src/layout/cursor.rs
```diff
@@ -864,6 +864,33 @@ impl Selection {
...
     }
 
+    /// Returns a new selection with the focus extended to the given point.
+    #[must_use]
+    pub fn extend_to_precise_point<B: Brush>(&self, layout: &Layout<B>, x: f32, y: f32) -> Self {
+        let cluster = Cursor::from_point(layout, x, y);
+        match self.anchor_base {
+            AnchorBase::Cluster => Self::new(self.anchor, cluster),
+            AnchorBase::Word(start, end) => {
+                let [anchor, focus] = cursor_min_max(layout, [cluster, cluster, start, end]);
```
DJMcNab: The focus always being the latter of these points doesn't quite feel right to me, but it's possible and/or likely that I'm mistaken.

DJMcNab (commented):
Welcome! Thanks for this - it looks pretty good.

Please can you also give this a CHANGELOG entry. Note that you will need to create new links at the bottom of the file for your name and the PR's number.

parley/src/layout/cursor.rs
kekelp: I think you're right. It's wrong both here and in the preexisting `extend_to_point`. Example:
- select a word with double-click
- hold and drag leftwards to extend to the previous word 
- now, shift + left/right will extend from the rightmost part of the selection, but it should be from the left.

I'll see if I can fix it.

parley/src/layout/cursor.rs
```diff
@@ -852,8 +851,29 @@ impl Selection {
...
+    }
+
+    /// Returns a new selection with the focus extended to the given point.
+    #[must_use]
+    pub fn shift_click_extension<B: Brush>(&self, layout: &Layout<B>, x: f32, y: f32) -> Self {
+        let target = Cursor::from_point(layout, x, y);
+        match self.anchor_base {
+            AnchorBase::Cluster => Self::new(self.anchor, target),
+            AnchorBase::Word(start, end) | AnchorBase::Line(start, end) => {
+                let [anchor, focus] = if target.index < start.index {
```
DJMcNab:
```suggestion
                // Place the focus where the user just clicked, and the anchor on the "far" side of the anchorbase.
                let [anchor, focus] = if target.index < start.index {
```

I think a little explanatory comment is helpful here; the meaning of the condition is not necessarily obvious.

parley/src/layout/cursor.rs
```diff
@@ -852,8 +851,29 @@ impl Selection {
...
+                    [end, target]
+                } else {
+                    [start, target]
+                };
                 Self {
                     anchor,
                     focus,
```
DJMcNab: Thoughts on:
```suggestion
                let anchor = if target.index < start.index {
                    end
                } else {
                    start
                };
                Self {
                    anchor,
                    focus: target,
```
The amount of space taken up for such a simple condition is a bit unfortunate, but it keeps the code clear.

DJMcNab (approved):
This looks really good - just a couple of thoughts on code clarity, but this could land as-is.

I'd like to get sign-off from @xorgy as well.

It would also be nice to see a couple of tests, but IIRC we don't really have the infra for "by point" tests, so it can be skipped. This still needs to be added to the CHANGELO

kekelp:
Thanks for the feedback. I accepted both of the code clarity suggestions and added the changelog entry.

xorgy (approved):
Thank you for being detailed about this, I agree with your choices here. :+ )

PR #389, opened 2025-07-09
Rename `Alignment` variants to better match CSS

waywardmonkeys:
* `Alignment::Justified` is now `Alignment::Justify`
* `Alignment::Middle` is now `Alignment::Center`


dfrg (approved):
Also seems reasonable to rename `Middle` in this PR. 

If we’re trying to match CSS, should that type name be `TextAlign`?

waywardmonkeys:
Okay, will update this to include `Middle` after I sleep. Not sure why I didn't think about that one.

PR #390, opened 2025-07-16
Update to accesskit 0.21

mwcampbell:


mwcampbell:
I need this so I can use the new accesskit release in masonry.

nicoburns (approved):
Seems reasonable, although I would note that will bring in a duplicate `windows` dependency for anyone using `wgpu` as the new `accesskit` is on `0.61` while `wgpu` is still on `0.58`.

You might want to consider specifying a range dependency unless `accesskit` actually requires new features (I would guess that no actual code changes were required for the upgrade, and thus that a range dependency would "just work").

nicoburns:
Good point!

Issue #327, opened 2025-03-30
Editor: password input mode


nicoburns:
A password input should have an underlying string as usual, but should visually render a `•` (we could also make the character configurable).

xorgy:
Seems like this maybe can be solved with a font.

waywardmonkeys:
Possibly complicating things is that some password entries echo the character that you typed for a second or so and then it switches to the obscured character.

PR #394, opened 2025-07-24
Implement the ability to hide the editor text with a given repeated character

StephanvanSchaik:
I am currently using a text input field in xilem as a password field, but there is currently no way to hide the text by replacing the actual buffer of characters with a repeated character such as an asterisk (i.e., `'*'`). Further looking into xilem and masonry, I found that the text input field leverages `PlainEditor` in parley, so I am trying to extend that first.

More specifically, this commit adds a `visible_buffer` field to hold the character buffer with the replaced characters and a `hide_symbol` field to hold the character that needs to be shown (if any is set). This is to ensure that the original `buffer` field stays intact everywhere else (together with the corresponding logic), and this seems to be the least intrusive way to achieve this. Further, I have extended `update_layout` to update the `visible_buffer` field if the symbol or the length of the original buffer changed and to decide which buffer needs to be shown based on whether the symbol is set or not. Finally, this adds two methods to set and clear the symbol used for hiding the text.

Extending this functionality to masonry/xilem should be as simple as calling those two methods once this lands. I think this is also neat because this makes it simple to add a show/hide password checkbox if desired.

Issue #327 is related, but we should keep that open even if this lands, because there are probably a bunch of corner cases or additional desired features, I am just trying to get the most awkward part out of the way which is that the password shouldn't be visible on the screen.

### DIFF
--- a/parley/src/layout/mod.rs
+++ b/parley/src/layout/mod.rs
@@ -22,6 +22,7 @@ use accesskit::{Node, NodeId, Role, TextDirection, TreeUpdate};
 use alignment::unjustify;
 #[cfg(feature = "accesskit")]
 use alloc::vec::Vec;
+use alloc::{string::String, sync::Arc};
 use core::{cmp::Ordering, ops::Range};
 use data::{ClusterData, LayoutData, LayoutItem, LayoutItemKind, LineData, LineItemData, RunData};
 #[cfg(feature = "accesskit")]
@@ -314,6 +315,8 @@ pub struct Style<B: Brush> {
     pub(crate) line_height: LayoutLineHeight,
     /// Per-cluster overflow-wrap setting
     pub(crate) overflow_wrap: OverflowWrap,
+    /// Replace graphemes with another grapheme.
+    pub(crate) grapheme_replacement: Option<Arc<String>>,
 }
 
 /// Underline or strikethrough decoration.
--- a/parley/src/resolve/mod.rs
+++ b/parley/src/resolve/mod.rs
@@ -8,7 +8,7 @@ pub(crate) mod tree;
 
 pub(crate) use range::RangedStyleBuilder;
 
-use alloc::{vec, vec::Vec};
+use alloc::{string::String, sync::Arc, vec, vec::Vec};
 
 use super::style::{
     Brush, FontFamily, FontFeature, FontSettings, FontStack, FontStyle, FontVariation, FontWeight,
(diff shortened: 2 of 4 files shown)


StephanvanSchaik:
I see that `cargo clippy` fails, but this should be affecting source files that this PR does not touch. I can fix the issues if desired.

nicoburns:
I'm also very keen to have this functionality, but I think we'll need more sophisticated logic to handle character counting (one `char` does not always correspond to one visible character), selection and editing.

StephanvanSchaik:
> I'm also very keen to have this functionality, but I think we'll need more sophisticated logic to handle character counting (one `char` does not always correspond to one visible character), selection and editing.

Oh, that is right, I forgot about this entirely (it is sometimes too easy to forget). Thanks for the reminder.

xorgy:
This is a desirable feature, but this is not the right design I think. Introducing more segmentation is not necessary if we use a style instead, for example; and it lets you set it up simply with a default style, or a range/tree style (and eventually, through Attributed Text).
Adding extra state to the editor, separate segmentation rules, etc. is not how we would solve this.

My suggestion is to address it through styles, call it grapheme replacement, maybe make it use a &'static str (or a fixed length string type which reasonably accommodates values)... but there may be complications with that, which I have not thought of.

nicoburns:
Sorry for the extended delay in replying to this!

> I am currently using `Option<Arc<String>>` for the replacement grapheme,

Probably we ought to use `Option<SmolStr>` from the smol_str crate, and or even just `Option<char>`. Although `Option<Arc<String>>` is... fine.

> I also set grapheme_replacement to None in the test_builder.rs test

Seems correct. We'll probably want some tests specifically for grapheme replacement too, but wouldn't want it enabled in existing tests.

> Finally, I am also unsure about whether the output layout should still contain Option<Arc<String>> or whether that grapheme should be processed such that we can use draw_glyphs directly in masonry_core.

I think the output layout shoudl be processed such that we can use draw_glyphs directly. I think this will be important so that things like mouse selection work properly (as they'll have the visually correct dimensions for the replaced glyphs).

I think the replacement needs to happen in the core shaping pipeline once the text is "itemized" into to shaping clusters. Instead of (or after) mapping the cluster to glyphs by shaping, we should map the cluster to the replacement glyph (the replacement character itself should have it's glyph determined by shaping, although we may want to limit it to max one glyph).

PR #395, opened 2025-07-28
Add unregister font method

taj-p:


nicoburns (approved):
This seems reasonable to me as-is (consistent with our existing APIs). However, I wonder whether we ought to introduce a unique ID for each specific font which:

- Would be returned from `register_fonts` instead of / as well as the family name and font info
- Could be used to lookup font data and metadata
- Could be used to unregister a specific font

taj-p:
> This seems reasonable to me as-is (consistent with our existing APIs). However, I wonder whether we ought to introduce a unique ID for each specific font which:
> 
> * Would be returned from `register_fonts` instead of / as well as the family name and font info
> * Could be used to lookup font data and metadata
> * Could be used to unregister a specific font

I really like this idea. I think that might be a more substantial change so I'll explore this idea separately. One question I have that is related to this is whether `FamilyId` or the newly proposed `FontId` should be able to be passed through Parley's builders?

Like:

## Today

```rs
    builder.push(
        StyleProperty::FontStack(parley::FontStack::Single(parley::FontFamily::Named(
            Cow::Borrowed("Open Sans"),
        ))),
        0..10,
    );
```

## Passing FamilyId/FontId

```rs
    builder.push(
        StyleProperty::FontStack(parley::FontStack::Single(parley::FontFamily::FamilyId(family_id))),
        0..10,
    );
    builder.push(
        StyleProperty::FontStack(parley::FontStack::Single(parley::FontFamily::FontId(font_id))),
        0..10,
    );
```

PR #401, opened 2025-08-22
Use a range to store deferred boxes during shaping

nicoburns:
# Objective

- Eliminate an unneccesary allocation
- Very slightly reduce memory usage

## Changes made

Use a range instead of a vec to store "deferred boxes" during the shaping process.

## Context

This came up as review feedback during the initial implementation of inline boxes but was deferred (ha) as a followup

### DIFF
--- a/parley/src/shape.rs
+++ b/parley/src/shape.rs
@@ -5,6 +5,7 @@
 //! and `swash` for text analysis.
 
 use core::mem;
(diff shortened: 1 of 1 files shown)


taj-p (approved):
Always great to save an allocation!

> was deferred (ha) as a followup

😆

PR #404, opened 2025-08-25
PlainEditor: Fix regression on deleting a selection from text start.

### LINKED ISSUE #403: PlainEditor: Cursor appears to advance after deleting selection from text start. [regression, v0.2..v0.3]
In c0d158b5d77c5726ed33fd3f2cbfcde95df374ef (edit: i.e. #170)  (between v0.2.0 and v0.3.0) there was a regression introduced around the handling of deletion of a selection that includes the first cluster in the layout, where when that selection is deleted, the visible caret advances to the downstream side of the first cluster (and this also makes it not possible to move the cursor one cluster downstream until it is reset by going the other way).

xorgy:
Fixes #403

### DIFF
--- a/parley/src/layout/cursor.rs
+++ b/parley/src/layout/cursor.rs
@@ -25,9 +25,15 @@ impl Cursor {
     /// Creates a new cursor from the given byte index and affinity.
     pub fn from_byte_index<B: Brush>(layout: &Layout<B>, index: usize, affinity: Affinity) -> Self {
         if let Some(cluster) = Cluster::from_byte_index(layout, index) {
+            let index = cluster.text_range().start;
             Self {
-                index: cluster.text_range().start,
-                affinity,
+                index,
+                affinity: if index != 0 {
+                    affinity
+                } else {
+                    // There is no Upstream cluster of the 0 position so we force Downstream affinity.
+                    Affinity::Downstream
+                },
             }
         } else {
             Self {


parley/src/layout/editor.rs
```diff
@@ -1155,7 +1155,7 @@ where
 
         self.update_layout(font_cx, layout_cx);
         let new_index = start.saturating_add(s.len());
-        let affinity = if s.ends_with("\n") {
+        let affinity = if s.ends_with("\n") || (start == 0 && s.is_empty()) {
```
DJMcNab: I *think* it would make sense to move this check into `Cursor::from_byte_index` (i.e. `if idx==0, always use downstream affinity`). That would nicely mirror the case for if the byte index is past the end of the string (where this affinity code is also wrong). Thoughts?

xorgy: I'll take a look at that. I did it here because it is the only case where it seems to present an issue, and because it means `Cursor::from_byte_index` isn't silently ignoring one of its parameters in another case, which to me is confusing (I have issues with almost everything to do with Cursor/Selection though lol, like how the affinities are named).

parley/src/layout/cursor.rs
```diff
@@ -25,9 +25,17 @@ impl Cursor {
     /// Creates a new cursor from the given byte index and affinity.
     pub fn from_byte_index<B: Brush>(layout: &Layout<B>, index: usize, affinity: Affinity) -> Self {
         if let Some(cluster) = Cluster::from_byte_index(layout, index) {
-            Self {
-                index: cluster.text_range().start,
-                affinity,
+            let index = cluster.text_range().start;
+            if index != 0 {
+                Self {
+                    index: cluster.text_range().start,
```
DJMcNab:
```suggestion
                    index,
```

parley/src/layout/cursor.rs
```diff
@@ -25,9 +25,17 @@ impl Cursor {
...
+                    affinity,
+                }
+            } else {
+                Self {
+                    index,
```
DJMcNab:
```suggestion
                    index: 0,
```

parley/src/layout/cursor.rs
```diff
@@ -25,9 +25,17 @@ impl Cursor {
...
+                    affinity: Affinity::Downstream,
```
DJMcNab:
```suggestion
                    //  There is no upstream cluster of the 0 position, so we force downstream affinity.
                    affinity: Affinity::Downstream,
```

parley/src/layout/editor.rs
xorgy: I've included something like what you suggested as the second commit. It seems yet again we have six of one, half dozen of another type situation. :+ )
Feel free to break the tie either way, if you can't think of any way the `from_byte_index` change is incorrect. It seems to work correctly at a surface level when I play with it, and makes sense in isolation, but honestly I don't know.

DJMcNab (approved):
The mirroring here makes a lot more sense to me. I've tested and can reproduce the original issue, and see that this fix works. This approval is conditional on the suggested changes, which should all be uncontroversial.

Thanks!

PR #405, opened 2025-08-25
Parley Benchmarks

taj-p:
Adds in an initial set of benchmarks to aid Parley development.

I've opted to use [Tango](https://github.com/bazhenov/tango) because I like the quick development loop. We can run 20 benchmark cases in ~10 seconds. It also "feels" more reliable.

Here's what its output looks like (results from an upcoming caching PR).

image

## Extra

- Moved font assets to `parley_dev`
- Created text samples for use in benchmarks and around Parley in `parley_dev/assets/text_samples`

cc @conor-93


xorgy (approved):
Looks good and runs for me. I get a roughly ± 2 % noise floor (comparing this branch to itself in Tango) when running with a pinned CPU and quieted IRQs on the thread... which is not too bad!
I settled on Divan recently for one of my things, because the less-sophisticated statistical analysis and research-driven sampling method helped a lot with certain types of micro benchmarks that were completely impossible to measure reliably with Criterion (due to a weird sampling behavior, or maybe Criterion was somehow poisoning the branch predictor for this particular function.. unsure).
Tango is very cool, I like the differential paired timing analysis, reminds me of balanced signal pairs/differential signalling. :+ )

parley_bench/.gitignore
```diff
@@ -0,0 +1 @@
+target
```
DJMcNab:
```suggestion
/target
```

I'd probably use `/target` here again, because we don't want to automatically ignore any child directories or files named `target`. This is the nittiest of nits.

parley_bench/build.rs
```diff
@@ -0,0 +1,9 @@
+// Copyright 2025 the Parley Authors
+// SPDX-License-Identifier: Apache-2.0 OR MIT
+
+//! # Parley Bench Build Script
+
+fn main() {
```
DJMcNab, line 6: Presumably, this is the boilerplate required by Tango? We should probably comment that there

parley_bench/README.md
```diff
@@ -0,0 +1,19 @@
+# Parley Benchmarks
+
+A suite of benchmarks used to evaluate Parley performance.
+
+## Setup
+
+Install `cargo-export` via `cargo install cargo-export`.
```
DJMcNab: We should use the follow to make it easier to copy-and-paste:

````suggestion
Install [`cargo-export`](https://crates.io/crates/cargo-export) via: 
```sh
$ cargo install cargo-export
```
````
We could also consider:
````suggestion
Install [`cargo-export`](https://crates.io/crates/cargo-export) via: 
```sh
$ cargo install cargo-export --locked
```
````

But I note that cargo-export itself doesn't suggest to use `--locked`. I'd stand by either suggestion.

File: parley_bench/README.md [hunk 4]
```diff
@@ -0,0 +1,19 @@
+# Parley Benchmarks
+
+A suite of benchmarks used to evaluate Parley performance.
```
DJMcNab: We should link to Tango somewhere here, so people have a chance to understand what's actually going on.

It looks like the right link is <https://github.com/bazhenov/tango>. Something like:
```md
This uses [Tango](https://github.com/bazhenov/tango) to perform paired benchmarking.
```

parley_dev/assets/text_samples/README.md
```diff
@@ -0,0 +1,11 @@
+# Sample Text Attribution
+
+All samples are licensed under the [Creative Commons Attribution-ShareAlike License](https://creativecommons.org/licenses/by-sa/4.0/) as copied to ./LICENSE.
```
DJMcNab: Wikisource documents that Moby Dick is in the public domain (see the note at the bottom of <https://en.wikisource.org/wiki/Moby-Dick_(1851)_US_edition>).
It does the same for the Tosa Diary source - if you see the note at the bottom of [that source](https://ja.wikisource.org/wiki/%E5%9C%9F%E4%BD%90%E6%97%A5%E8%A8%98_(%E5%9C%8B%E6%96%87%E5%A4%A7%E8%A7%80)), it states that it's in the public domain in the US. We are already relying on this public domain status for them to be licensed under the license which we claim, and I believe that the intention of Wikisource's license is to cover all the copyrightable content (e.g. the intro at the top of the Moby Dick page).

The situation for Al-Kindi - First Philosophy is slightly less clear, but given that the texts were written more than 1000 years ago, I imagine the situation is similar. I've not found any sources which discuss this (in English).

As such, I think we have two pathways here:
- We can either just use them without copying in the license, still mentioning that we got them from Wikisource. Since they are in the public domain. This is the direction I'd choose, if you agree with my analysis.
- Use this form, which also requires us to mention that these files are licensed as discussed in the root README. That is, this is a step which hasn't been done in this PR; the text at https://github.com/linebender/parley?tab=readme-ov-file#license would need to be updated.

parley_dev/src/lib.rs
```diff
@@ -0,0 +1,71 @@
...
+}
+
+/// The font families that are available in the assets/fonts directory.
+pub const FONT_FAMILIES: &[&str] = &["Arimo", "Roboto", "Noto Kufi Arabic"];
+
+/// A sample to be used for development.
+#[derive(Debug)]
+pub struct Sample {
+    /// The name of the sample.
+    pub name: &'static str,
```
DJMcNab, line 28: I'd be tempted to also add a "title" field here, although I'm not sure where we'd use it.

parley_dev/assets/fonts/arimo_fonts/Arimo-VariableFont_wght.ttf
DJMcNab, line 1: As this file has moved, it's reference in the root README needs to be updated. This is blocking

parley_dev/Cargo.toml
DJMcNab, line 1: I'd like for this folder to have a README. It should have an intro:
```md
Data and fonts used in Parley's tests and benchmarks.
```
and something like:
```
Some files included in this folder are under different licenses to the rest of this workspace:

- The font file `Arimo-VariableFont_wght.ttf` in `./assets/arimo_fonts/` is licensed solely as documented in that folder (and is licensed under the Apache License, Version 2.0).
- The font file `Roboto-Regular.ttf` in `./assets/roboto_fonts/` is licensed solely as documented in that folder (and is licensed under the Apache License, Version 2.0).
- The font file `NotoKufiArabic-Regular.otf` in `./assets/noto_fonts/` is licensed solely as documented in that folder (and is licensed under the SIL Open Font License, Version 1.1).

```

DJMcNab (changes requested):
There is one blocker, which is updating the file paths for the fonts in the root README (and also removing them from `parley/README.md`)

DJMcNab:
By the way, when I spoke to Denis at RustWeek, they mentioned that they have had some success using this technique on GitHub's CI runners (~10% noise floor). It could be worth experimenting with automatically running some of these benchmarks, and somehow reporting results (they need only be used as a signal to trigger manual benchmarking)

parley_bench/README.md
DJMcNab: I also just saw that Tango would require manual changes and nightly on Windows - we should also document that (even if it's just "Windows might require additional setup, see Tango's docs for details")

parley_dev/assets/fonts/arimo_fonts/Arimo-VariableFont_wght.ttf
taj-p: Nice catch! Thank you - fixed in https://github.com/linebender/parley/pull/405/commits/711c0111d0af971ef772c953ecf345c0d3a174e6

parley_dev/Cargo.toml
taj-p: Great idea - implemented in https://github.com/linebender/parley/pull/405/commits/711c0111d0af971ef772c953ecf345c0d3a174e6

parley_dev/src/lib.rs
taj-p: Added in https://github.com/linebender/parley/pull/405/commits/385e49cb134ab76ff81e97d1c560196e2d723331

parley_dev/assets/text_samples/README.md
taj-p: Yes, I agree with your assessment. I've made this change in https://github.com/linebender/parley/pull/405/commits/e5f2e4670739909ebbb9b597ced5fadffc565c2a

parley_bench/README.md
taj-p: Thank you! Added in https://github.com/linebender/parley/pull/405/commits/4d5a16bfa4ecef98b72184d43b2a3d3e24164f07

parley_bench/build.rs
taj-p: Yup! Added in cf13c64c0cc21bf2d0146a0b1d52e2e82fd37e85

parley_dev/assets/text_samples/README.md
```diff
@@ -0,0 +1,9 @@
+# Sample Text Attribution
+
+The sample text amendments, source, and title are as follows:
```
DJMcNab:
```suggestion
The sample texts in this folder are all in the public domain, and were obtained from the linked pages (if modified, with the documented amendments):
```

(The main reason for this suggestion is emphasising the public domain-ness)

parley_dev/assets/text_samples/README.md
```diff
@@ -0,0 +1,9 @@
...
+
+| Sample            | Title                        | Amendments    | Source                                                                                                                                                               |
+| ----------------- | ---------------------------- | ------------- | -------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| `./arabic.txt`    | Al-Kindi - First Philosophy  | None          | [Wiki Source](https://ar.wikisource.org/wiki/%D8%A7%D9%84%D9%83%D9%86%D8%AF%D9%8A_-_%D8%A7%D9%84%D9%81%D9%84%D8%B3%D9%81%D8%A9_%D8%A7%D9%84%D8%A3%D9%88%D9%84%D9%89) |
```
DJMcNab: Maximum nit: Wikisource tends to be stylised "Wikisource" instead of "Wiki Source". This is non-blocking

CHANGELOG.md
```diff
@@ -20,6 +20,7 @@ This release has an [MSRV] of 1.82.
 #### Parley
 
 - Shift-click support through `Selection::shift_click_extension` and `PlainEditorDriver::shift_click_extension`. ([#385][] by [@kekelp][])
+- Benchmarks ([#405]() by [@taj-p][])
```
DJMcNab: Uber nit:
```suggestion
- Benchmarks. ([#405]() by [@taj-p][])
```

README.md
```diff
@@ -101,11 +101,7 @@ Licensed under either of
 
 at your option.
 
-Some files used for tests are under different licenses:
-
-- The font file `Arimo-VariableFont_wght.ttf` in `/parley/tests/assets/arimo_fonts/` is licensed solely as documented in that folder (and is licensed under the Apache License, Version 2.0).
-- The font file `Roboto-Regular.ttf` in `/parley/tests/assets/roboto_fonts/` is licensed solely as documented in that folder (and is licensed under the Apache License, Version 2.0).
-- The font file `NotoKufiArabic-Regular.otf` in `/parley/tests/assets/noto_fonts/` is licensed solely as documented in that folder (and is licensed under the SIL Open Font License, Version 1.1).
+Some files used for tests and benchmarks are under different licenses. See `./parley_dev/README.md` for details.
```
DJMcNab, line 104: This still needs to be reflected in `./parley/README.md` (i.e. the "Some files used for tests are under different licenses:" section can just be removed).

taj-p: 🤦 - Thanks for this catch. Removed in https://github.com/linebender/parley/pull/405/commits/2e61fa686eca25db940883ff6058bfd289429a9e

parley_dev/assets/text_samples/README.md
taj-p: Removed in https://github.com/linebender/parley/pull/405/commits/ca1845a1e7887a6da54f0c2b4b12ebf452174d3b

PR #406, opened 2025-08-26
Cache HarfRust structs

taj-p:
Caches HarfRust `ShapeData`, `ShaperInstance`, and `ShapePlan`s using an LRU inspired by Swash's implementation.

### DIFF
--- a/parley/src/lru_cache.rs
+++ b/parley/src/lru_cache.rs
@@ -1,0 +1,184 @@
+/// A lookup key is distinct from the ID type. This allows the lookup key
+/// to not require ownership of the underlying ID data, which could require
+/// allocations.
+pub(crate) trait LookupKey<ID> {
+    fn eq(&self, other: &ID) -> bool;
+    fn to_id(self) -> ID;
+}
+
+/// An entry in the cache.
+pub(crate) struct Entry<ID, T> {
+    pub epoch: u64,
+    pub id: ID,
+    pub data: T,
+}
+
+/// A least-recently-used cache. This cache uses a linear scan of its entries
+/// to find a given entry - it is optimised for a low number of entries. Preferably
+/// keep `max_entries` low in the order of tens.
+pub(crate) struct LruCache<ID, T> {
+    entries: Vec<Entry<ID, T>>,
+    epoch: u64,
+    max_entries: usize,
+}
+
+impl<ID, T> LruCache<ID, T> {
+    pub(crate) fn new(max_entries: usize) -> Self {
+        Self {
+            entries: Default::default(),
+            epoch: 0,
+            max_entries,
+        }
+    }
+
+    /// Returns a reference to the entry with the given ID. If the entry is not
+    /// found, it is created and returned using `make_data`.
+    pub(crate) fn entry<'a>(
+        &'a mut self,
+        id: impl LookupKey<ID>,
+        make_data: impl FnOnce() -> T,
+    ) -> &'a T {
+        match self.find_entry(id, make_data) {
+            (true, index) => {
+                let entry = &mut self.entries[index];
+                entry.epoch = self.epoch;
+                &entry.data
+            }
+            (false, index) => {
+                self.epoch += 1;
+                let entry = &mut self.entries[index];
+                entry.epoch = self.epoch;
+                &entry.data
+            }
+        }
+    }
+
+    fn find_entry(
+        &mut self,
+        id: impl LookupKey<ID>,
+        make_data: impl FnOnce() -> T,
+    ) -> (bool, usize) {
+        let epoch = self.epoch;
+        let mut lowest_serial = epoch;
+        let mut lowest_index = 0;
+        for (i, entry) in self.entries.iter().enumerate() {
+            if id.eq(&entry.id) {
+                return (true, i);
+            }
+            if entry.epoch < lowest_serial {
+                lowest_serial = entry.epoch;
+                lowest_index = i;
+            }
+        }
+        if self.entries.len() < self.max_entries {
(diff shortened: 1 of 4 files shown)


parley/src/shape/mod.rs
```diff
@@ -319,12 +319,7 @@ fn shape_item<'a, B: Brush>(
         };
         let script = swash_convert::script_to_harfrust(item.script);
         let language = if let Some(lang) = item.locale {
-            let lang_tag = lang.language();
-            if let Ok(harf_lang) = lang_tag.parse::<harfrust::Language>() {
-                Some(harf_lang)
-            } else {
-                None
-            }
+            lang.language().parse::<harfrust::Language>().ok()
```
nicoburns: Suggestion:
```rust
let language = item.locale.and_then(|lang| lang.language().parse::<harfrust::Language>().ok());
```

parley/src/lru_cache.rs
```diff
@@ -0,0 +1,185 @@
...
+
+use alloc::vec::Vec;
+
+/// A lookup key is distinct from the ID type. This allows the lookup key
+/// to not require ownership of the underlying ID data, which could require
+/// allocations.
+pub(crate) trait LookupKey<ID> {
+    fn eq(&self, other: &ID) -> bool;
+    fn to_id(self) -> ID;
+}
```
nicoburns: This is fine. But I believe this is normally done with the `Borrow` and `ToOwned` traits (see `std::HashMap`) or the `Equivalent` trait (https://docs.rs/equivalent/1.0.2/equivalent/) which is a newer design used in `HashBrown`.

parley/src/shape/mod.rs
```diff
@@ -281,18 +291,67 @@ fn shape_item<'a, B: Brush>(
...
-        // TODO: cache this upstream?
-        let shaper_data = harfrust::ShaperData::new(&font_ref);
-        let instance = harfrust::ShaperInstance::from_variations(
-            &font_ref,
-            variations_iter(&font.font.synthesis, rcx.variations(item.variations)),
+        let shaper_data = scx
+            .shape_data_cache
+            .entry(cache::ShapeDataKey::new(font.font.blob.id()), || {
+                harfrust::ShaperData::new(&font_ref)
+            });
```
nicoburns: I believe this should also key on `font.font.index`, otherwise all of the fonts stored in an Open Type Collection (otc/ttc) file will (incorrectly) share ShapeData (note the index is required to construct the `FontRef`)

xorgy:
I am seeing very similar performance differences... this is definitely above the noise floor. :+ )

parley/src/shape/cache.rs
```diff
@@ -0,0 +1,128 @@
...
+// SPDX-License-Identifier: Apache-2.0 OR MIT
+
+use crate::{
+    FontVariation,
+    lru_cache::{LookupKey, LruCache},
+};
+use alloc::boxed::Box;
+
+/// A cache of `ShaperData` instances.
+pub(crate) type ShapeDataCache = LruCache<u64, harfrust::ShaperData>;
```
nicoburns: Don't feel like these type aliases are adding much. They just obscure the type? I'd consider ditching the aliases and importing types like `ShaperData` with `use` so that the full type is shorter.

parley/src/shape/cache.rs
```diff
@@ -0,0 +1,128 @@
...
+    }
+}
+
+impl LookupKey<u64> for ShapeDataKey {
+    fn eq(&self, other: &u64) -> bool {
+        self.font_id == *other
+    }
+    fn to_id(self) -> u64 {
+        self.font_id
+    }
```
nicoburns: Consider adding `#[inline(always)]` on these. Bit of a micro-optimisation, but you definitely want these inlined.

parley/src/lru_cache.rs
nicoburns: There is also https://docs.rs/litemap, although it probably won't quite work for this use cases.

nicoburns:
This is bringing the time for Blitz to construct Parley layouts (single-threaded mode) from ~51ms-52ms down to ~47ms-48ms.

parley/src/shape/mod.rs
taj-p: Implemented in https://github.com/linebender/parley/pull/406/commits/41f6ce16e657047c211f65f5d973c06215bbf2e4

parley/src/lru_cache.rs
taj-p: Oohhhh `Equivalent` worked very well. Thank you for this suggestion 🙏 . Implemented in https://github.com/linebender/parley/pull/406/commits/39597fade2d4c59933f9248c910a86ff608620d3 and https://github.com/linebender/parley/pull/406/commits/ca5dce2720b60d670101ec5d609b239c6a86db3b

parley/src/shape/mod.rs
taj-p: Yes! Nice catch - implemented in https://github.com/linebender/parley/pull/406/commits/a93fd7fd46b13d4cbc49012b41424fdc638df33c and https://github.com/linebender/parley/pull/406/commits/10bd024e5143f8de8e9c41216b9e596924956e3c

parley/src/shape/cache.rs
taj-p: Removed in https://github.com/linebender/parley/pull/406/commits/343509c9eaf1d7fca10a222d0b9dead88fb2415e

parley/src/shape/cache.rs
taj-p: I inlined `equivalent` and `into`s

PR #412, opened 2025-08-29
Cache charmaps

taj-p:
Caches charmap construction.

We should see even greater gains when https://github.com/googlefonts/fontations/pull/1647 is pulled in. I saw this function show up unnecessarily in profiles (we recompute it in our cache currently), so I'm hopeful to see more gains with its incorporation.

## Results

image

### Blitz

Before

18ms pconstruct

image

After

7ms pconstruct

image

### DIFF
--- a/parley/src/shape/cache.rs
+++ b/parley/src/shape/cache.rs
@@ -1,7 +1,7 @@
 // Copyright 2025 the Parley Authors
 // SPDX-License-Identifier: Apache-2.0 OR MIT
 
-use crate::FontVariation;
+use crate::{FontVariation, lru_cache::LruCache};
 use alloc::boxed::Box;
 use hashbrown::Equivalent;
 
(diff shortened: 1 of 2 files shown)


nicoburns:
@taj-p Could you share what you were measuring with those "pconstruct" numbers (url and whether construction was done single-threaded or multi-threaded)? I am able to measure a decent boost from this PR, but nothing quite that dramatic as a percentage.

For `just open https://en.wikipedia.org/wiki/Internet` in single-threaded mode, I'm seeing "pconstruct" times of ~43ms vs  with ~47ms `main` and ~53ms with 29bcffe572cc726f976c76a2a607c69a123164bc (pre the caching PR). 43ms is notably around what I was seeing for the swash shaper (using parley before the upgrade).

With multi-threading it seems to be closer to a 1ms difference (hard to tell due to lack of measurement precision), which seems about what should be expected given I have a 10-core machine.

https://github.com/googlefonts/fontations/pull/1647 unfortunately doesn't seem to make much difference for me.

Patched in with:
```toml
[patch.crates-io]
skrifa = { git = "https://github.com/googlefonts/fontations", rev = "07df8391a0b33f2402cf8ab78f5bf01c90c3a647" }
```

nicoburns (approved):

taj-p:
Oops @nicoburns - I got confused. There are multiple `pconstructs` and it looks like I didn't realise they aren't all the same 😅 

image

taj-p:
Could it be the 8ms one is cached?

nicoburns:
Hmmm, I was seeing this PR make things faster, even after the HarfRust 0.2 bump. But now I'm seeing it's making things slower. Someone else should measure, and I'll try again in the morning.

parley/src/shape/cache.rs
```diff
@@ -176,3 +176,27 @@ impl<'a> From<ShapePlanKey<'a>> for ShapePlanId {
         }
     }
 }
+
+/// Cache for charmap mapping indices.
+pub(crate) struct CharmapCache {
+    cache: LruCache<u64, skrifa::charmap::MappingIndex>,
```
dfrg, line 182: The key here should include the font index as well. Fonts in ttc files often share the same cmap table but not always so this could potentially lead to incorrect font selection.

dfrg:
I created #413 as a possible alternative to this

PR #413, opened 2025-08-30
Cache charmap metadata in fontique

dfrg:
And use this in parley for font selection. This is an alternative to #412.

@nicoburns care to check the performance of this one?

### DIFF
--- a/fontique/src/charmap.rs
+++ b/fontique/src/charmap.rs
@@ -1,6 +1,11 @@
 // Copyright 2024 the Parley Authors
 // SPDX-License-Identifier: Apache-2.0 OR MIT
 
+//! Mapping codepoints to nominal glyph identifiers.
+
+// TODO(dfrg): move this code to read-fonts so it can be shared among other
+// crates.
+
 use read_fonts::{
     FontData, FontRead, FontRef, TableProvider, TopLevelTable,
     tables::cmap::{Cmap, CmapSubtable},
@@ -56,8 +61,11 @@ pub struct Charmap<'a> {
 impl Charmap<'_> {
     /// Returns the glyph identifier for the given codepoint.
     pub fn map(&self, codepoint: impl Into<u32>) -> Option<u32> {
+        const ASCII_MAX: u32 = 0x7F;
         let mut c = codepoint.into();
-        if self.is_mac_roman && c > 0x7F {
+        // The Mac Roman encoding requires special processing for codepoints
+        // above the ASCII range.
+        if self.is_mac_roman && c > ASCII_MAX {
             c = unicode_to_mac_roman(c);
         }
         let result = match &self.subtable {
@@ -65,14 +73,10 @@ impl Charmap<'_> {
             CmapSubtable::Format6(table) => table.map_codepoint(c),
             CmapSubtable::Format10(table) => {
                 if let Some(index) = c.checked_sub(table.start_char_code()) {
-                    if index < table.num_chars() {
-                        table
-                            .glyph_id_array()
-                            .get(index as usize)
-                            .map(|gid| GlyphId::from(gid.get()))
-                    } else {
-                        None
-                    }
+                    table
+                        .glyph_id_array()
+                        .get(index as usize)
+                        .map(|gid| GlyphId::from(gid.get()))
                 } else {
                     None
                 }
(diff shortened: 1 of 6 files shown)


dfrg:
Running the benchmarks seems to show a surprising improvement:

Screenshot 2025-08-30 at 9 45 09 AM

nicoburns:
@dfrg This is dramatically faster. There's some variance, but for single-threaded `just open https://en.wikipedia.org/wiki/Internet` I'm seeing numbers around 30ms compared to 53ms for initial HarfRust version 45ms on `main`, and IIRC 43ms for the swash verion of Parley.

dfrg:
Thanks Nico. That seems to match the roughly 30% improvement shown in the benchmarks.

fontique/src/charmap.rs
```diff
@@ -0,0 +1,123 @@
...
+    subtable: CmapSubtable<'a>,
+    is_symbol: bool,
+    is_mac_roman: bool,
+}
+
+impl Charmap<'_> {
+    /// Returns the glyph identifier for the given codepoint.
+    pub fn map(&self, codepoint: impl Into<u32>) -> Option<u32> {
+        let mut c = codepoint.into();
+        if self.is_mac_roman && c > 0x7F {
```
nicoburns: Nit: It would be nice to avoid the magic `0x7F` constant. That didn't mean anything to me. Perhaps an `is_ascii` function? Or just a comment?

parley/src/shape/mod.rs
```diff
@@ -516,21 +516,19 @@ impl<'a, 'b, B: Brush> FontSelector<'a, 'b, B> {
...
-            let charmap = font_ref.charmap();
             let map_status = cluster.map(|ch| {
                 charmap
                     .map(ch)
                     .map(|g| {
-                        g.to_u32()
-                            .try_into()
-                            .expect("Swash requires u16 glyph, so we hope that the glyph fits")
+                        // HACK: in reality, we're only using swash to compute
+                        // coverage so the actual gid doesn't matter
```
nicoburns: This could probably do with some explanation that `0` is used to indicate that the glyph doesn't exist. But I assume we'll be replacing this code soon anyway, so I'm not going to worry too much.

I have satisfied myself that the swash function called does indeed only care whether the id is zero or non-zero

nicoburns (approved):

fontique/src/charmap.rs
```diff
@@ -0,0 +1,123 @@
...
+            c = unicode_to_mac_roman(c);
+        }
+        let result = match &self.subtable {
+            CmapSubtable::Format0(table) => table.map_codepoint(c),
+            CmapSubtable::Format6(table) => table.map_codepoint(c),
+            CmapSubtable::Format10(table) => {
+                if let Some(index) = c.checked_sub(table.start_char_code()) {
+                    if index < table.num_chars() {
```
taj-p: For my understanding, is this an intentional fast path? As in, it's cheaper to manual blinds check on `.num_chars()` than it is to rely on the bounds check on `.get` access to  `.glyph_id_array()`?

taj-p (approved):
This is far better than #412. Thank you @dfrg 🙏 .

fontique/src/charmap.rs
dfrg: Ah, good catch. This code is copied from harfrust and is something I’d like to move into read-fonts eventually. The extra check here is vestigial from a time when our cmap10 codegen was incorrect (and for some reason, it’s also missing a map_codepoint method).

I’ll fix this before merging and add a todo to get this whole chunk of code into read-fonts so it can be shared by skrifa, harfrust, fontique and others.

fontique/src/charmap.rs
dfrg: Good call. I’ll make this more clear.

parley/src/shape/mod.rs
dfrg: Yeah, hopefully this code should be gone soon and that should give us an additional performance boost. I’ll add a comment noting that we only care about zero and non-zero for the purposes of this value.

nicoburns:
@dfrg Why does this `Charmap` support different `cmap` subtables formats (0, 4, 6, 10, 12, 13) to `skrifa::charmap::Charmap` (only 4 and 12)?

PR #414, opened 2025-09-09
Remove kurbo and peniko dependencies

nicoburns:
We should probably wait until we've backported `linebender_resource_handle` to `peniko` 0.4 to land this. Then the font/blob conversion code can disappear.

We may want to name `parley::Rect` something more specific like `parley::SelectionArea`

Depends on:
- https://github.com/linebender/parley/pull/415

### DIFF
--- a/fontique/Cargo.toml
+++ b/fontique/Cargo.toml
@@ -17,7 +17,7 @@ workspace = true
 
 [features]
 default = ["system"]
-std = ["read-fonts/std", "linebender_resource_handle/std", "dep:memmap2"]
+std = ["read-fonts/std", "dep:memmap2"]
 libm = ["read-fonts/libm", "dep:core_maths"]
 icu_properties = ["dep:icu_properties"]
 unicode_script = ["dep:unicode-script"]
--- a/fontique/Cargo.toml
+++ b/fontique/Cargo.toml
@@ -17,8 +17,8 @@ workspace = true
 
 [features]
 default = ["system"]
-std = ["read-fonts/std", "peniko/std", "dep:memmap2"]
-libm = ["read-fonts/libm", "peniko/libm", "dep:core_maths"]
+std = ["read-fonts/std", "linebender_resource_handle/std", "dep:memmap2"]
+libm = ["read-fonts/libm", "dep:core_maths"]
 icu_properties = ["dep:icu_properties"]
 unicode_script = ["dep:unicode-script"]
 # Enables support for system font backends
@@ -37,7 +37,7 @@ system = [
 [dependencies]
 bytemuck = { workspace = true }
 read-fonts = { workspace = true }
-peniko = { workspace = true }
+linebender_resource_handle = { workspace = true }
 smallvec = "1.15.0"
 memmap2 = { version = "0.9.5", optional = true }
 unicode-script = { version = "0.5.7", optional = true }
--- a/parley/Cargo.toml
+++ b/parley/Cargo.toml
@@ -18,13 +18,7 @@ workspace = true
 
 [features]
 default = ["system"]
-std = [
-    "fontique/std",
-    "harfrust/std",
-    "peniko/std",
-    "skrifa/std",
-    "swash/std",
-]
+std = ["fontique/std", "harfrust/std", "peniko/std", "skrifa/std", "swash/std"]
 libm = ["fontique/libm", "peniko/libm", "skrifa/libm", "swash/libm", "dep:core_maths"]
 # Enables support for system font backends
 system = ["std", "fontique/system"]
--- a/parley/Cargo.toml
+++ b/parley/Cargo.toml
@@ -21,7 +21,6 @@ default = ["system"]
 std = [
     "fontique/std",
     "harfrust/std",
-    "linebender_resource_handle/std",
     "peniko/std",
     "skrifa/std",
     "swash/std",
--- a/parley/Cargo.toml
+++ b/parley/Cargo.toml
@@ -18,7 +18,14 @@ workspace = true
 
 [features]
 default = ["system"]
-std = ["fontique/std", "harfrust/std", "peniko/std", "skrifa/std", "swash/std"]
+std = [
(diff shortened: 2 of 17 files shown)


parley/Cargo.toml
```diff
@@ -36,4 +43,5 @@ harfrust = { workspace = true }
 
 [dev-dependencies]
 parley_dev = { workspace = true }
+peniko = { workspace = true }
```
T-256, line 39: it's only there for `peniko::kurbo::Size` in  `tests::utils::renderer::RenderingConfig`. Couldn't just create new `Size` in there? Just like what's done with `Rect`.

DJMcNab: These tests do have a full renderer in them, so FWIW I think keeping the Peniko dependency is reasonable. When/if we move these to Vello CPU, it's plausible that we would directly have a dependency through to Peniko.

parley/src/util.rs
```diff
@@ -14,3 +14,53 @@ pub(crate) fn nearly_eq(x: f32, y: f32) -> bool {
 pub(crate) fn nearly_zero(x: f32) -> bool {
     nearly_eq(x, 0.)
 }
+
+/// A rectangle.
+#[derive(Clone, Copy, Default, PartialEq)]
+pub struct Rect {
+    /// The minimum x coordinate (left edge).
+    pub x0: f64,
+    /// The minimum y coordinate (top edge in y-down spaces).
```
DJMcNab: Does Parley not already assume a y-down space? That is,

```suggestion
    /// The minimum y coordinate (top edge in Parley's y-down space).
```

parley/src/util.rs
```diff
@@ -14,3 +14,53 @@ pub(crate) fn nearly_eq(x: f32, y: f32) -> bool {
 pub(crate) fn nearly_zero(x: f32) -> bool {
     nearly_eq(x, 0.)
 }
+
+/// A rectangle.
+#[derive(Clone, Copy, Default, PartialEq)]
+pub struct Rect {
```
DJMcNab: What do you think about making this `non_exhaustive` (and removing `pub fn new`/`Default`), to gently discourage more general use. Obviously if people can get a single `Rect`, since it's `Copy`, they can easily make new ones, but it should be sufficient roadblocks to prevent people doing that.

I'd also consider renaming to something like `OutputRect` or similar. In particular, I *never* want Rust Analyzer to suggest this type when I am looking to import `Rect`.
Basically strongly saying to applications that they should standardise on a rectangle type, and it shouldn't be this one.

That would make this PR much easier for me to approve, as it makes this API much clearer.
But I'd also suggest that we can talk about this in office hours.

DJMcNab (commented):
This is a good way to achieve the goal, and I agree with the goal. I've got some thoughts about making sure that we make clear to our users and tooling what the purpose of this new type is.

parley/src/util.rs
nicoburns: I guess so. I don't really care if anybody uses it, but I do agree that it would be good to remove it from autocomplete for `Rect`. How about a rename to `BoundingBox` (with an eye on using [BoundingBox](https://docs.rs/font-types/latest/font_types/struct.BoundingBox.html) from the `font-types` crate once `font-types` hits 1.0.

Or we could even be really specific with `CursorArea`.

DJMcNab: The rename is the more important of the two, I agree; `BoundingBox` is a pretty good name here.

nicoburns:
This turns out to have a nice compile time benefit for `fontique` (which people may well wish to use without using other parts of the linebender ecosystem - e.g. https://xi.zulipchat.com/#narrow/channel/296640-font-tech/topic/Refactoring.20fontique.20backend.20into.20separate.20crate).

`peniko` itself is quite lightweight, but it depends on `color` and `kurbo` which are both small-but-not-nothing.

parley/src/util.rs
nicoburns: I have made these comments specific to Parley's coordinate space seeing as this type is not intended to be used elsewhere.

parley/src/util.rs
nicoburns: Renamed to `BoundingBox`

fontique/Cargo.toml
```diff
@@ -17,8 +17,8 @@ workspace = true
 
 [features]
 default = ["system"]
-std = ["read-fonts/std", "peniko/std", "dep:memmap2"]
-libm = ["read-fonts/libm", "peniko/libm", "dep:core_maths"]
+std = ["read-fonts/std", "linebender_resource_handle/std", "dep:memmap2"]
```
DJMcNab:
```suggestion
std = ["read-fonts/std", "dep:memmap2"]
```

As the `std` feature on Linebender Resource Handle doesn't actually do anything at the moment.

parley/src/util.rs
```diff
@@ -14,3 +14,53 @@ pub(crate) fn nearly_eq(x: f32, y: f32) -> bool {
...
+#[derive(Clone, Copy, Default, PartialEq)]
+pub struct BoundingBox {
+    /// The left edge
+    pub x0: f64,
+    /// The top edge
+    pub y0: f64,
+    /// The right edge
+    pub x1: f64,
+    /// The bottom edge
+    pub y1: f64,
```
DJMcNab:
```suggestion
    /// The left edge.
    pub x0: f64,
    /// The top edge.
    pub y0: f64,
    /// The right edge.
    pub x1: f64,
    /// The bottom edge.
    pub y1: f64,
```

parley/Cargo.toml
```diff
@@ -18,7 +18,14 @@ workspace = true
 
 [features]
 default = ["system"]
-std = ["fontique/std", "harfrust/std", "peniko/std", "skrifa/std", "swash/std"]
+std = [
+    "fontique/std",
+    "harfrust/std",
+    "linebender_resource_handle/std",
```
DJMcNab:
```suggestion
```

PR #415, opened 2025-09-15
Fix `clippy::obfuscated_if_else` lints

nicoburns:
Another round of clippy lint fixes for new rustc. Not sure that I love these changes (/this lint), but they're not egregious enough for me to want to fight against them.

### DIFF
--- a/fontique/src/font.rs
+++ b/fontique/src/font.rs
@@ -430,14 +430,16 @@ fn read_attributes(font: &FontRef<'_>) -> (FontWidth, FontStyle, FontWeight) {
 
     fn from_head(head: Head<'_>) -> (FontWidth, FontStyle, FontWeight) {
         let mac_style = head.mac_style();
-        let style = mac_style
-            .contains(MacStyle::ITALIC)
-            .then_some(FontStyle::Italic)
-            .unwrap_or_default();
-        let weight = mac_style
-            .contains(MacStyle::BOLD)
-            .then_some(700.0)
-            .unwrap_or_default();
+        let style = if mac_style.contains(MacStyle::ITALIC) {
+            FontStyle::Italic
+        } else {
+            FontStyle::default()
+        };
+        let weight = if mac_style.contains(MacStyle::BOLD) {
+            700.0
+        } else {
+            0.0
+        };
         (FontWidth::default(), style, FontWeight::new(weight))
     }
 
--- a/parley/src/layout/cursor.rs
+++ b/parley/src/layout/cursor.rs
@@ -1050,14 +1050,16 @@ enum AnchorBase {
 }
 
 fn cursor_rect<B: Brush>(cluster: &Cluster<'_, B>, at_end: bool, size: f32) -> Rect {
-    let line_x = (cluster.visual_offset().unwrap_or_default()
-        + at_end.then(|| cluster.advance()).unwrap_or_default()) as f64;
+    let mut line_x = cluster.visual_offset().unwrap_or_default();
+    if at_end {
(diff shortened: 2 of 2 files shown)


fontique/src/font.rs
```diff
@@ -430,14 +430,16 @@ fn read_attributes(font: &FontRef<'_>) -> (FontWidth, FontStyle, FontWeight) {
...
-            .then_some(FontStyle::Italic)
-            .unwrap_or_default();
-        let weight = mac_style
-            .contains(MacStyle::BOLD)
-            .then_some(700.0)
-            .unwrap_or_default();
+        let style = if mac_style.contains(MacStyle::ITALIC) {
+            FontStyle::Italic
+        } else {
+            Default::default()
```
DJMcNab:
```suggestion
            FontStyle::default()
```

I've proposed making this required in https://github.com/linebender/linebender.github.io/pull/122, but in this context it's more readable. Note that the `FontWidth::default` is already below, so this is also the prevailing style.

fontique/src/font.rs
```diff
@@ -430,14 +430,16 @@ fn read_attributes(font: &FontRef<'_>) -> (FontWidth, FontStyle, FontWeight) {
...
+        };
+        let weight = if mac_style.contains(MacStyle::BOLD) {
+            700.0
+        } else {
+            Default::default()
```
DJMcNab:
```suggestion
            f32::default()
```
Which then pretty immediately falls through to

```suggestion
            0.0
```
Which doesn't seem right as a weight...
I'd suggest these branches should use `FontWeight::BOLD` and `FontWeight::NORMAL`, but I'm not sure.
I realise that would be a change of semantics, and so feel free to defer this (but if you do so and my understanding that this is suspect is correct, move this comment out into its own issue).

Either way, this lint has helped discover this.

parley/src/layout/cursor.rs
```diff
@@ -1051,7 +1051,11 @@ enum AnchorBase {
 
 fn cursor_rect<B: Brush>(cluster: &Cluster<'_, B>, at_end: bool, size: f32) -> Rect {
     let line_x = (cluster.visual_offset().unwrap_or_default()
-        + at_end.then(|| cluster.advance()).unwrap_or_default()) as f64;
+        + if at_end {
+            cluster.advance()
+        } else {
+            Default::default()
+        }) as f64;
```
DJMcNab: Less clear on the code style here; I don't think either the old proposal or the new one is particularly good, although I'm not sure that the new one is better...

I'm really not sure what is best here; I think I'm happy for this change to land as-is, but I'd be keen to try an alternative if you can think of one.

DJMcNab (approved):
Thanks. Approving iff the suggestion to use `T::default` instead of `Default::default` is applied.

It isn't clear why this blocks #414

parley/src/layout/cursor.rs
nicoburns:
```rust
let mut line_x = cluster.visual_offset().unwrap_or_default();
if at_end {
     line_x += cluster.advance() as f64;
}
```

would be the obvious alternative. Normally it's better to make everything immutable where possible, but maybe there's an exception here.

fontique/src/font.rs
nicoburns: Agree. That looks like a bug to me. I think I'll fix it in a separate PR though so it isn't sneaked into a PR that looks like a form-only change.

parley/src/layout/cursor.rs
nicoburns: I've gone with this.

PR #416, opened 2025-09-15
[Fontique] Fix: fallback to `FontWeight::default()` rather than `FontWeight(0.0)`

nicoburns:
Noticed in https://github.com/linebender/parley/pull/415#discussion_r2349395111

Fontique currently uses `FontWeight(0.0)` (`f32::default()`) when it should probably be using `FontWeight(400.0)` (`FontWeight::default()`) as the fallback weight when the font is not bold.


DJMcNab:
I don't have the context to meaningfully review this, unfortunately. It certainly seems wrong.

I guess ideally we'd have some relevant tests here...

dfrg (approved):
Yes, this was broken but the fallback to the head table probably never happens so nobody noticed. Fix LGTM

PR #424, opened 2025-10-01
updated peniko to 0.5

rydb:
Updates peniko to 0.5. Uses specific vello commit for example while the new crate with peniko 0.5v hasn't been pushed yet.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4,9 +4,9 @@ version = 3
 
 [[package]]
 name = "ab_glyph"
-version = "0.2.29"
+version = "0.2.32"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "ec3672c180e71eeaaac3a541fbbc5f5ad4def8b747c595ad30d674e43049f7b0"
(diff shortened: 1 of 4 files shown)


nicoburns:
The solution to this may be to move the `vello_editor` example out of the workspace and make a standlone crate. Or we can just bump MSRV for Parley.

Given that this is just for an example, it probably makes sense to wait until vello 0.6 has been published to land this.

DJMcNab:
https://github.com/linebender/parley/releases/tag/v0.6.0 is out now, so we can make this update happen, I think.

Bumping MSRV to match Vello should be fine.

rydb:
> > vello requires rust edition 2024
> 
> Wait, what do you mean by that? You mean Vello has a higher MSRV?

yes, vello's MSRV is 1.86.

PoignardAzur:
You don't need to change the edition then, only to raise the MSRV of `vello_editor`. No need to move it out of the workspace.

nicoburns:
Let's handle this as part of a Vello 0.7 upgrade once that's released.

PR #426, opened 2025-10-02
Prepare for 0.6.0 release

nicoburns:
Bump version numbers for 0.6.0

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,14 +8,14 @@ Subheadings to categorize changes are `added, changed, deprecated, removed, fixe
 
 # Changelog
 
-The latest published Parley release is [0.6.0](#060---2025-10-03) which was released on 2025-10-03.
-You can find its changes [documented below](#060---2025-10-03).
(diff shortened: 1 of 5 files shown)


DJMcNab:
I've not been carefully keeping up with Parley, so I'm not best placed to evaluate whether we need a release, if the changelog is up to scratch, etc.

Before we land this, I'd like to see Masonry and Xilem brought up to the latest main here. Based on the changelog, I believe that should be trivial, but if not, then the changelog needs to be updated.

(There's also a lot of malformed formatting in the changelog, incidentally, which should be fixed)

taj-p (approved):

nicoburns:
@DJMcNab Yes please - this is blocking a Blitz 0.2 release. I've updated the date in the changelog, but feel free to make any additional changes you need.

nicoburns:
> But I want sign-off from someone else that this is sensible

I don't really get why `cargo` decides that makes Parley unpublishable in the first place, but I am confident that removing the version specifier of an unpublished dev-dependency is harmless.

PR #428, opened 2025-10-03
Rename `Parley::Font` to `FontData`

DJMcNab:
This matches the name other crates export this as.

Perhaps there was a reason this name was a different name in this crate?


DJMcNab:
Actually, let's make it non-breaking with a deprecation.

nicoburns (dismissed):
Deprecated alias also seems fine to me, although we didn't bother doing that with `peniko`!

DJMcNab:
Sorry... GitHub doesn't let us avoid TOCTOU on dismissing reviews, apparently.

nicoburns (approved):

PR #430, opened 2025-10-03
Remove mentions of peniko from the README

nicoburns:
Parley no longer depends on `peniko`. I do not think `linebender_resource_handle` is signficant enough to worth mentioning so I've just removed the section on `peniko` without replacement.

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,18 +13,10 @@
 </div>
 
 Parley provides an API for implementing rich text layout.
-It is backed by [Swash](https://github.com/dfrg/swash).
 
 ## The Parley text stack
 
-Currently, Parley directly depends on five crates: Fontique, HarfRust, Swash, Skrifa, and Peniko.
-These crates cover different pieces of the text-rendering process.
-
-### Peniko
-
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -17,15 +17,9 @@ It is backed by [Swash](https://github.com/dfrg/swash).
 
 ## The Parley text stack
 
-Currently, Parley directly depends on five crates: Fontique, HarfRust, Swash, Skrifa, and Peniko.
+Currently, Parley directly depends on four crates: Fontique, HarfRust, Skrifa, and Swash.
```
DJMcNab: Maybe we should say "four main crates" or "four key crates" or something.

I don't want to be asserting something which isn't true.

This doesn't block this PR, as e.g. we also depend on hashbrown.

DJMcNab (approved):
LGTM

README.md
nicoburns: I've reworded this to "key dependencies". There's probably still room for improvement (and we could do with a more detailed description of what Parley itself does), but I'll leave that as a follow-up.

PR #435, opened 2025-10-19
make PlainEditor::cursor_at() and ::set_selection() public fns

narodnik:
I need to use these fns in my app to manually set the selection range in the editor.


narodnik:
I just noticed `.select_byte_range()` which works. Its annoying I have to make the driver which requires the font and layout context whereas this code doesnt but ok.

PR #437, opened 2025-10-24
Add maximum screenshot size and optimize existing screenshots

PoignardAzur:
Make tests refuse to save screenshots above a given size.

Add oxipng dev dependency.
Optimize all existing screenshots with oxipng.
Add `max_screenshot_size` setting for the rare screenshot that needs more space.

### DIFF
--- a/parley/src/tests/utils/env.rs
+++ b/parley/src/tests/utils/env.rs
@@ -370,7 +370,7 @@ impl TestEnv {
         // We special-case android targets because oxipng doesn't build in Android CI.
         #[cfg(target_os = "android")]
         fn save_image(_image: &Pixmap, _path: &PathBuf, _max_size: Option<usize>) {
-            panic!("Saving screenshots is not supported on wasm targets");
+            panic!("Saving screenshots is not supported on Android targets");
         }
 
         // Max size a screenshot can have, in bytes.


parley/src/tests/utils/env.rs
```diff
@@ -63,6 +64,7 @@ pub(crate) struct TestEnv {
     // TODO: Add core::panic::Location for case.
     errors: Vec<(PathBuf, String)>,
     next_test_case_name: String,
+    max_screenshot_size: Option<usize>,
```
DJMcNab, line 67: We should probably mention that this is the file size in bytes in the places it's mentioned.

parley/src/tests/utils/env.rs
```diff
@@ -336,15 +343,37 @@ impl TestEnv {
...
+            // but it's simpler to cover all cases.
+            std::fs::write(path, data).unwrap();
+
+            if let Some(max_size) = max_size {
+                if saved_len > max_size {
+                    panic!(
+                        "New screenshot file ({saved_len} bytes) was larger than the supported file size ({max_size} bytes).",
+                    );
+                }
+            }
```
DJMcNab, line 367: I think we need to make sure that this check also applies to the existing file (when we didn't have this in Vello, we had a case where someone ignored the failure and manually renamed the file and submitted the test anyway). If I'm understanding this correctly, this check in this form currently only happens if the images didn't match in size, right?

DJMcNab (changes requested):
Looks good to me, but I do have a blocking concern about when we perform this check.

parley/src/tests/utils/env.rs
PoignardAzur: It only applies to new images, yes.

You think we should check existing images? We currently don't to that in Xilem AFAIK.

DJMcNab: I believe that was the intention of this branch - https://github.com/linebender/xilem/blob/4edf461d4fc18e0d0e8365c6c3cdc08a74131c5d/masonry_testing/src/harness.rs#L1093-L1099

If that doesn't actually test the screenshot size of existing files in Masonry, that's a bug.

PoignardAzur: Okay, I'll add that.

PoignardAzur:
Man, oxipng is a fiddly dependency. I wish there was a "Rust-only version" that was guaranteed to never ever call clang at any point.

parley/src/tests/utils/env.rs
```diff
@@ -336,15 +343,56 @@ impl TestEnv {
...
+                        "New screenshot file size ({saved_len}b) was larger than the supported file size ({max_size}b). See TestEnv::max_screenshot_size to change the maximum.",
+                    );
+                }
+            }
+        }
+
+        // We special-case android targets because oxipng doesn't build in Android CI.
+        #[cfg(target_os = "android")]
+        fn save_image(_image: &Pixmap, _path: &PathBuf, _max_size: Option<usize>) {
+            panic!("Saving screenshots is not supported on wasm targets");
```
DJMcNab:
```suggestion
            panic!("Saving screenshots is not supported on Android targets");
```

PR #439, opened 2025-10-27
Document precondition for `RangedStyleBuilder::push_default`

DJMcNab:
PR mostly exists to test new branch protection, but is useful on its own.

As a note for review, this code is entirely `pub(crate)`.

### DIFF
--- a/parley/src/resolve/range.rs
+++ b/parley/src/resolve/range.rs
@@ -22,7 +22,10 @@ impl<B: Brush> Default for RangedStyleBuilder<B> {
         Self {
             properties: vec![],
             root_style: ResolvedStyle::default(),
-            len: !0,
+            // We use `usize::MAX` as a sentinel that `begin` hasn't been called.
+            // This is required (rather than requiring the root style in the constructor)
+            // as we want to support re-using this value.
+            len: usize::MAX,
         }
     }
 }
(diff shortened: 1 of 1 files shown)


parley/src/resolve/range.rs
```diff
@@ -38,14 +38,28 @@ impl<B: Brush> RangedStyleBuilder<B> {
...
 
     /// Change a property of the root style, which covers the full range of text.
+    ///
+    /// # Panics
+    ///
+    /// If [`begin`](Self::begin) has not been called before using this method.
     pub(crate) fn push_default(&mut self, property: ResolvedProperty<B>) {
-        assert!(self.len != !0);
+        assert!(
+            self.len != !0,
```
PoignardAzur: Wait, `self.len != !0`? Does this mean `usize::MAX`?

DJMcNab: Yep! That's exactly why I documented this stuff, because I found the condition extremely confusing

DJMcNab: Do you want me to change the tombstone to explicitly use `usize::MAX` in this PR?

PoignardAzur: The more I look at this code, the more questions I have.

But in this case, given that RangedStyleBuilder is (kind of) a singleton, there's no need to optimize its size. An Option would be better than a sentinel value. If we really want a sentinel value, then yeah, MAX is better.

DJMcNab: Continues in [#parley > RangedStyleBuilder API](https://xi.zulipchat.com/#narrow/channel/205635-parley/topic/RangedStyleBuilder.20API/with/547470958)

PR #440, opened 2025-10-28
Split off various modules into "editing" folder.

PoignardAzur:
Beside code moves, the only changes in this PR are:
- Changed imports.
- Some items going from private to pub(crate).

### DIFF
--- a/parley/src/layout/mod.rs
+++ b/parley/src/layout/mod.rs
@@ -35,7 +35,7 @@ pub use line::{GlyphRun, LineMetrics, PositionedInlineBox, PositionedLayoutItem}
 pub use run::RunMetrics;
 
 // TODO - Deprecation not yet active to ease internal code migration.
-#[deprecated(since="TBD", note="Access from the `editing` module instead.")]
+#[deprecated(since = "TBD", note = "Access from the `editing` module instead.")]
 pub use crate::editing::{Cursor, Selection};
 
 /// Alignment of a layout.


PoignardAzur:
I'm trying to redo the work from https://github.com/linebender/parley/pull/237 in smaller chunks. This was the part people were most enthusiastic about.

parley/src/layout/mod.rs
```diff
@@ -31,13 +28,15 @@ use swash::text::cluster::Boundary;
...
 pub use cluster::{Affinity, ClusterPath, ClusterSide};
-pub use cursor::{Cursor, Selection};
 pub use data::BreakReason;
 pub(crate) use line::LineItem;
 pub use line::greedy::BreakLines;
 pub use line::{GlyphRun, LineMetrics, PositionedInlineBox, PositionedLayoutItem};
 pub use run::RunMetrics;
 
+// TODO - Remove
+pub use crate::editing::{Cursor, Selection};
```
DJMcNab:
```suggestion
// TODO - Deprecation not yet active to ease internal code migration.
#[deprecated(since="TBD", note="Access from the `editing` module instead.")]
pub use crate::editing::{Cursor, Selection};
```

If you use `TBD`, it won't error yet, but it will make the future migration easier.

parley/src/lib.rs
```diff
@@ -114,6 +114,7 @@ mod shape;
 mod swash_convert;
 mod util;
 
+pub mod editing;
```
DJMcNab, line 117: There are arguments for this to be something like `edit` here, but editing is also fine.

PR #443, opened 2025-10-29
Improve performance of rendering many paragraphs of text

taj-p:
## Intent

To use a running calculation of line height rather than a separate iteration after breaking. 

This sees not much change to benchmarks except for longer latin pieces. 

image

The motivation for this change occurs for laying out many paragraphs of latin text as shown in the before and after profiles below (which measures laying out the entire latin sample text). 

### Before

Time to render 10,000 latin samples: 109s

image

Breaking dominates the performance profile spending most of its time calculating line height.

image

## After

Time to render 10,000 latin samples: 33s

Breaking is no longer the node in the profile.

image

image

image

### DIFF
--- a/parley/src/layout/line/greedy.rs
+++ b/parley/src/layout/line/greedy.rs
@@ -306,7 +306,7 @@ impl<'a, B: Brush> BreakLines<'a, B> {
                                 self.state.add_line_height(cluster_height(
                                     &self.layout.data,
                                     cluster.data,
-                                    &run.data,
+                                    run.data,
                                 ));
                                 return self.start_new_line();
                             }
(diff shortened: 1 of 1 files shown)


nicoburns:
This will need to change with `vertical-align` implementation:

- A separate `ascent` and `descent` height above/below the alignment point would need to be tracked for each node
- Some nodes are alignment relative to the line box and would need to be positioned in a second pass once the first-pass line-height is computed. Those can then further alter the line height.

nicoburns (approved):
As it turns out, I actually need the running line height for float layout (use case: abandon line layout in the current position if the running line height exceeds the available height). So I think this is probably worth doing even aside from performance.

PR #444, opened 2025-10-29
Split contents of `layout/mod.rs`  file

PoignardAzur:
This splits type declarations into their respective files, to make the code easier to search.

This PR has no functional changes, only moves and import changes.

### DIFF
--- a/parley/src/layout/mod.rs
+++ b/parley/src/layout/mod.rs
@@ -12,7 +12,11 @@ mod line;
 mod line_break;
 mod run;
 
-#[expect(clippy::module_inception, reason = "TODO - rename parent module")]
+// TODO - Add to allowed lint set?
+#[expect(
+    clippy::module_inception,
+    reason = "Private inner module for code organisation"
+)]
 mod layout;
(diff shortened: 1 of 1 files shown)


parley/src/layout/mod.rs
```diff
@@ -3,285 +3,42 @@
...
 mod alignment;
 mod cluster;
+mod glyph;
 mod line;
+mod line_break;
 mod run;
 
-pub(crate) mod data;
+#[expect(clippy::module_inception, reason = "TODO - rename parent module")]
+mod layout;
```
nicoburns, line 20: IMO no reason needed, and I would expect us to leave parent module as-is

DJMcNab: I'd be in favour of adding this as an allowed lint to the lint set - see https://github.com/rust-lang/rust-clippy/issues/13259. That would certainly match our other philosophy here.

For now for this PR, I agree that this TODO is not good; I'd use a reason like `"Private inner module for code organisation."`

PR #446, opened 2025-10-30
Remove superfluous CoreFloat imports

PoignardAzur:


tomcur:
Changelog claims these were moved into `core` in 1.84: https://releases.rs/docs/1.84.0/.

PoignardAzur:
Alright, I'm changing the status to draft until we either bump the MSRV or find a better way to gate this warning.

Issue #450, opened 2025-11-06
AtomicU64 is not available on all Rust targets


nikp123:
Hi, sorry if this seems out of scope for this project, but I've ran into this issue while trying to compile Slint for an embedded platform running on ARMv6 (Raspberry Pi Zero W, arm-unknown-linux-gnueabi)

That platform offers no Atomic 64-bit operations and therefore doesn't include that functionality in std at all, causing the build to fail. This is based off of preliminary research done here: https://users.rust-lang.org/t/unresolved-import-core-atomicu64/109568/7

There are 4 spots in the entire repository where 64-bit Atomics were used and I'm not sure how important they are: https://github.com/search?q=repo%3Alinebender%2Fparley%20AtomicU64&type=code

If you can somehow fix this in order to prevent future builds from breaking that'd be great.

DJMcNab:
This also has the potential to be closely related to #107.

PR #451, opened 2025-11-10
Fontique: fix build on platforms without 64bit atomics

nicoburns:
- Fixes https://github.com/linebender/parley/issues/450

### DIFF
--- a/fontique/src/family.rs
+++ b/fontique/src/family.rs
@@ -24,6 +24,7 @@ impl FamilyId {
     pub fn new() -> Self {
         /// Make sure this is larger than the largest generic family id.
         static ID_COUNTER: AtomicCounter = AtomicCounter::new(64);
+        #[allow(clippy::useless_conversion, reason = "Not useless on 32-bit platforms")]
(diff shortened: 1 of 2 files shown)


fontique/src/family.rs
```diff
@@ -22,8 +23,9 @@ impl FamilyId {
     #[allow(clippy::new_without_default)]
     pub fn new() -> Self {
         /// Make sure this is larger than the largest generic family id.
-        static ID_COUNTER: AtomicU64 = AtomicU64::new(64);
-        Self(ID_COUNTER.fetch_add(1, Ordering::Relaxed))
+        static ID_COUNTER: AtomicCounter = AtomicCounter::new(64);
+        #[allow(clippy::useless_conversion, reason = "Not useless on 32-bit platforms")]
+        Self(ID_COUNTER.fetch_add(1, Ordering::Relaxed).into())
```
DJMcNab, line 28: This `fetch_add` being potential 32 bit introduces a new overflow point, and so needs an overflow comment (same in the other instance of this). This can be the same as in Linebender Resource Handle.

PR #458, opened 2025-11-12
`attributed_text`: switch to structured `Error` + `ErrorKind` with bo…

waywardmonkeys:
…undary details

- Replace `ApplyAttributeError` with `Error { kind: ErrorKind, start, end, len, boundary }`
- `AttributedText::apply_attribute` now returns Result<(), Error>
- Add `Endpoint` and `BoundaryInfo` (reports offending endpoint/index and enclosing codepoint span)
- Enforce UTF-8 boundaries via `TextStorage::is_char_boundary` and surface detailed context
- Enrich `Display` to include range, len, and boundary info
- Re-export `Error`, `ErrorKind`, `BoundaryInfo`, `Endpoint` at crate root
- Update tests to assert kinds and key details in messages

### DIFF
--- a/attributed_text/src/attributed_text.rs
+++ b/attributed_text/src/attributed_text.rs
@@ -5,37 +5,7 @@ use alloc::vec::Vec;
 use core::fmt::Debug;
 use core::ops::Range;
 
-use crate::TextStorage;
-
-/// The errors that might happen as a result of [applying] an attribute.
-///
-/// [applying]: AttributedText::apply_attribute
-#[derive(Copy, Clone, Debug, PartialEq)]
-#[non_exhaustive]
-pub enum ApplyAttributeError {
-    /// The bounds given were invalid.
-    ///
-    /// TODO: Store some data about this here.
-    InvalidBounds,
-    /// The range has `start > end`.
-    InvalidRange,
-    /// Either `start` or `end` is not on a UTF-8 character boundary.
-    NotOnCharBoundary,
-}
-
-impl core::fmt::Display for ApplyAttributeError {
-    fn fmt(&self, f: &mut core::fmt::Formatter<'_>) -> core::fmt::Result {
-        match self {
-            Self::InvalidBounds => f.write_str("attribute range is out of bounds"),
-            Self::InvalidRange => f.write_str("attribute range start is greater than end"),
-            Self::NotOnCharBoundary => {
-                f.write_str("attribute range must align to UTF-8 char boundaries")
-            }
-        }
-    }
-}
-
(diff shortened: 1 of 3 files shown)


attributed_text/src/error.rs
```diff
@@ -0,0 +1,219 @@
...
+    if text.is_char_boundary(index) {
+        return Some((index, index));
+    }
+
+    // Previous boundary (max 3 bytes back)
+    let mut s = index;
+    for _ in 0..4 {
+        if s == 0 {
+            break;
+        }
```
tomcur: In principle this should be impossible, as `text.is_char_boundary(0)` should always be `true`.

```suggestion
        // We can never wrap around `0` here, as when `s == 0` before the decrement,
        // the previous call to `text.is_char_boundary` (either in this loop or above)
        // will have returned `true`.
        debug_assert!(s != 0, "`s` should never wrap around 0").
```

attributed_text/src/attributed_text.rs
```diff
@@ -143,32 +127,76 @@ mod tests {
...
-            Err(ApplyAttributeError::InvalidBounds)
-        );
+        assert!(at.apply_attribute(0..3, TestAttribute::Keep).is_ok());
+        assert!(at.apply_attribute(0..6, TestAttribute::Keep).is_ok());
+        match at.apply_attribute(4..3, TestAttribute::Keep) {
+            Err(e) => {
+                assert_eq!(e.kind(), ErrorKind::InvalidRange);
+                let msg = format!("{}", e);
+                assert!(msg.contains("4..3"));
+                assert!(msg.contains("start > end") || msg.contains("invalid range"));
```
tomcur: Do we expect to always keep saying both that the range is invalid, and why it's invalid?

```suggestion
                assert!(msg.contains("start > end"));
                assert!(msg.contains("invalid range"));
```

attributed_text/src/attributed_text.rs
```diff
@@ -143,32 +127,76 @@ mod tests {
...
+                assert!(msg.contains("0..7"));
+                assert!(msg.contains("len 6"));
+            }
+            _ => panic!("expected InvalidBounds"),
+        }
+        match at.apply_attribute(7..8, TestAttribute::Keep) {
+            Err(e) => {
+                assert_eq!(e.kind(), ErrorKind::InvalidBounds);
+                let msg = format!("{}", e);
+                assert!(msg.contains("7..8") || msg.contains("8"));
```
tomcur: Similar to above, this perhaps should be `&&`, but at the same time, as written the second expression is implied by the first.

PR #459, opened 2025-11-12
Remove the `select_*` methods from `PlainEditorDriver`

ickshonpe:
`PlainEditorDriver` has about a dozen `select_*` functions that have matching `move_*` functions that are identical except that they call `set_selection` (or whatever) with a `true` value for its `extend` parameter. When handling keyboard inputs you check the state of the modifier keys and call the appropriate function:
```rust
                    Key::Named(NamedKey::ArrowLeft) => {
                        if action_mod {
                            if shift {
                                drv.select_word_left();
                            } else {
                                drv.move_word_left();
                            }
                        } else if shift {
                            drv.select_left();
                        } else {
                            drv.move_left();
                        }
                    }
```

This PR removes `select_*` functions and add an `extend` parameter to its matching `move_*` functions instead, which allows the keyboard handling logic to be simplified like so:
```rust
                    Key::Named(NamedKey::ArrowLeft) => {
                        if action_mod {
                            drv.move_word_left(shift);
                        } else {
                            drv.move_left(shift);
                        }
                    }
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,23 @@ You can find its changes [documented below](#060---2025-10-06).
 
 This release has an [MSRV] of 1.82.
 
+### Migration 
+
+The `move_*` methods on `PlainEditorDrv` now have a `bool` parameter, pass `false` for their previous behavior. The `select_*` selection extending methods on `PlainEditorDrv` have been removed, instead use the corresponding `move_*` methods passing `true`.
+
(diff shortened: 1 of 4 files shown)


nicoburns (changes requested):
This change makes a lot of sense to me. However:

- We should document what the `extend` parameter does on each of the functions that now has one.
- We will need a changelog entry

PR #460, opened 2025-11-13
Introduce `parley_flow`

waywardmonkeys:
This introduces a small, flow‑first layer on top of Parley that treats a document as an ordered set of text blocks, with explicit container rectangles defining where each block lives. The flow owns geometry and the “join” between neighboring blocks (e.g., space vs newline), while blocks only supply layout and text. On top of this, the crate provides hit‑testing, selection geometry, and cross‑block text extraction, so you can drag through multiple paragraphs/labels and get predictable results.

Motivation

- Borrow from proven platform models (e.g., TextKit containers): separate shaping/layout from flow/regions, keep APIs small and durable.
- Reduce ambiguity and API surface: one explicit flow path for hit‑testing, selection, and copy, instead of guessing based on per‑layout state.
- Keep integration simple: a convenience builder stacks blocks vertically when you don’t want to hand author rects; otherwise, supply rectangles you already have in your UI.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2784,6 +2784,13 @@ name = "parley_dev"
 version = "0.0.0"
 
 [[package]]
+name = "parley_flow"
+version = "0.0.0"
+dependencies = [
+ "parley",
+]
+
+[[package]]
 name = "paste"
 version = "1.0.15"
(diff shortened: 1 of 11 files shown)


waywardmonkeys:
I don’t think I will do this sort of work here and no one else has shown interest.

Issue #464, opened 2025-11-16
Confusing behaviour when libfontconfig is missing


brynedwards:
Hello,

No text appears when running a program using parley without libfontconfig in LD_LIBRARY_PATH. For example, here is how the hello-masonry example from Xilem appears:
<img width="344" height="345" alt="Image" src="https://github.com/user-attachments/assets/1596fcb3-c45c-46b4-a3df-bbe5ab041e11" />

I did not know why there was no text until I did a git bisect to find it started with #378. I realise that this might only affect unconventional setups (in my case I use [Nix](nixos.org) for development environments and my shell.nix did not include libfontconfig) but some kind of indication of this behaviour being due to libfontconfig missing would be very helpful. Thank you.

PR #467, opened 2025-11-20
fontique: Make the yeslogic-fontconfig-sys/dlopen feature optional

ogoffart:
The problem is that yeslogic-fontconfig-sys is actually not following the cargo best practice to have the feature as additive, and the dlopen feature changes the API of the crate.
Thus, if you have in your dependency tree crates that uses yeslogic-fontconfig-sys with the dlopen feature and some that do not, it breaks the crates that depends on it without enabling the feature.

This makes the dlopen an optional feature of fontique. It is disabled by default because the feature set of fontique without fontconfig doesn't make much sense.

Linking directly against the fontcinfig library makes the dependency explicit and allow tool to figure out extra dependency easily.

This should therefore also solve issue #464

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -119,6 +119,10 @@ jobs:
           targets: x86_64-unknown-none,thumbv8m.main-none-eabihf
           components: clippy
 
+      - name: Install fontconfig-dev
+        if: runner.os == 'Linux'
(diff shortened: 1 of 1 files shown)


nicoburns (approved):
Ah, this looks like the intended way to do this (provided by `yeslogic-fontconfig-sys`). Perhaps we get a similar PR merged into Servo too.

PR #471, opened 2025-11-24
Fontique: rename `dlopen` feature to `fontconfig-dlopen`

nicoburns:
The name `dlopen` is quite generic. This feature applies specifically to the `fontconfig` dependency.

No release has yet been made with this feature flag, so this change is non-breaking.

### DIFF
--- a/fontique/src/backend/fontconfig.rs
+++ b/fontique/src/backend/fontconfig.rs
@@ -202,7 +202,7 @@ impl Clone for Pattern {
                 LIB,
                 FcPatternReference,
                 self.inner.as_ptr()
-            )
+            );
         };
         Self { inner: self.inner }
     }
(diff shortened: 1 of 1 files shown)


fontique/Cargo.toml
```diff
@@ -33,8 +33,11 @@ system = [
     "dep:yeslogic-fontconfig-sys",
     "dep:roxmltree",
 ]
-# Use dlopen to load the fontconfig library.
-dlopen = ["yeslogic-fontconfig-sys?/dlopen"]
+
+# Use dlopen to load the fontconfig library. This allows Fontique to compile even if the
+# system does not have fontconfig installed, although the font fallback to system fonts
```
DJMcNab, line 38:
```suggestion
# Use dlopen to load the fontconfig library. This allows Fontique to run even if the end-user
# (Unix) system does not have fontconfig installed, although the font fallback to system fonts
```

PR #473, opened 2025-12-03
Bake composite properties data

robertbastian:
There is no need for ICU4X's data provider infrastructure, we can directly bake the `CodePointTrie<u32>` and access that. This gets around the limitation of not being able to bake custom data markers, and removes the serde serialization.

#452


robertbastian:
Yes you should use PackTab if you can get better performance out of that. For ICU4X in general that's more complicated, because all PackTab results would need to be the same Rust type (which is not a requirement for your singleton). We could experiment with baking function pointers as part of the data, which might optimize as well.

PR #474, opened 2025-12-05
unicode_data_gen: Create `composite` dir

conor-93:
`cargo run -p unicode_data_gen <output-path>` was failing when the `composite` dir did not exist:

```
thread 'main' panicked at unicode_data_gen/src/lib.rs:106:88:
called `Result::unwrap()` on an `Err` value: Os { code: 2, kind: NotFound, message: "No such file or directory" }
```


conor-93:
> After handling the directory already existing, LGTM

Thanks!

PR #475, opened 2025-12-09
Use ICU4X `Script` type in `fontique`

robertbastian:
ICU4X has two script types: a string-based one (`icu::locale::subtag::Script`) and an enum one (`icu::properties::props::Script`). This replaces `fontique`'s string-based one by `icu`'s string-based one, a like-for-like replacement that shouldn't have any performance impact.

### DIFF
--- a/fontique/src/backend/android.rs
+++ b/fontique/src/backend/android.rs
@@ -143,11 +143,9 @@ impl SystemFonts {
                                                             script_fallback
                                                                 .push((script!("Hani"), *family));
                                                         }
-                                                        locale_fallback.push((
-                                                            locale.to_string().into(),
-                                                            *family,
-                                                        ));
                                                     }
+                                                    locale_fallback
+                                                        .push((locale.to_string().into(), *family));
                                                 }
                                             }
                                         }


fontique/src/script.rs
```diff
@@ -3,278 +3,256 @@
...
     }
-}
 
-impl From<&str> for Script {
-    fn from(value: &str) -> Self {
-        Self(value.as_bytes().try_into().unwrap_or_default())
+    #[cfg(feature = "icu_properties")]
+    fn from_properties_script(value: icu_properties::props::Script) -> Self {
+        value.into()
     }
```
taj-p, line 58: Do we still need this method given that consumers can just call `.into()`?

fontique/src/script.rs
```diff
@@ -3,278 +3,256 @@
...
-}
 
-impl From<[u8; 4]> for Script {
-    fn from(value: [u8; 4]) -> Self {
-        Self(value)
+    /// Returns the associated [`icu_properties::props::Script`] value.
+    #[cfg(feature = "icu_properties")]
+    fn properties_script(self) -> icu_properties::props::Script {
+        self.into()
     }
```
taj-p, line 46: Do we still need this method if consumers can `.into()`?

fontique/src/backend/android.rs
```diff
@@ -123,26 +124,30 @@ impl SystemFonts {
...
                                                             // This works around ambiguous han char­
                                                             // acters going unmapped with current
                                                             // fallback code. This should be done in
                                                             // a locale-dependent manner, since that
                                                             // is the norm.
                                                             script_fallback
-                                                                .push(("Hani".into(), *family));
+                                                                .push((script!("Hani"), *family));
                                                         }
+                                                        locale_fallback.push((
```
taj-p: I think this behavioural change (so that only locales with explicit scripts like "zh-Hans" or "zh-Hant" are added to locale_fallback) is correct

robertbastian: oh that's unintended, I can move it back

fontique/src/script.rs
robertbastian: probably not. although if we remove that method, we also should remove the `icu_properties` dependency. and then the `.into()` might not work, because it requires the `compiled_data` feature...

PR #477, opened 2025-12-09
Move baked data into `parley`

robertbastian:
Having a semver boundary in what is essentially an implementation detail might complicate improvements in the future. ICU4X baked data in particular does not have stability guarantees across even minor versions. Moving the data to the `parley` crate also simplifies some code.


robertbastian:
Yes that is a way to avoid these issues, but you have to be careful with this. I don't see the *advantage* of having the data in a separate crate.

PR #478, opened 2025-12-09
Fix ICU4X data

robertbastian:
* Clean up old deps
* Use `.with_segmenter_models` to remove segmentation models, instead of setting the language to `en`
* Set ICU4X dependencies to `~` as baked data is only compatible within minor versions


taj-p (approved):
Thank you 🙏

PR #479, opened 2025-12-09
Create Parley Draw Crate

taj-p:
## Parley Draw 

Parley Draw provides APIs for efficiently rendering glyphs and paint styles like underline.

### Goals

Parley Draw is under rapid development. Consider it experimental for now. Its goals are to:

- Provide an API surface that accepts glyphs and their positions and renders them to a surface.
- Cache those glyphs so that repeated renders of a glyph are fast.
- Support rendering paint styles like underline, strikethrough, and brush color.
- Share expensive structs and data between the shaper and renderer like the hinting instance and hinted advance.

---

## Plan

At a high level, the implementation plan for the first phase of Parley Draw is to support Vello CPU's existing text capabilities _without caching_ - simply a copy paste from Vello to Parley Draw. We will use Vello CPU until the renderer abstraction is ready and then onboard to that (if it makes sense).

Thereafter, I am interested in investigating hint aware shaping, so I will then expose a hinting cache between Parley (for use by HarfRust) and Parley Draw (for actually rendering).

### Steps

1. Copy paste glyph rendering logic from Vello to Parley Draw ([i.e. this file](https://github.com/linebender/vello/blob/main/sparse_strips/vello_common/src/glyph.rs)) to support rendering glyphs with Vello CPU.
2. Create hinting cache shareable with Parley Draw and Parley
3. Caching glyphs, supporting paint styles, and achieving our other goals will require short term API changes (and outside the scope of this initial implementation)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3335,6 +3335,10 @@ name = "parley_dev"
 version = "0.0.0"
 
 [[package]]
+name = "parley_draw"
+version = "0.0.0"
+
+[[package]]
 name = "parse-zoneinfo"
 version = "0.5.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -7,6 +7,7 @@ members = [
     "parley_bench",
     "parley_core",
     "parley_dev",
+    "parley_draw",
     "unicode_data",
     "unicode_data_gen",
     "examples/tiny_skia_render",
--- a/parley_draw/Cargo.toml
+++ b/parley_draw/Cargo.toml
@@ -1,0 +1,24 @@
+[package]
+name = "parley_draw"
+version = "0.0.0"
+description = "Parley Draw provides APIs for efficiently rendering text."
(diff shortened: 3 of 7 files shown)


nicoburns (approved):
Being as this is just a placeholder, it seems uncontroversial - approved.

That being said, a few notes:

- I feel moderately strongly that this should not be "parley" branded. This will end up being the replacement for Swash's glyph rasterization capbabilities, which are widely used in the Rust ecosystem, including by many projects that do not use Parley. In general, this will not be coupled to Parley (I assume). And I feel like it would be clearer and easier to communicate if we reserve the "parley" branding for layout/shaping/bidi type concerns (where I am considering hinting and rasterization to be out of that scope).
- I would like to point you towards https://github.com/emilk/egui/pull/7694, which implements quite a bit of what I imagine might end up in this crate.
- Regarding abstraction: I wonder if it might make sense for Parley Draw to define it's own abstraction (similarly to how Skrifa does for outlines) that could then be adapted to to the other abstractions of which we now seem to have at least 3 (AnyRender, Vello API, and Understory).

PR #480, opened 2025-12-09
Rename `unicode_data` to `parley_data`

robertbastian:
Please don't publish a crate called `unicode_data` if it's specific to your project.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -6,12 +6,12 @@ members = [
     "parley",
     "parley_bench",
     "parley_core",
-    "parley_dev",
-    "parley_draw",
     "parley_data",
     "parley_data_gen",
-    "examples/tiny_skia_render",
+    "parley_dev",
+    "parley_draw",
     "examples/swash_render",
+    "examples/tiny_skia_render",
     "examples/vello_editor",
     "xtask",
 ]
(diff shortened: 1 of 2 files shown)


parley_data_gen/Cargo.toml
```diff
@@ -13,7 +13,7 @@ icu_locale = { workspace = true }
 icu_properties = { workspace = true, features = ["compiled_data"] }
 icu_collections = { workspace = true, features = ["databake"] }
 icu_provider_source = { workspace = true, features = ["networking", "use_wasm"] }
-unicode_data = { workspace = true, default-features = false, features = ["datagen"] }
+parley_data = { workspace = true, default-features = false, features = ["datagen"] }
```
taj-p: Can you please fix the ordering? Unfortunately these dependencies are not alphabetical or otherwise ordered, but grouping the `icu_*` deps together helps improve the ability for one to scan the deps

Cargo.toml
```diff
@@ -8,8 +8,8 @@ members = [
     "parley_core",
     "parley_dev",
     "parley_draw",
-    "unicode_data",
-    "unicode_data_gen",
+    "parley_data",
+    "parley_data_gen",
```
taj-p, line 10: Please move `parley_data*` to before `parley_dev` so that members at this level are alphabetically ordered and match the file tree

PR #481, opened 2025-12-09
[Parley Draw]: Add Vello Copy

taj-p:
## Copying Vello's Text Logic to Parley Draw

To kick off implementation of Parley Draw, we're firstly copying over the Vello implementation.

***EVERYTHING WILL CHANGE SIGNIFICANTLY*** (especially APIs!). For now, we're simply copying the code across for the purposes of using it as a foundation to build off and to allow us to start exploring hint aware shaping.

This approach has been OK'd by @dfrg . 

### Changes made to the copy

- Added a `vello_cpu_render` example that uses Parley Draw so that we can keep track of how the API evolves. After we're happy with Parley Draw, I recommend applying it to our other examples.
- Feature gated `vello_cpu`'s implementation of `GlyphRender` and `ColrRender` so that consumers aren't required to use it. However, we still depend on `vello_common`.
- We have to pain to Vello's github repository because `Pixmap` hasn't yet been moved from `vello_cpu` to `vello_common` in the latest 0.0.4 release. 
- We pass the `GlyphCaches` into `GlyphRunBuilder` instead of having the trait own them. This reflects how we will eventually support sharing the hinting instance cache between the shaper and renderer. But, of course, this is all subject to change.
- Updated MSRV to 1.88 to use Vello (which has an MSRV of 1.88).

### Immediate Follow Ups

- I decided not to port the tests over at the same time. I suspect we might want to discuss the best way to execute tests for Parley Draw, so I think that's best discussed in a follow up after this is merged. Right now, I'm thinking of moving the existing `parley` tests out to `parley_tests` and using Parley Draw for rasterising those tests (alongside additional draw only tests like COLR emoji rasterisation tests).

### Example

This is the output of the new example:

vello_cpu_render

## Reviews

Since most of this code has already been reviewed in Vello, I suggest we focus the review on the changes listed above.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -31,7 +31,10 @@ bytemuck = { version = "1.24.0", default-features = false }
 databake = { version = "0.2", default-features = false }
 fontique = { version = "0.7.0", default-features = false, path = "fontique" }
 harfrust = { version = "0.3.2", default-features = false }
-hashbrown = { version = "0.16.1", default-features = false, features = ["default-hasher", "raw-entry"] }
+hashbrown = { version = "0.16.1", default-features = false, features = [
+    "default-hasher",
+    "raw-entry",
+] }
 icu_codepointtrie_builder = { version = "0.5.1", default-features = false, features = ["wasm"] }
 icu_collections = { version = "2.1.1", default-features = false }
 icu_locale = { version = "2.1.1", default-features = false }


parley_draw/src/colr.rs
```diff
@@ -0,0 +1,403 @@
+// Copyright 2025 the Parley Authors
```
xStrom:
```suggestion
// Copyright 2025 the Vello Authors and the Parley Authors
```

parley_draw/src/glyph.rs
```diff
@@ -0,0 +1,1020 @@
+// Copyright 2025 the Parley Authors
```
xStrom:
```suggestion
// Copyright 2025 the Vello Authors and the Parley Authors
```

parley_draw/src/math.rs
```diff
@@ -0,0 +1,29 @@
+// Copyright 2025 the Parley Authors
```
xStrom:
```suggestion
// Copyright 2025 the Vello Authors and the Parley Authors
```

parley_draw/src/renderers/vello_cpu.rs
```diff
@@ -0,0 +1,174 @@
+// Copyright 2025 the Parley Authors
```
xStrom:
```suggestion
// Copyright 2025 the Vello Authors and the Parley Authors
```

PR #490, opened 2025-12-17
Fix bidi state leaking across layouts

### LINKED ISSUE #489: Bidi state is leaking between layouts
If you render RTL text and then render LTR text, the LTR text glyphs are in reverse order. Presumably, bidi state is leaking between layouts.

I.e., this test fails:

```rs
#[test]
fn glyph_order_rtl_then_ltr() {

    // First, render some RTL text
    {
        let text = "مرحبا"; // Arabic "Hello"
        let builder = env.ranged_builder(text);
        layout.break_all_lines(Some(200.0));
        layout.align(None, Alignment::Start, AlignmentOptions::default());

        assert!(layout.is_rtl());
    }

    // Now render LTR text - this shouldn't be affected by the previous RTL layout
    {
        let text = "ABC";
        let builder = env.ranged_builder(text);
        layout.break_all_lines(Some(200.0));
        layout.align(None, Alignment::Start, AlignmentOptions::default());

        assert!(!layout.is_rtl());

        let line = layout.lines().next().unwrap();
        let item = line.items().next().unwrap();
        let glyph_run = match item {
            crate::PositionedLayoutItem::GlyphRun(glyph_run) => glyph_run,
            crate::PositionedLayoutItem::InlineBox(_) => unreachable!(),
        };

        assert!(!glyph_run.run().is_rtl());

        // For LTR text, positioned glyphs should have increasing x coordinates
        let positions: Vec<f32> = glyph_run.positioned_glyphs().map(|g| g.x).collect();
        for i in 1..positions.len() {
            assert!(
                positions[i] > positions[i - 1],
                "LTR positioned glyphs should have increasing x coordinates. Got: {:?}",
                positions
            );
        }
    }
}
```

taj-p:
Fixes #489 

Also consolidates regression tests to a separate file.

### DIFF
--- a/parley/src/tests/test_issues.rs
+++ b/parley/src/tests/test_issues.rs
@@ -5,7 +5,7 @@ use super::utils::TestEnv;
 use crate::{Alignment, AlignmentOptions, test_name};
 
 /// Test that rendering RTL text doesn't affect subsequent LTR layouts.
-/// See https://github.com/linebender/parley/issues/489.
+/// See <https://github.com/linebender/parley/issues/489>.
 #[test]
 fn issue_489() {
     let mut env = TestEnv::new(test_name!(), None);
@@ -53,7 +53,7 @@ fn issue_489() {
 }
 
 /// Test that justified text is correctly aligned.
-/// See https://github.com/linebender/parley/issues/409.
+/// See <https://github.com/linebender/parley/issues/409>.
 #[test]
 fn issue_409() {
     let mut env = TestEnv::new(test_name!(), None);


parley/src/context.rs
```diff
@@ -151,6 +151,7 @@ impl<B: Brush> LayoutContext<B> {
         self.styles.clear();
         self.inline_boxes.clear();
         self.info.clear();
+        self.bidi.clear();
```
taj-p, line 154: This is how we fixed the bug.

PR #493, opened 2025-12-22
Fix crash for empty layout

nicoburns:
With latest Parley `main`, Parley is crashing in the empty layout case (`text = ""`).  There is code to add a dummy space (not ideal seeing as this actually gets added to text inputs - but that was a pre-existing issue), but Parley now expects there to be at least one style.

Without this fix, Parley crashes on the `panic!("No style info");` below.

An alternative fix (that I think would also work) would be to ensure that the `TreeBuilder` always pushes at least one style span (even in the empty case). Let me know if you prefer that and want me to try it.

### DIFF
--- a/parley/src/tests/test_builders.rs
+++ b/parley/src/tests/test_builders.rs
@@ -80,6 +80,7 @@ fn assert_builders_produce_same_result<'a, 'b>(
     root_style: &'a TextStyle<'b, ColorBrush>,
     with_ranged_builder: impl Fn(&mut RangedBuilder<'_, ColorBrush>),
     with_tree_builder: impl Fn(&mut TreeBuilder<'_, ColorBrush>),
+    expect_empty: bool,
 ) {
     let mut fcx = create_font_context();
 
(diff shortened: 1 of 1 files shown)


parley/src/analysis/mod.rs
```diff
@@ -325,7 +325,15 @@ pub(crate) fn analyze_text<B: Brush>(lcx: &mut LayoutContext<B>, text: &str) {
         }
     }
 
-    let text = if text.is_empty() { " " } else { text };
+    if text.is_empty() {
+        text = " ";
+        if lcx.styles.is_empty() {
+            lcx.styles.push(RangedStyle {
+                style: ResolvedStyle::default(),
+                range: 0..0,
```
conor-93, line 333: It shouldn't matter if the crash is fixed, but should this be `0..1`, as the text is one character long? Up to you if you want to change this.

parley/src/analysis/cluster.rs
```diff
@@ -257,20 +257,15 @@ impl CharCluster {
     }
 }
 
-#[derive(Copy, Clone, Debug, PartialEq, Eq)]
+#[derive(Copy, Clone, Debug, Default, PartialEq, Eq)]
 #[allow(clippy::upper_case_acronyms)]
 enum FormKind {
+    #[default]
```
conor-93, line 263: 👍

parley/src/analysis/mod.rs
nicoburns: I actually originally had it as `0..1`, but that was still crashing. I think we might not be changing `text` to be longer early enough for `1` to be a valid index.

Ideally I would really like to remove the "push a dummy space in the case an empty layout" hack entirely (and have the analysis/shaping/layout pipeline properly handle the empty case). It causes a lot of problems with form inputs where the extra space can end up in the submitted value unless you manually delete it after inputing another value!

PR #494, opened 2025-12-25
Add StyleProperty::map_brush

macmv:
This adds `StyleProperty::map_brush`, which is used to convert the brush type in a `StyleProperty`.

I ran into this while dealing with multiple types of brushes, and figured it might be useful.


macmv:
good idea point, yeah i'll update it.

PR #495, opened 2025-12-29
Add `text_style` + `styled_text` crates (CSS-like resolution + document blocks)

waywardmonkeys:
- Add `text_style`: a `no_std` + `alloc` style vocabulary for text (inline + paragraph declarations with `Specified` semantics).
- Add `text_style_resolve`: engine layer for specified→computed resolution, including OpenType settings parsing and structured errors.
- Add `styled_text`: span application and a lightweight block/document model built on `attributed_text` + `text_style`, yielding resolved run iterators.
- Add `styled_text_parley`: Parley backend that lowers resolved runs into `parley::StyleProperty` calls to build `parley::Layout`.

### DIFF
--- a/examples/vello_cpu_render_styled_text/src/main.rs
+++ b/examples/vello_cpu_render_styled_text/src/main.rs
@@ -1,0 +1,253 @@
+// Copyright 2025 the Parley Authors
+// SPDX-License-Identifier: Apache-2.0 OR MIT
+
+//! Renders styled text using `styled_text` + `styled_text_parley`, then paints glyph outlines using
+//! Vello CPU through Parley Draw.
+//!
+//! Note: Emoji rendering is not currently implemented in this example. See the swash example if
+//! you need emoji rendering.
+
+#![expect(clippy::cast_possible_truncation, reason = "Deferred")]
+
+use core::ops::Range;
+
+use parley::{
+    Alignment, AlignmentOptions, FontContext, GlyphRun, Layout, LayoutContext, PositionedLayoutItem,
+};
+use parley_draw::{GlyphCaches, GlyphRunBuilder};
+use styled_text::{
+    ComputedInlineStyle, ComputedParagraphStyle, FontFeature, FontFeatures, FontSize,
+    FontVariation, FontVariations, InlineStyle, StyledText, Tag,
+};
+use styled_text_parley::build_layout_from_styled_text;
+use vello_cpu::{RenderContext, kurbo, peniko::Color};
+
+#[derive(Clone, Copy, Debug, PartialEq)]
+struct ColorBrush {
+    color: Color,
+}
+
+impl Default for ColorBrush {
+    fn default() -> Self {
+        Self {
+            color: Color::BLACK,
+        }
+    }
+}
+
+fn find_range(text: &str, needle: &str) -> Range<usize> {
+    let start = text
+        .find(needle)
+        .unwrap_or_else(|| panic!("missing substring {needle:?}"));
+    start..start + needle.len()
+}
+
+fn main() {
+    let text = concat!(
+        "StyledText + Parley\n",
+        "A small, no_std + alloc rich-text layer.\n",
+        "\n",
+        "BIG, tiny, underline, strike, and OpenType settings.\n",
+        "Some bidirectional sample: English العربية.\n"
+    )
+    .to_string();
+
+    let display_scale = 1.0;
+    let quantize = true;
+    let max_advance = Some(520.0 * display_scale);
+
+    let foreground_color = Color::BLACK;
+    let background_color = Color::from_rgb8(250, 250, 252);
+    let padding: u16 = 24;
+
+    let base_inline = ComputedInlineStyle::default().with_font_size_px(18.0);
+    let base_paragraph = ComputedParagraphStyle::default();
+    let mut styled = StyledText::new(text.as_str(), base_inline, base_paragraph);
+
+    let styled_text_style = InlineStyle::new()
+        .with_font_size(FontSize::Px(42.0))
+        .with_underline(true);
+    let parley_style = InlineStyle::new()
+        .with_font_size(FontSize::Px(30.0))
+        .with_strikethrough(true);
+    let big_style = InlineStyle::new().with_font_size(FontSize::Em(2.0));
+    let tiny_style = InlineStyle::new().with_font_size(FontSize::Px(12.0));
+    let underline_style = InlineStyle::new().with_underline(true);
+    let strike_style = InlineStyle::new().with_strikethrough(true);
+    let opentype_features_style = InlineStyle::new().with_font_features(FontFeatures::list(vec![
+        FontFeature::new(Tag::new(b"liga"), 0),
+        FontFeature::new(Tag::new(b"kern"), 0),
+    ]));
+    let variation_style =
+        InlineStyle::new().with_font_variations(FontVariations::list(vec![FontVariation::new(
+            Tag::new(b"wght"),
+            750.0,
+        )]));
+
+    styled.apply_span(
+        styled.range(find_range(&text, "StyledText")).unwrap(),
+        styled_text_style,
+    );
+
+    styled.apply_span(
+        styled.range(find_range(&text, "Parley")).unwrap(),
+        parley_style,
+    );
+
+    styled.apply_span(styled.range(find_range(&text, "BIG")).unwrap(), big_style);
+
+    styled.apply_span(styled.range(find_range(&text, "tiny")).unwrap(), tiny_style);
+
+    styled.apply_span(
+        styled.range(find_range(&text, "underline")).unwrap(),
+        underline_style,
+    );
+
+    styled.apply_span(
+        styled.range(find_range(&text, "strike")).unwrap(),
+        strike_style,
+    );
+
+    styled.apply_span(
+        styled
+            .range(find_range(&text, "OpenType settings"))
+            .unwrap(),
+        opentype_features_style,
+    );
+
+    styled.apply_span(
+        styled.range(find_range(&text, "rich-text")).unwrap(),
+        variation_style,
+    );
+
+    let mut font_cx = FontContext::new();
+    let mut layout_cx = LayoutContext::new();
+
+    let foreground_brush = ColorBrush {
+        color: foreground_color,
+    };
+    let mut layout: Layout<ColorBrush> = build_layout_from_styled_text(
+        &mut layout_cx,
+        &mut font_cx,
+        &styled,
+        display_scale,
+        quantize,
+        foreground_brush,
+    );
+
+    layout.break_all_lines(max_advance);
+    layout.align(max_advance, Alignment::Start, AlignmentOptions::default());
+
+    let width = layout.width().ceil() as u16;
+    let height = layout.height().ceil() as u16;
+    let padded_width = width + padding * 2;
+    let padded_height = height + padding * 2;
+
+    let mut renderer = RenderContext::new(padded_width, padded_height);
+    let mut glyph_caches = GlyphCaches::new();
+
+    renderer.set_paint(background_color);
+    renderer.fill_rect(&kurbo::Rect::new(
+        0.0,
+        0.0,
+        padded_width as f64,
+        padded_height as f64,
+    ));
+    renderer.set_transform(kurbo::Affine::translate(kurbo::Vec2::new(
+        padding as f64,
+        padding as f64,
+    )));
+
+    for line in layout.lines() {
+        for item in line.items() {
+            match item {
+                PositionedLayoutItem::GlyphRun(glyph_run) => {
+                    render_glyph_run(&mut renderer, &mut glyph_caches, &glyph_run);
+                }
+                PositionedLayoutItem::InlineBox(inline_box) => {
+                    renderer.set_paint(foreground_color);
+                    let (x0, y0) = (inline_box.x as f64, inline_box.y as f64);
+                    let (x1, y1) = (x0 + inline_box.width as f64, y0 + inline_box.height as f64);
+                    renderer.fill_rect(&kurbo::Rect::new(x0, y0, x1, y1));
+                }
+            }
+        }
+    }
+
+    let mut pixmap = vello_cpu::Pixmap::new(padded_width, padded_height);
+    renderer.render_to_pixmap(&mut pixmap);
+    glyph_caches.maintain();
+
+    let output_path = {
+        let path = std::path::PathBuf::from(file!());
+        let mut path = std::fs::canonicalize(path).unwrap();
+        path.pop();
+        path.pop();
+        path.pop();
+        path.push("_output");
+        drop(std::fs::create_dir(path.clone()));
+        path.push("vello_cpu_render_styled_text.png");
+        path
+    };
+    let png = pixmap.into_png().unwrap();
+    std::fs::write(output_path, png).unwrap();
+}
+
+fn render_glyph_run(
+    renderer: &mut RenderContext,
+    glyph_caches: &mut GlyphCaches,
+    glyph_run: &GlyphRun<'_, ColorBrush>,
+) {
+    renderer.set_paint(glyph_run.style().brush.color);
+    let run = glyph_run.run();
+    GlyphRunBuilder::new(run.font().clone(), *renderer.transform(), renderer)
+        .font_size(run.font_size())
+        .hint(true)
+        .normalized_coords(run.normalized_coords())
+        .fill_glyphs(
+            glyph_run
+                .positioned_glyphs()
+                .map(|glyph| parley_draw::Glyph {
+                    id: glyph.id,
+                    x: glyph.x,
+                    y: glyph.y,
+                }),
+            glyph_caches,
+        );
+
+    let style = glyph_run.style();
+    if let Some(decoration) = &style.underline {
+        let offset = decoration.offset.unwrap_or(run.metrics().underline_offset);
+        let size = decoration.size.unwrap_or(run.metrics().underline_size);
+        render_decoration(renderer, &decoration.brush, glyph_run, offset, size);
+    }
+    if let Some(decoration) = &style.strikethrough {
+        let offset = decoration
+            .offset
+            .unwrap_or(run.metrics().strikethrough_offset);
+        let size = decoration.size.unwrap_or(run.metrics().strikethrough_size);
+        render_decoration(renderer, &decoration.brush, glyph_run, offset, size);
+    }
+}
+
+fn render_decoration(
+    renderer: &mut RenderContext,
+    brush: &ColorBrush,
+    glyph_run: &GlyphRun<'_, ColorBrush>,
+    offset: f32,
+    size: f32,
+) {
+    renderer.set_paint(brush.color);
+
+    let run = glyph_run.run();
+    let x0 = glyph_run.offset();
+    let x1 = x0 + run.advance();
+    let y = glyph_run.baseline() - offset;
+
+    renderer.fill_rect(&kurbo::Rect::new(
+        x0 as f64,
+        (y - size * 0.5) as f64,
+        x1 as f64,
+        (y + size * 0.5) as f64,
+    ));
+}
--- a/styled_text/src/document.rs
+++ b/styled_text/src/document.rs
@@ -1,0 +1,115 @@
+// Copyright 2025 the Parley Authors
+// SPDX-License-Identifier: Apache-2.0 OR MIT
+
+use alloc::vec::Vec;
+use core::fmt::Debug;
+
+use crate::{ComputedInlineStyle, ComputedParagraphStyle};
+use attributed_text::TextStorage;
+
+use crate::block::Block;
+
+/// A flat sequence of semantic blocks.
+///
+/// This is intended to carry semantic structure (headings, lists, etc.) suitable for accessibility,
+/// even before layout provides geometry.
+#[derive(Debug)]
+pub struct StyledDocument<T: Debug + TextStorage, A: Debug> {
+    root_inline: ComputedInlineStyle,
+    root_paragraph: ComputedParagraphStyle,
+    blocks: Vec<Block<T, A>>,
+}
+
+impl<T: Debug + TextStorage, A: Debug> Default for StyledDocument<T, A> {
+    fn default() -> Self {
+        Self::new()
+    }
+}
+
+impl<T: Debug + TextStorage, A: Debug> StyledDocument<T, A> {
+    /// Creates an empty document.
+    #[inline]
+    pub fn new() -> Self {
+        Self {
+            root_inline: ComputedInlineStyle::default(),
+            root_paragraph: ComputedParagraphStyle::default(),
+            blocks: Vec::new(),
+        }
+    }
+
+    /// Creates an empty document with explicit root styles.
+    ///
+    /// Root styles are used for root-relative units such as `rem`.
+    #[inline]
+    pub fn new_with_root(
+        root_inline: ComputedInlineStyle,
+        root_paragraph: ComputedParagraphStyle,
+    ) -> Self {
+        Self {
+            root_inline,
+            root_paragraph,
+            blocks: Vec::new(),
+        }
+    }
+
+    /// Sets the root styles for this document, updating all existing blocks.
+    pub fn set_root_styles(
+        &mut self,
+        root_inline: ComputedInlineStyle,
+        root_paragraph: ComputedParagraphStyle,
+    ) {
+        self.root_inline = root_inline;
+        self.root_paragraph = root_paragraph;
+        for block in &mut self.blocks {
+            block
+                .text
+                .set_root_styles(self.root_inline.clone(), self.root_paragraph.clone());
+        }
+    }
+
+    /// Returns the root inline style.
+    #[inline]
+    pub fn root_inline_style(&self) -> &ComputedInlineStyle {
+        &self.root_inline
+    }
+
+    /// Returns the root paragraph style.
+    #[inline]
+    pub fn root_paragraph_style(&self) -> &ComputedParagraphStyle {
+        &self.root_paragraph
+    }
+
+    /// Appends a block to the document.
+    ///
+    /// The document's current root styles are applied to the block.
+    pub fn push(&mut self, mut block: Block<T, A>) {
+        block
+            .text
+            .set_root_styles(self.root_inline.clone(), self.root_paragraph.clone());
+        self.blocks.push(block);
+    }
+
+    /// Appends a block to the document without modifying its root styles.
+    #[inline]
+    pub fn push_raw(&mut self, block: Block<T, A>) {
+        self.blocks.push(block);
+    }
+
+    /// Returns an iterator over blocks.
+    #[inline]
+    pub fn iter(&self) -> impl Iterator<Item = &Block<T, A>> {
+        self.blocks.iter()
+    }
+
+    /// Returns the number of blocks.
+    #[inline]
+    pub fn len(&self) -> usize {
+        self.blocks.len()
+    }
+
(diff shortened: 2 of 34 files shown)


waywardmonkeys:
There's at least one big problem with this ... `text-style` is already a taken name, so `text_style` can't be the name we use.

This is a great bike shedding opportunity for people (the name).

waywardmonkeys:
I'll be pushing a proposal to address some of this shortly... and then I'll be passing out. :)

waywardmonkeys:
(It would still need some of the code that we have for some of these types in the existing stuff ... like parse functions, display impls, and bytemuck stuff. But that's easy to bring over.)

examples/tiny_skia_render_styled_text/src/main.rs
```diff
@@ -0,0 +1,342 @@
...
+            r_styled_text,
+            InlineStyle::new()
+                .font_size(Specified::Value(FontSize::Px(42.0)))
+                .underline(Specified::Value(true)),
+        )
+        .unwrap();
+
+    styled
+        .apply_span(
+            find_range(&text, "Parley"),
```
taj-p: This isn't how a consumer would use this - it's also incorrect if the needle can be found twice in the search string. Would it be possible to explicitly specify the range instead of performing the O(n) search for each span?

dfrg:
> If so, I could separate that out from this PR

Yes, this looks much better to me and I think it makes sense to split it out into a separate PR so we can bikeshed a bit on name and get it merged.

Thanks!

waywardmonkeys:
Splitting a very small `font_stack` from `parley` would be useful to me as well.

dfrg:
> Splitting a very small `font_stack` from `parley` would be useful to me as well.

Adding this to `text_primitives` (with the name `FontFamily` to match CSS) makes sense to me. We just need to figure out how to handle names in a way that makes allocation optional.

waywardmonkeys:
And the extraction of `FontFamily` (with improvements to parsing along the same lines as #501) is now in #502.

styled_text_parley/src/lib.rs
```diff
@@ -0,0 +1,195 @@
...
+        StyleProperty::LineHeight(to_parley_line_height(run.line_height()))
+    );
+    push_if!(
+        run.word_spacing_px() != default.word_spacing_px(),
+        StyleProperty::WordSpacing(run.word_spacing_px())
+    );
+
+    if run.letter_spacing_px() != default.letter_spacing_px() {
+        builder.push(StyleProperty::LetterSpacing(run.letter_spacing_px()), range);
+    }
```
taj-p:
```suggestion
    push_if!(
        run.letter_spacing_px() != default.letter_spacing_px(),
        StyleProperty::LetterSpacing(run.letter_spacing_px())
    );
```

nit: It's fine to clone 2 `usize`

styled_text_parley/src/lib.rs
```diff
@@ -0,0 +1,195 @@
...
+    builder.build(text)
+}
+
+fn push_inline_defaults<B: Brush + Clone>(
+    builder: &mut RangedBuilder<'_, B>,
+    style: &ComputedInlineStyle,
+    brush: B,
+) {
+    builder.push_default(StyleProperty::Brush(brush));
+    builder.push_default(StyleProperty::FontFamily(style.font_family().clone()));
```
taj-p: Do we have ideas about how we can remove this allocation in future?

taj-p, line 112: If a consumer has an owned `String` for a font family name (e.g., from user input, config files, or runtime construction), the current API forces multiple allocations even when reusing the same `FontFamily` across runs.

Consider a consumer with:

```rs
let my_font: String = "User's Custom Font".to_string();
```

When they need to pass the font to `InlineStyle`, they need to allocate:

```rs
        let font_family = FontFamily::Single(FontFamilyName::Named(Cow::Owned(
            dynamic_font_name.clone(), //  ALLOCATION
       )));
```

Then, they pass it to `InlineStyle`:


```rs
        let style = InlineStyle::new().font_family(Specified::Value(font_family));
```

Then, it's cloned again in `engine.rs` during resolution:

https://github.com/waywardmonkeys/parley/blob/9f033f9d0a0d7c8868a1dd672bcb7db6dbbe7e38/styled_text/src/resolve/engine.rs#L68

Then, it's cloned again when lowering to Parley:

https://github.com/waywardmonkeys/parley/blob/9f033f9d0a0d7c8868a1dd672bcb7db6dbbe7e38/styled_text_parley/src/lib.rs#L112

---

Although to fully remove the allocations, we may require a change in Parley, I think we should opt for an API here that does not allocate by design (and pay any allocation costs at the styled_text <> parley interface until such time that we can address this in Parley).

---

We could accepting `Arc`. I note how `ComputedInlineStyle` uses `Arc` for `FontVariation` and `FontSetting` but not `FontFamily`.

<details>
<summary> AI assisted reproduction </summary>

```rs
// Copyright 2025 the Parley Authors

#![allow(unsafe_code)]

//! Minimal reproduction demonstrating allocation overhead for dynamic font families.
//!
//! This module shows that consumers with dynamically-created font family strings
//! cannot reuse their allocation across frames — the current API forces multiple
//! String clones per frame.
//!
//! Run with: `cargo test -p vello_cpu_render_styled_text --lib`

use std::alloc::{GlobalAlloc, Layout, System};
use std::borrow::Cow;
use std::sync::atomic::{AtomicUsize, Ordering};

use parley::{FontContext, LayoutContext};
use styled_text::{
    ComputedInlineStyle, ComputedParagraphStyle, FontFamily, FontFamilyName, InlineStyle,
    Specified, StyledText,
};
use styled_text_parley::build_layout_from_styled_text;

// --------------------------------------------------------------------------
// Allocation counting allocator
// --------------------------------------------------------------------------

#[global_allocator]
static ALLOCATOR: CountingAllocator = CountingAllocator;

static ALLOCATION_COUNT: AtomicUsize = AtomicUsize::new(0);
static ALLOCATION_BYTES: AtomicUsize = AtomicUsize::new(0);

struct CountingAllocator;

unsafe impl GlobalAlloc for CountingAllocator {
    unsafe fn alloc(&self, layout: Layout) -> *mut u8 {
        ALLOCATION_COUNT.fetch_add(1, Ordering::SeqCst);
        ALLOCATION_BYTES.fetch_add(layout.size(), Ordering::SeqCst);
        // SAFETY: Delegating to the system allocator.
        unsafe { System.alloc(layout) }
    }

    unsafe fn dealloc(&self, ptr: *mut u8, layout: Layout) {
        // SAFETY: Delegating to the system allocator.
        unsafe { System.dealloc(ptr, layout) }
    }

    unsafe fn realloc(&self, ptr: *mut u8, layout: Layout, new_size: usize) -> *mut u8 {
        ALLOCATION_COUNT.fetch_add(1, Ordering::SeqCst);
        ALLOCATION_BYTES.fetch_add(new_size, Ordering::SeqCst);
        // SAFETY: Delegating to the system allocator.
        unsafe { System.realloc(ptr, layout, new_size) }
    }
}

fn reset_counts() {
    ALLOCATION_COUNT.store(0, Ordering::SeqCst);
    ALLOCATION_BYTES.store(0, Ordering::SeqCst);
}

fn get_counts() -> (usize, usize) {
    (
        ALLOCATION_COUNT.load(Ordering::SeqCst),
        ALLOCATION_BYTES.load(Ordering::SeqCst),
    )
}

// --------------------------------------------------------------------------
// Reproduction
// --------------------------------------------------------------------------

/// Simulates a consumer who has a dynamically-created font family string
/// (e.g., from user input, config file, or runtime construction) and wants
/// to reuse it across multiple frames without allocating.
#[test]
fn dynamic_font_family_allocates_every_frame() {
    // Setup (done once at app startup)
    let base_inline = ComputedInlineStyle::default();
    let base_paragraph = ComputedParagraphStyle::default();

    // Consumer's dynamically-created font family string.
    // In a real app, this might come from user preferences, a config file, etc.
    let dynamic_font_name: String = "My Custom Dynamic Font".to_string();

    // Simulate multiple frames
    for frame in 0..3 {
        reset_counts();

        // -----------------------------------------------------------------
        // PROBLEM 1: No From<String> impl, must clone to keep original
        // -----------------------------------------------------------------
        // The consumer wants to reuse `dynamic_font_name` across frames,
        // but FontFamily takes ownership via Cow::Owned. To keep the original,
        // they must clone.
        let font_family = FontFamily::Single(FontFamilyName::Named(Cow::Owned(
            dynamic_font_name.clone(), // 🔴 ALLOCATION: clone to preserve original
        )));

        let (alloc_after_fontfamily, _) = get_counts();
        println!("Frame {frame}: After FontFamily creation: {alloc_after_fontfamily} allocations");

        // Create styled text with the font family
        let style = InlineStyle::new().font_family(Specified::Value(font_family));
        let mut styled =
            StyledText::new("Hello, world!", base_inline.clone(), base_paragraph.clone());
        styled.apply_span(styled.range(0..5).unwrap(), style);

        let (alloc_after_style, _) = get_counts();
        println!("Frame {frame}: After apply_span: {alloc_after_style} allocations");

        // -----------------------------------------------------------------
        // PROBLEM 2: Resolution clones the FontFamily
        // -----------------------------------------------------------------
        // Inside build_layout_from_styled_text -> resolved_inline_runs_coalesced
        // -> resolve_inline_declarations, the font_family is cloned:
        //
        //   out.font_family = resolve_specified(...).clone();
        //
        // This clones Cow::Owned(String), causing another allocation.

        // -----------------------------------------------------------------
        // PROBLEM 3: Parley lowering clones again
        // -----------------------------------------------------------------
        // In push_run_diffs:
        //
        //   StyleProperty::FontFamily(run.font_family().clone())
        //
        // Another clone of the FontFamily -> another String allocation.

        let _layout =
            build_layout_from_styled_text(&mut layout_cx, &mut font_cx, &styled, 1.0, true, ());

        let (alloc_after_layout, bytes) = get_counts();
        println!(
            "Frame {frame}: After layout build: {alloc_after_layout} allocations ({bytes} bytes)"
        );
        println!();

        // We'd expect 0 allocations if we could reuse our String,
        // but instead we see multiple allocations per frame.
        assert!(
            alloc_after_layout >= 2,
            "Expected at least 2 allocations for dynamic font family, got {alloc_after_layout}"
        );
    }
}

/// Contrast: static font families don't allocate on clone because
/// Cow::Borrowed(&'static str) just copies the pointer.
#[test]
fn static_font_family_does_not_allocate() {
    let base_inline = ComputedInlineStyle::default();
    let base_paragraph = ComputedParagraphStyle::default();

    // Static string - Cow::Borrowed, cloning is free
    let font_family = FontFamily::Single(FontFamilyName::Named(Cow::Borrowed("Arial")));

    reset_counts();

    let style = InlineStyle::new().font_family(Specified::Value(font_family));
    let mut styled = StyledText::new("Hello!", base_inline.clone(), base_paragraph.clone());
    styled.apply_span(styled.range(0..5).unwrap(), style);

    let _layout =
        build_layout_from_styled_text(&mut layout_cx, &mut font_cx, &styled, 1.0, true, ());

    let (alloc_count, _) = get_counts();
    println!("Static font family: {alloc_count} allocations");

    // With static strings, cloning Cow::Borrowed is just a pointer copy.
    // There will still be some allocations (Vec for runs, etc.) but not
    // for the font family string itself.
}
```
</details>

styled_text/src/style/declarations.rs
```diff
@@ -0,0 +1,361 @@
...
+/// A set of specified inline declarations for a span.
+///
+/// This is a declaration list (not a “one field per property” struct). When multiple declarations
+/// of the same property are present, the last declaration wins.
+#[derive(Clone, Debug, Default, PartialEq)]
+pub struct InlineStyle {
+    declarations: Vec<InlineDeclaration>,
+}
+
+impl InlineStyle {
```
taj-p, line 54: Most of these methods can be inline

examples/vello_cpu_render_styled_text/src/main.rs
```diff
@@ -0,0 +1,253 @@
...
+    ]));
+    let variation_style =
+        InlineStyle::new().with_font_variations(FontVariations::list(vec![FontVariation::new(
+            Tag::new(b"wght"),
+            750.0,
+        )]));
+
+    styled.apply_span(
+        styled.range(find_range(&text, "StyledText")).unwrap(),
+        styled_text_style,
```
taj-p, line 89: What are your thoughts about the allocation patterns of this API?

It's currently difficult to reuse allocations across evolutions of rich text.

`InlineStyle` allocates a `Vec<InlineDeclaration>`. 
 - It has `.clear()` for reuse, but `apply_span` takes ownership, preventing scratch-storage patterns.

`StyledText::new` allocates vectors for:
 - `AttributedText::attributes` (span storage)
 - `ParagraphStyle::declarations`

There's no `StyledText::clear()` or `reset()` to reuse these allocations across different text. For example, `ParagraphStyle` lacks a `.clear()`.

Is `StyledText` where consumers will store their text model? Or do you see `StyledText` as a proxy that sits between user land models and various downstream text crates? I.e., does a user store their text model in `StyledText` or do they lower their own model into `StyledText`?

Going by the current API, given the "write once, read for layout" approach and inability to "read back" styles that were subsequently input, I believe this is a transient intermediary - is that true?

If so, then I think we will want to support a "clear and refill" pattern to allow for allocation reuse across builds and have a stated goal of cheap construction and rebuild. 

---

That said, perhaps I've totally missed the mark here, which I am wont to do 😅 !!! So, I would appreciate more context if that's the case.

taj-p (commented):
I spent my allotted review time this morning getting back up to speed with `attributed_text` (that crate is very nice! 🏆 ) and understanding the high level structure of the code. 

I haven't yet delved into low or mid level details entirely and will follow that up either tomorrow or Monday or later today if we're lucky!

examples/tiny_skia_render_styled_text/src/main.rs
waywardmonkeys: You a specify the range explicitly ... just that it makes for fragile code for tests.

styled_text_parley/src/lib.rs
waywardmonkeys: Applied locally.

styled_text/src/style/declarations.rs
waywardmonkeys: Done. (in next push)

examples/vello_cpu_render_styled_text/src/main.rs
waywardmonkeys: `StyledText` is intended to be a durable attributed-text model (think “attributed string”): it can be retained as an app’s in-memory representation (often parsed from markup), but it can also be used transiently as a layout input if you already have your own model. We should document that intent more clearly.

On allocation reuse: agreed. Even without a full editing/mutation story yet, it should support a “clear and refill” rebuild pattern so consumers can reuse capacity across evolutions of the same document. I’ve added `StyledText::set_text`/`clear_*`-style APIs (and lower-level `AttributedText::set_text` + `ParagraphStyle::clear`) that clear spans/declarations while retaining storage. Longer-term, we’ll need richer mutation APIs (insert/delete with span adjustment), but this is a good first step toward cheap rebuilds.

styled_text_parley/src/lib.rs
waywardmonkeys: Good call — with an owned String font name, the current “owned-in-declarations → owned-in-computed → owned-in-lowering” path can cause repeated allocations/clones.

A couple of clarifications + what I think we should do:

* In the first step, the example doesn’t have to allocate: you can avoid dynamic_font_name.clone() by moving the `String` into `Cow::Owned(dynamic_font_name)` if you only need to build it once. But that doesn’t address the later clones you pointed out.
* The later clones are the real issue: we currently treat `FontFamily` as an owned value that gets cloned as styles are resolved and then lowered. That’s fine for static/borrowed font lists, but it’s not great for runtime-created family names.

I agree with your suggestion on direction: we should aim for an API that is “non-allocating by design” for this case, and push any unavoidable allocation costs to the `styled_text` ↔ Parley boundary until Parley can participate.

Concretely, the shape I’d like to explore next is: store dynamic family names behind shared storage (e.g. `Arc<str>` or an interned string) in the `styled_text` model so cloning is cheap and allocation is paid once per distinct family name. Then lowering can borrow from that shared storage. Whether Parley needs changes depends on how far we want this to propagate, but we can improve things substantially on the `styled_text` side even before that.

I’ll follow up with a separate PR proposal focused specifically on dynamic font-family storage/cloning (so it doesn’t complicate this one).

styled_text/src/resolve/computed.rs
```diff
@@ -0,0 +1,201 @@
...
+impl Default for ComputedInlineStyle {
+    fn default() -> Self {
+        Self {
+            font_family: crate::style::GenericFamily::SansSerif.into(),
+            font_size_px: 16.0,
+            font_width: FontWidth::NORMAL,
+            font_style: FontStyle::default(),
+            font_weight: FontWeight::NORMAL,
+            font_variations: Arc::from([]),
+            font_features: Arc::from([]),
```
taj-p, line 60: These allocations trigger even when these settings are overwritten with new settings or if they're never used. What do you think of making them `Option` to avoid these allocations?

```diff
diff --git a/styled_text/src/resolve/computed.rs b/styled_text/src/resolve/computed.rs
index 2356a79..4ce4034 100644
--- a/styled_text/src/resolve/computed.rs
+++ b/styled_text/src/resolve/computed.rs
@@ -37,8 +37,8 @@ pub struct ComputedInlineStyle {
     pub(crate) font_width: FontWidth,
     pub(crate) font_style: FontStyle,
     pub(crate) font_weight: FontWeight,
-    pub(crate) font_variations: Arc<[FontVariation]>,
-    pub(crate) font_features: Arc<[FontFeature]>,
+    pub(crate) font_variations: Option<Arc<[FontVariation]>>,
+    pub(crate) font_features: Option<Arc<[FontFeature]>>,
     pub(crate) locale: Option<Language>,
     pub(crate) underline: bool,
     pub(crate) strikethrough: bool,
@@ -56,8 +56,8 @@ impl Default for ComputedInlineStyle {
             font_width: FontWidth::NORMAL,
             font_style: FontStyle::default(),
             font_weight: FontWeight::NORMAL,
-            font_variations: Arc::from([]),
-            font_features: Arc::from([]),
+            font_variations: None,
+            font_features: None,
             locale: None,
             underline: false,
             strikethrough: false,
@@ -110,13 +110,13 @@ impl ComputedInlineStyle {
     /// Returns computed font variation settings (OpenType axis values).
     #[inline]
     pub fn font_variations(&self) -> &[FontVariation] {
-        &self.font_variations
+        self.font_variations.as_deref().unwrap_or(&[])
     }
 
     /// Returns computed font feature settings (OpenType feature values).
     #[inline]
     pub fn font_features(&self) -> &[FontFeature] {
-        &self.font_features
+        self.font_features.as_deref().unwrap_or(&[])
     }
 
     /// Returns the locale/language tag, if any.
diff --git a/styled_text/src/resolve/engine.rs b/styled_text/src/resolve/engine.rs
index a7ec2c4..0eb64df 100644
--- a/styled_text/src/resolve/engine.rs
+++ b/styled_text/src/resolve/engine.rs
@@ -232,25 +232,25 @@ fn resolve_line_height(
 #[inline]
 fn resolve_variations(
     specified: &Specified<FontVariations>,
-    parent: &Arc<[FontVariation]>,
-    initial: &Arc<[FontVariation]>,
-) -> Arc<[FontVariation]> {
+    parent: &Option<Arc<[FontVariation]>>,
+    initial: &Option<Arc<[FontVariation]>>,
+) -> Option<Arc<[FontVariation]>> {
     match specified {
-        Specified::Inherit => Arc::clone(parent),
-        Specified::Initial => Arc::clone(initial),
-        Specified::Value(value) => Arc::clone(value.as_arc_slice()),
+        Specified::Inherit => parent.clone(),
+        Specified::Initial => initial.clone(),
+        Specified::Value(value) => Some(Arc::clone(value.as_arc_slice())),
     }
 }
 
 #[inline]
 fn resolve_features(
     specified: &Specified<FontFeatures>,
-    parent: &Arc<[FontFeature]>,
-    initial: &Arc<[FontFeature]>,
-) -> Arc<[FontFeature]> {
+    parent: &Option<Arc<[FontFeature]>>,
+    initial: &Option<Arc<[FontFeature]>>,
+) -> Option<Arc<[FontFeature]>> {
     match specified {
-        Specified::Inherit => Arc::clone(parent),
-        Specified::Initial => Arc::clone(initial),
-        Specified::Value(value) => Arc::clone(value.as_arc_slice()),
+        Specified::Inherit => parent.clone(),
+        Specified::Initial => initial.clone(),
+        Specified::Value(value) => Some(Arc::clone(value.as_arc_slice())),
     }
 }
```

styled_text/src/style/settings.rs
```diff
@@ -0,0 +1,46 @@
...
+/// Font feature settings (OpenType feature values).
+///
+/// This is a typed wrapper over a list of [`FontFeature`] values.
+#[derive(Clone, PartialEq, Debug, Default)]
+pub struct FontFeatures(Arc<[FontFeature]>);
+
+impl FontFeatures {
+    /// Creates settings from a parsed list.
+    #[inline]
+    pub fn list(list: impl Into<Arc<[FontFeature]>>) -> Self {
```
taj-p, line 37: I'd like your consultation about this signature. On one hand, taking `impl Into<Arc<...>>` makes the API ergonomic, but on the other hand, it can hide allocation cost.

In a hot loop, someone might use a `vec!` directly because that's the path of least resistance:

```rs
for span in my_spans {
    InlineStyle::new().with_font_features(FontFeatures::list(vec![
        FontFeature::new(Tag::new(b"liga"), 1),
    ]))
}
```

Instead of the more performant:

```rs
let enabled_liga = Arc::from([FontFeature::new(Tag::new(b"liga"), 1)]);
for span in my_spans {
    InlineStyle::new().with_font_features(FontFeatures::list(Arc::clone(&enabled_liga)))
}
```

If we're focused on allocation efficiency and performance, it might make more sense to make the allocation more visible - aligning with our goal of minimising allocations.

Users who care about performance will appreciate the honesty, whereas users who don't can simply wrap with `Arc::from()`.

An alternative would be to change `list` to take `Arc<T>` directly.

---

That said, I'm curious what your thoughts are here. Certainly, this might be better overcome with documentation - I'm curious on your stance and philosophy when it comes to signatures like these.

examples/vello_cpu_render_styled_text/src/main.rs
taj-p: I love that clean change to support a better clear and refill pattern via `set_text` and `clear_*` API. Very nice!!

One question I am wondering about is how we would tackle the allocations lost to `InlineStyle::declarations`? 

I think if we aim to be both a durable model and a transient stepping stone to rich text layout, we might want to address those allocations.

Currently, `apply_span` takes ownership of the attribute (`InlineStyle`), which means each span's `Vec<InlineDeclaration>` gets moved into the attributed text storage. When `clear_attributes()` is called, those `InlineStyle` values are dropped along with their internal allocations (i.e. `InlineStyles::declarations`).

With the current API, I imagine we could use a pooling object from which consumers get the `InlineStyle` and `StyledText` returns `InlineStyle` allocations in "clear"-like operations. But that may require API change (which perhaps could be additive).

```rs
let mut pool = InlineStylePool::new();

let style = pool.get();
style.push_declaration(...);
styled.apply_span(range, style);

styled.clear_spans_into(&mut pool);
```

Separately, I've noted that there are other cases where we may want to control allocations (e.g. the `ResolvedInlineRuns` struct). So, another option would be for the entry point into any `StyledText` API to be through some context struct where we can manage allocations (durable pools and transient allocations) if we want to hide that from consumers:

```rs
let ctx = StyledTextContext::new();
let inline_style = ctx.create_inline_style();
let styled_text = ctx.create_styled_text();

...

ctx.resolve_inline_runs(styled_text); // ctx manages re-useable allocations for `resolve_inline_runs`
```

This approach may better balance the goal of minimal allocations and API ergonomics - otherwise, to achieve minimal allocations, consumers may need to manage cache objects and pass them in on an as-needs basis (but that approach may also work!).

In any case, I highlight this case for your consideration - perhaps it's something that may help direct strategy.

styled_text/src/document.rs
```diff
@@ -0,0 +1,115 @@
...
+use attributed_text::TextStorage;
+
+use crate::block::Block;
+
+/// A flat sequence of semantic blocks.
+///
+/// This is intended to carry semantic structure (headings, lists, etc.) suitable for accessibility,
+/// even before layout provides geometry.
+#[derive(Debug)]
+pub struct StyledDocument<T: Debug + TextStorage, A: Debug> {
```
taj-p, line 17: Question: Do we want to extract this into a `styled_document` crate (along with `Block` and `BlockKind`)?

I'm curious about the strategy here - does `StyledDocument`'s purpose align with `StyledText`'s dual role as both durable storage and transient intermediary?

I am assuming that some consumers may never use `StyledDocument` (e.g., those translating their own document model into StyledText transiently). 

By extracting `StyledDocument` to its own crate, we'd get:

- API discipline: Forces `StyledText` functionality needed by document abstractions to be pub, which means any consumer building their own document model gets the same capabilities.
- Clear layering: `styled_document` becomes a reference implementation of "document built on StyledText" rather than privileged internal code.

That said, StyledDocument is currently quite small, so this may be premature. Curious whether you see it growing significantly or staying minimal.

styled_text/src/runs.rs
```diff
@@ -0,0 +1,269 @@
...
+#[derive(Clone, Debug)]
+pub struct ResolvedInlineRuns<'a, T: Debug + TextStorage, A: Debug + HasInlineStyle> {
+    pub(crate) styled: &'a StyledText<T, A>,
+    pub(crate) boundaries: Vec<usize>,
+    start_offsets: Vec<usize>,
+    start_events: Vec<usize>,
+    end_offsets: Vec<usize>,
+    end_events: Vec<usize>,
+    spans: Vec<Span<'a>>,
+    active: Vec<usize>,
```
taj-p, line 62: Do we expect any of these to grow above `u32::MAX`?

We can halve the cost of these structures on 64 bit systems by using `u32`. It means that we have to `as usize` for access, but since these are internal implementation details, that's preferable for performance.

styled_text/src/runs.rs
```diff
@@ -0,0 +1,269 @@
...
+    pub(crate) index: usize,
+}
+
+impl<'a, T: Debug + TextStorage, A: Debug + HasInlineStyle> ResolvedInlineRuns<'a, T, A> {
+    pub(crate) fn new(styled: &'a StyledText<T, A>) -> Self {
```
taj-p, line 67: This allocates a lot. From the perspective of a transient consumer, it would be beneficial if we could pass in a context or allocations structure from which `new` could re-use allocations. Would it be possible to enable passing in a context or similar?

styled_text/src/resolve/computed.rs
waywardmonkeys: I think that these and the font family are maybe not a great design from top to bottom (through the entire stack) ... and maybe we need a more fundamental re-think and address it from `text_primitives` up through `parley` and into here.

styled_text/src/style/settings.rs
waywardmonkeys: I think this is more evidence that the design of this part of things needs consideration from top to bottom.

styled_text/src/runs.rs
waywardmonkeys: I've done this locally and ... it makes the code a bit grosser.

Parley already imposes u32 requirements on some stuff, so it isn't like we can actually address more than `u32::MAX`.

I don't think this is a clear win, but I'll try it out more.

waywardmonkeys:
#631 is the new iteration on this. Wildly different. Yet benefitted from things in this review (like sweep-line stuff).

PR #499, opened 2025-12-30
`attributed_text`: Some utility methods

waywardmonkeys:


attributed_text/src/attributed_text.rs
```diff
@@ -81,6 +96,13 @@ impl<T: Debug + TextStorage, Attr: Debug> AttributedText<T, Attr> {
         Ok(())
     }
 
+    /// Iterate over all attributes and the ranges they apply to.
+    ///
+    /// Attributes are yielded in the order they were applied.
+    pub fn iter_attributes(&self) -> impl Iterator<Item = (&Range<usize>, &Attr)> {
```
tomcur:
```suggestion
    pub fn attributes_iter(&self) -> impl Iterator<Item = (&Range<usize>, &Attr)> {
```

tomcur: Perhaps this could be an [`ExactSizeIterator`](https://doc.rust-lang.org/stable/core/iter/trait.ExactSizeIterator.html).

waywardmonkeys: Done and force pushed.

PR #500, opened 2025-12-30
Text primitives extract

waywardmonkeys:
This PR introduces a new core vocabulary crate, `text_primitives`, and
updates `parley`/`fontique` to use it for shared “leaf” text types
(OpenType tags/settings, font attributes, generic families, language
tags, bidi controls, and wrap/break enums).

The goal is to keep these fundamental types small, reusable,
`no_std`-friendly, and insulated from higher-level dependencies,
while making it easier for the rest of the text stack to share
the same representations.

## What’s in `text_primitives`

- `Tag`/`Setting` (OpenType interop) with `Tag` stored as `[u8; 4]`.
- Font attributes: `FontWeight`, `FontWidth`, `FontStyle`.
- CSS-ish wrap/break controls: `WordBreak`, `OverflowWrap`, `TextWrapMode`, plus `BaseDirection`.
- Bidi controls: `BidiControl`, `BidiDirection`, `BidiOverride`.
- `GenericFamily` (CSS generic family vocabulary).
- `Language` as a compact, zero-allocation `language[-Script][-REGION]`
  prefix type, with strict parsing and a `parse_prefix` API for
  extracting the remainder.

## Notable behavior / API changes

- Enums remain exhaustive (avoids `#[non_exhaustive]`) where stability/panic-safety matters.
- CSS parsing helpers are explicitly named `parse_css` (font weight/width/style).
- `Display` for font attributes is aligned with valid CSS output (e.g.
  numeric font-weight prints as a number; oblique prints as
  `oblique <angle>deg`).
- `Language` parsing is stricter to avoid silent data loss; trailing subtags
  are structurally validated then discarded; `extlang` is rejected;
  `parse_prefix` returns `(Language, remainder)`.
- Added `#[must_use]` to `Language` accessors.

## Integration changes

- `parley` and `fontique` now use `text_primitives` types instead of defining/duplicating
  equivalents.
- Fontconfig numeric mappings remain in `fontique`

### DIFF
--- a/text_primitives/src/generic_family.rs
+++ b/text_primitives/src/generic_family.rs
@@ -6,7 +6,6 @@ use core::fmt;
 /// Generic font families, named after CSS.
 #[derive(Clone, Copy, Debug, PartialEq, Eq, Hash)]
 #[repr(u8)]
-#[non_exhaustive]
 pub enum GenericFamily {
     /// Glyphs have finishing strokes, flared or tapering ends, or have actual serifed endings.
     Serif = 0,


waywardmonkeys:
As discussed in #495.

This needs some bike shedding by @dfrg and maybe others.

It needs a better name for this crate as well.

text_primitives/src/tag.rs
```diff
@@ -0,0 +1,180 @@
+// Copyright 2025 the Parley Authors
+// SPDX-License-Identifier: Apache-2.0 OR MIT
+
+use core::fmt;
+
+/// A 4-byte OpenType tag (for example `wght`, `liga`).
+#[derive(Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Hash, Debug)]
+#[repr(transparent)]
+pub struct Tag(u32);
```
nicoburns: It would probably be better if this wrapped `[u8; 4]`. For non-ambiguity of endianess, and also to match `font_types::Tag`.

text_primitives/src/font.rs
```diff
@@ -0,0 +1,553 @@
...
+            "normal" => Self::NORMAL,
+            "bold" => Self::BOLD,
+            _ => Self(s.parse::<f32>().ok()?),
+        })
+    }
+
+    /// Creates a weight value from a Fontconfig weight.
+    ///
+    /// The values are determined based on the `fonts.conf` documentation.
+    pub fn from_fontconfig(weight: i32) -> Self {
```
dfrg: The `from_fontconfig` constructors are odd (we don't offer this for CoreText or DirectWrite) and should probably just be moved back to `fontique` as private functions.

text_primitives/src/tag.rs
waywardmonkeys: Done.

text_primitives/src/font.rs
waywardmonkeys: Moved.

text_primitives/src/text.rs
```diff
@@ -0,0 +1,58 @@
...
+    Anywhere,
+    /// Like `Anywhere`, but treated differently for min-content sizing in some engines.
+    BreakWord,
+}
+
+/// Control over non-"emergency" line-breaking.
+///
+/// See: <https://www.w3.org/TR/css-text-4/#text-wrap-mode>
+#[derive(Clone, Copy, Debug, PartialEq, Eq, Default)]
+#[non_exhaustive]
```
taj-p: Why are these `non_exhaustive`? I'm not sure whether we should encode such volatility in a vocabulary crate. If CSS adds a new `TextWrapMode` or `OverflowWrap` then it's probably significant enough to warrant a major version bump.

I'm a little worried that the lack of stability of a `non_exhaustive` enum means that downstream consumers then pay the price of silently failing when a new variant is added (from, for example, catch all arms added in `fontique/src/font.rs` and `parley/src/analysis/mod.rs`).

text_primitives/src/font.rs
```diff
@@ -0,0 +1,472 @@
...
+    pub const EXTRA_BOLD: Self = Self(800.0);
+
+    /// Weight value of 900.
+    pub const BLACK: Self = Self(900.0);
+
+    /// Weight value of 950.
+    pub const EXTRA_BLACK: Self = Self(950.0);
+
+    /// Creates a new weight value.
+    pub fn new(weight: f32) -> Self {
```
taj-p:
```suggestion
    pub const fn new(weight: f32) -> Self {
```

text_primitives/src/font.rs
```diff
@@ -0,0 +1,472 @@
...
+        Self(weight)
+    }
+
+    /// Returns the underlying weight value.
+    pub fn value(self) -> f32 {
```
taj-p:
```suggestion
    pub const fn value(self) -> f32 {
```

taj-p: This and many other methods in this file should be `#[inline(always)]`

text_primitives/src/font.rs
```diff
@@ -0,0 +1,472 @@
...
+    ///
+    /// ```
+    /// use text_primitives::FontWeight;
+    ///
+    /// assert_eq!(FontWeight::parse("normal"), Some(FontWeight::NORMAL));
+    /// assert_eq!(FontWeight::parse("bold"), Some(FontWeight::BOLD));
+    /// assert_eq!(FontWeight::parse("850"), Some(FontWeight::new(850.0)));
+    /// assert_eq!(FontWeight::parse("invalid"), None);
+    /// ```
+    pub fn parse(s: &str) -> Option<Self> {
```
taj-p: From an API perspective, it's odd for us to support values like `SEMI_LIGHT`, `EXTRA_BOLD`, etc and even displaying `100` outputs the string `"thin"` but we're only able to parse CSS style strings ("normal", "bold", or a number). 

I think this should be named `parse_css` (for all three types) so that it's honest about what it does and leaves room for a more permissive `parse` or `FromStr` impl later.

text_primitives/src/font.rs
```diff
@@ -0,0 +1,472 @@
...
+    pub const EXPANDED: Self = Self(1.25);
+
+    /// Width that is 150% of normal.
+    pub const EXTRA_EXPANDED: Self = Self(1.5);
+
+    /// Width that is 200% of normal.
+    pub const ULTRA_EXPANDED: Self = Self(2.0);
+
+    /// Creates a new width value with the given ratio.
+    pub fn from_ratio(ratio: f32) -> Self {
```
taj-p: As above, could we please inline all and const the functions for which it's possible?

text_primitives/src/font.rs
```diff
@@ -0,0 +1,472 @@
...
+        }
+    }
+}
+
+/// Visual width of a font — a relative change from the normal aspect ratio.
+///
+/// In variable fonts, this can be controlled with the `wdth` axis. This uses an `f32` so that it
+/// can represent the full range of values possible with variable fonts.
+///
+/// In CSS, this corresponds to the `font-width` (`font-stretch`) property.
```
taj-p: We're losing some information from the original documentation.

```
/// Visual width of a font-- a relative change from the normal aspect
/// ratio, typically in the range `0.5` to `2.0`.
///
/// The default value is [`FontWidth::NORMAL`] or `1.0`.
///
/// In variable fonts, this can be controlled with the `wdth` [axis]. This
/// is an `f32` so that it can represent the same range of values as the
/// `wdth` axis.
///
/// In Open Type, the `u16` [`usWidthClass`] field has 9 values, from 1-9,
/// which doesn't allow for the wide range of values possible with variable
/// fonts.
///
/// See <https://fonts.google.com/knowledge/glossary/width>
///
/// In CSS, this corresponds to the [`font-width`] property.
///
/// This has also been known as "stretch" and has a legacy CSS name alias,
/// [`font-stretch`].
///
/// [axis]: crate::AxisInfo
/// [`usWidthClass`]: https://learn.microsoft.com/en-us/typography/opentype/spec/os2#uswidthclass
/// [`font-width`]: https://www.w3.org/TR/css-fonts-4/#font-width-prop
/// [`font-stretch`]: https://www.w3.org/TR/css-fonts-4/#font-stretch-prop
```

text_primitives/src/font.rs
```diff
@@ -0,0 +1,472 @@
...
+// SPDX-License-Identifier: Apache-2.0 OR MIT
+
+use core::fmt;
+
+/// Visual weight class of a font, typically on a scale from 1.0 to 1000.0.
+///
+/// In variable fonts, this can be controlled with the `wght` axis. This uses an `f32` so that it
+/// can represent the full range of values possible with variable fonts.
+///
+/// In CSS, this corresponds to the `font-weight` property.
```
taj-p: We are losing some information compared to the original (namely, the link):

```
/// Visual weight class of a font, typically on a scale from 1.0 to 1000.0.
///
/// The default value is [`FontWeight::NORMAL`] or `400.0`.
///
/// In variable fonts, this can be controlled with the `wght` [axis]. This
/// is an `f32` so that it can represent the same range of values as the
/// `wght` axis.
///
/// See <https://fonts.google.com/knowledge/glossary/weight>
///
/// In CSS, this corresponds to the [`font-weight`] property.
///
/// [axis]: crate::AxisInfo
/// [`font-weight`]: https://www.w3.org/TR/css-fonts-4/#font-weight-prop
```

text_primitives/src/font.rs
```diff
@@ -0,0 +1,472 @@
...
+        } else {
+            write!(f, "{}%", self.percentage())
+        }
+    }
+}
+
+/// Visual style or “slope” of a font.
+///
+/// In variable fonts, this can be controlled with the `ital` and `slnt` axes for italic and
+/// oblique styles, respectively.
```
taj-p, line 295: Losing some information here.

```
/// Visual style or 'slope' of a font.
///
/// The default value is [`FontStyle::Normal`].
///
/// In variable fonts, this can be controlled with the `ital`
/// and `slnt` [axes] for italic and oblique styles, respectively.
/// This uses an `f32` for the `Oblique` variant so so that it
/// can represent the same range of values as the `slnt` axis.
///
/// See <https://fonts.google.com/knowledge/glossary/style>
///
/// In CSS, this corresponds to the [`font-style`] property.
///
/// [axes]: crate::AxisInfo
/// [`font-style`]: https://www.w3.org/TR/css-fonts-4/#font-style-prop
```

text_primitives/src/font.rs
```diff
@@ -0,0 +1,472 @@
...
+#[derive(Clone, Copy, Debug, PartialEq, Default)]
+#[non_exhaustive]
+pub enum FontStyle {
+    /// `normal`.
+    #[default]
+    Normal,
+    /// `italic`.
+    Italic,
```
taj-p:
```suggestion
    /// An upright or "roman" style.
    #[default]
    Normal,
    /// Generally a slanted style, originally based on semi-cursive forms.
    /// This often has a different structure from the normal style.
    Italic,
```

text_primitives/src/bidi.rs
taj-p, line 1: Per prior comment, why do we add `non_exhaustive` to all these enums? I think I'm missing some context - from a consumer's perspective, it seems like it'll make it easy to fall into silent bugs due to requiring catch all arms.

fontique/src/font.rs
```diff
@@ -132,6 +132,7 @@ impl FontInfo {
                         }
                     }
                 }
+                _ => {}
```
taj-p: (This is an example match arm I'm talking about)

text_primitives/src/generic_family.rs
```diff
@@ -0,0 +1,115 @@
...
+#[repr(u8)]
+#[non_exhaustive]
+pub enum GenericFamily {
+    /// Glyphs have finishing strokes, flared or tapering ends, or have actual serifed endings.
+    Serif = 0,
+    /// Glyphs have stroke endings that are plain.
+    SansSerif = 1,
+    /// All glyphs have the same fixed width.
+    Monospace = 2,
+    /// Glyphs are partially or completely connected, resembling handwriting.
```
taj-p:
```suggestion
    /// Glyphs in cursive fonts generally have either joining strokes or other
    /// cursive characteristics beyond those of italic typefaces. The glyphs
    /// are partially or completely connected, and the result looks more like
    /// handwritten pen or brush writing than printed letter work.
```

text_primitives/src/generic_family.rs
```diff
@@ -0,0 +1,115 @@
...
+    UiSerif = 6,
+    /// The default user interface sans-serif font.
+    UiSansSerif = 7,
+    /// The default user interface monospace font.
+    UiMonospace = 8,
+    /// The default user interface font that has rounded features.
+    UiRounded = 9,
+    /// Fonts that are specifically designed to render emoji.
+    Emoji = 10,
+    /// Fonts designed to represent mathematics.
```
taj-p:
```suggestion
    /// This is for the particular stylistic concerns of representing
    /// mathematics: superscript and subscript, brackets that cross several
    /// lines, nesting expressions, and double struck glyphs with distinct
    /// meanings.
```

text_primitives/src/generic_family.rs
```diff
@@ -0,0 +1,115 @@
...
+    Math = 11,
+    /// A Chinese character style between Song and Kai forms.
```
taj-p:
```suggestion
    /// A particular style of Chinese characters that are between serif-style
    /// Song and cursive-style Kai forms. This style is often used for
    /// government documents.
```

text_primitives/src/generic_family.rs
```diff
@@ -0,0 +1,115 @@
...
+    FangSong = 12,
+}
```
taj-p, line 45:
```suggestion
    // NOTICE: If a new value is added, be sure to modify `MAX_VALUE`.
}
```

fontique/src/impl_fontconfig.rs
```diff
@@ -0,0 +1,137 @@
...
+///
+/// This is intentionally kept in `fontique` (rather than `text_primitives`) because it is a
+/// Fontconfig-specific mapping.
+pub trait FromFontconfig: Sized {
+    /// Creates a value from the corresponding Fontconfig numeric representation.
+    fn from_fontconfig(value: i32) -> Self;
+}
+
+impl FromFontconfig for FontWidth {
+    fn from_fontconfig(width: i32) -> Self {
```
taj-p: Could we not lose this doco?

```
    /// Creates a new width attribute with the given value from Fontconfig.
    ///
    /// The values are determined based on the [fonts.conf documentation].
    ///
    /// [fonts.conf documentation]: https://www.freedesktop.org/software/fontconfig/fontconfig-user.html
```

fontique/src/impl_fontconfig.rs
```diff
@@ -0,0 +1,137 @@
...
+/// Conversion from Fontconfig numeric values.
+///
+/// This is intentionally kept in `fontique` (rather than `text_primitives`) because it is a
+/// Fontconfig-specific mapping.
+pub trait FromFontconfig: Sized {
+    /// Creates a value from the corresponding Fontconfig numeric representation.
+    fn from_fontconfig(value: i32) -> Self;
+}
+
+impl FromFontconfig for FontWidth {
```
taj-p: The `font.rs` file orders these in `FontWeight`, `FontWidth`, `FontStyle` - could we please be consistent?

fontique/src/impl_fontconfig.rs
```diff
@@ -0,0 +1,137 @@
...
+            125 => Self::EXPANDED,
+            150 => Self::EXTRA_EXPANDED,
+            200 => Self::ULTRA_EXPANDED,
+            _ => Self::from_ratio(width as f32 / 100.0),
+        }
+    }
+}
+
+impl FromFontconfig for FontWeight {
+    fn from_fontconfig(weight: i32) -> Self {
```
taj-p, line 31: Could we add this doco?

```
    /// Creates a new weight attribute with the given value from Fontconfig.
    ///
    /// The values are determined based on the [fonts.conf documentation].
    ///
    /// [fonts.conf documentation]: https://www.freedesktop.org/software/fontconfig/fontconfig-user.html
```

fontique/src/impl_fontconfig.rs
```diff
@@ -0,0 +1,137 @@
...
+                return Self::new(ot_a + (ot_b - ot_a) * t);
+            }
+        }
+
+        Self::EXTRA_BLACK
+    }
+}
+
+impl FromFontconfig for FontStyle {
+    fn from_fontconfig(slant: i32) -> Self {
```
taj-p, line 101: Could we add this doco?

```
    /// Creates a new style attribute with the given value from Fontconfig.
    ///
    /// The values are determined based on the [fonts.conf documentation].
    ///
    /// [fonts.conf documentation]: https://www.freedesktop.org/software/fontconfig/fontconfig-user.html
```

fontique/src/impl_fontconfig.rs
taj-p, line 1: Maybe we just need the font config link somewhere?

parley/src/analysis/mod.rs
taj-p, line 1: Do we want to pay the maintenance cost of these catch all match arms?

text_primitives/src/lib.rs
```diff
@@ -0,0 +1,55 @@
...
+// See https://linebender.org/wiki/canonical-lints/
+// These lints shouldn't apply to examples or tests.
+#![cfg_attr(not(test), warn(unused_crate_dependencies))]
+// These lints shouldn't apply to examples.
+#![warn(clippy::print_stdout, clippy::print_stderr)]
+// Targeting e.g. 32-bit means structs containing usize can give false positives for 64-bit.
+#![cfg_attr(target_pointer_width = "64", warn(clippy::trivially_copy_pass_by_ref))]
+// END LINEBENDER LINT SET
+#![cfg_attr(docsrs, feature(doc_cfg))]
+#![no_std]
```
taj-p, line 39: What is the `std` feature in the crate doing? Is it needed? I was expecting something like:

```
#![cfg_attr(not(feature = "std"), no_std)]
```

text_primitives/src/text.rs
```diff
@@ -0,0 +1,58 @@
...
+    KeepAll,
+}
+
+/// Control over "emergency" line-breaking.
+///
+/// See: <https://www.w3.org/TR/css-text-3/#overflow-wrap-property>
+#[derive(Clone, Copy, Debug, PartialEq, Eq, Default)]
+#[non_exhaustive]
+pub enum OverflowWrap {
+    /// Only break at opportunities specified by word-breaking rules.
```
taj-p:
```suggestion
    /// Even with extremely long words, lines can only break at places specified in [`WordBreak`].
```

text_primitives/Cargo.toml
```diff
@@ -0,0 +1,29 @@
...
+
+[package.metadata.docs.rs]
+all-features = true
+# There are no platform specific docs.
+default-target = "x86_64-unknown-linux-gnu"
+targets = []
+
+[features]
+default = ["std"]
+std = []
```
taj-p, line 22: Is this used?

parley/src/shape/mod.rs
```diff
@@ -483,6 +483,10 @@ fn variations_iter<'a>(
     synthesis: &'a fontique::Synthesis,
     item: Option<&'a [FontVariation]>,
 ) -> impl Iterator<Item = harfrust::Variation> + 'a {
+    fn tag_to_harfrust(tag: crate::setting::Tag) -> harfrust::Tag {
+        harfrust::Tag::new(&tag.to_bytes())
+    }
```
taj-p: If this is only used once, let's just inline it. It's probably not even worth using on line 391.

```
                    tag: harfrust::Tag::new(&variation.tag.to_bytes()),
```


```suggestion
```

text_primitives/src/language.rs
```diff
@@ -0,0 +1,192 @@
...
+        match self {
+            Self::InvalidLanguage => f.write_str("invalid primary language subtag"),
+        }
+    }
+}
+
+impl FromStr for Language {
+    type Err = ParseLanguageError;
+
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
```
taj-p, line 296: Could we add tests for this?

text_primitives/src/language.rs
```diff
@@ -0,0 +1,192 @@
...
+    /// The “undefined” language (`und`).
+    pub const UND: Self = Self::from_language_bytes(*b"und", 3);
+
+    /// Parses a language tag, keeping only language/script/region.
+    pub fn parse(s: &str) -> Result<Self, ParseLanguageError> {
+        s.parse()
+    }
+
+    /// Returns the canonical string form (`language[-Script][-REGION]`).
+    pub fn as_str(&self) -> &str {
```
taj-p, line 65: Could we always inline this?

text_primitives/src/language.rs
```diff
@@ -0,0 +1,192 @@
...
+        core::str::from_utf8(&self.bytes[..self.len as usize])
+            .expect("Language stores only ASCII bytes")
```
taj-p:
```suggestion
            .expect("only ASCII")
```

text_primitives/src/language.rs
```diff
@@ -0,0 +1,192 @@
...
+    }
+
+    /// Returns the primary language subtag (lowercase).
+    pub fn language(&self) -> &str {
+        core::str::from_utf8(&self.bytes[..self.language_len as usize])
+            .expect("Language stores only ASCII bytes")
```
taj-p:
```suggestion
            .expect("only ASCII")
```

text_primitives/src/language.rs
```diff
@@ -0,0 +1,192 @@
...
+    /// Returns the script subtag (titlecase), if present.
+    pub fn script(&self) -> Option<&str> {
+        if self.script_len == 0 {
+            return None;
+        }
+        let start = self.language_len as usize + 1;
+        let end = start + self.script_len as usize;
+        Some(
+            core::str::from_utf8(&self.bytes[start..end])
+                .expect("Language stores only ASCII bytes"),
```
taj-p:
```suggestion
                .expect("only ASCII"),
```

text_primitives/src/language.rs
```diff
@@ -0,0 +1,192 @@
...
+        }
+        let mut start = self.language_len as usize;
+        if self.script_len != 0 {
+            start += 1 + self.script_len as usize;
+        }
+        start += 1;
+        let end = start + self.region_len as usize;
+        Some(
+            core::str::from_utf8(&self.bytes[start..end])
+                .expect("Language stores only ASCII bytes"),
```
taj-p:
```suggestion
                .expect("only ASCII"),
```

Let's save a small handful of bytes

text_primitives/src/language.rs
taj-p, line 1: Maybe `std` could be used here by implementing Error when enabled?

```rs
#[cfg(feature = "std")]
impl std::error::Error for ParseLanguageError {}
```

text_primitives/src/language.rs
taj-p, line 65: Should we add `#[must_use]` to these accessors?

text_primitives/src/language.rs
```diff
@@ -0,0 +1,192 @@
...
+
+impl Language {
+    /// The maximum length of the canonical `language[-Script][-REGION]` form.
+    pub const MAX_LEN: usize = 12;
+
+    /// The “undefined” language (`und`).
+    pub const UND: Self = Self::from_language_bytes(*b"und", 3);
+
+    /// Parses a language tag, keeping only language/script/region.
+    pub fn parse(s: &str) -> Result<Self, ParseLanguageError> {
```
taj-p, line 38: Could we always inline this?

text_primitives/src/language.rs
```diff
@@ -0,0 +1,192 @@
...
+            || !language_bytes.iter().all(|b| b.is_ascii_alphabetic())
+        {
+            return Err(ParseLanguageError::InvalidLanguage);
+        }
+
+        let mut out = Self {
+            bytes: [0; 12],
+            len: 0,
+            language_len: u8::try_from(language_bytes.len())
+                .expect("language subtag length is at most 3"),
```
taj-p:
```suggestion
            #[allow(clippy::cast_possible_truncation, reason = "language subtag length is at most 3")]
            language_len: language_bytes.len() as u8,
```

We just tested this above - could simply add a unit test that verifies this truncation to avoid the expect. We could do this for region and script too.

text_primitives/src/language.rs
taj-p, line 296: This parsing does not error for invalid regions or scripts. For example, the following invalid subtags are dropped. I'm not sure what the correct behaviour is, but it's odd for a parse function and `ParseLanguageError` to not error on malformed input.

Since this is a primitive crate, it might be worth leaning towards stricter parsing because otherwise consumers will suffer silent data loss: "en-Latin-US" loses "US" with no indication making debugging difficult.

Maybe it's worth adding `InvalidScript` and `InvalidRegion` to `ParseLanguageError`. For example:

```rs
    #[test]
    fn invalid_script_drops_subsequent_region() {
        // "Latin" (5 chars) fails script check, fails region check, 
        // and "US" is never examined because we only check one subtag for region
        let lang = Language::parse("en-Latin-US").unwrap();
        assert_eq!(lang.as_str(), "en");
        assert_eq!(lang.script(), None);
        assert_eq!(lang.region(), None); // US is lost!
    }
```

I believe ICU4X errors on [invalid subtags](https://github.com/unicode-org/icu4x/blob/04a93b1b2dc6c0503cdc63f04b4e415a4012103c/components/locale_core/src/parser/langid.rs#L68)

text_primitives/src/language.rs
```diff
@@ -0,0 +1,192 @@
...
+#[derive(Clone, Copy, PartialEq, Eq, Hash)]
+pub struct Language {
+    bytes: [u8; 12],
+    len: u8,
+    language_len: u8,
+    script_len: u8,
+    region_len: u8,
+}
+
+impl Language {
```
taj-p, line 29: @dfrg and @waywardmonkeys, I have one strategic question about `text_primitives`. 

Ultimately, we're going to need to pass some of these types to ICU4X where we will need to pay a translation cost. So, the question I'm asking is whether it's worth our time investment to do this versus improving the ICU4X types / reduce their allocations - if ICU4X allocates, then we're going to likely allocate because we need to use their APIs.

That said, perhaps this is standard operating procedure for text crates? I note that Harfrust defines its own `Setting` / `Tag` / `Language` / etc? I also like how this keeps `text_primitives` lean without any heavy dependencies, so maybe we do both?

taj-p (commented):
Nice work!! This is great. I've left some minor comments and questions. If @dfrg is happy with the high level approach (please see the below comment @dfrg about ICU4X), then I'm happy to approve the impl details once they're addressed.

text_primitives/src/text.rs
nicoburns: +1 for exhaustive enums in most places (unless we expect the type to change frequently). I think it's going to be hard/awkward to make the APIs panic-safe otherwise.

text_primitives/src/font.rs
dfrg: This inconsistency bothered me too (and I think I'm the one that originally wrote it!). +1 to renaming to `parse_css` and adding more permissive conversions for those that want to capture more values.

Maybe add a `to_css` method that returns only valid CSS values as well.

text_primitives/src/language.rs
dfrg: I'm guessing our MSRV will be high enough to use `core::error::Error`?

text_primitives/src/language.rs
dfrg: > So, the question I'm asking is whether it's worth our time investment to do this versus improving the ICU4X types / reduce their allocations - if ICU4X allocates, then we're going to likely allocate because we need to use their APIs.

Beyond allocations, my primary concern is being able to update the icu4x dependencies without causing a breaking change to parley et al. We've attempted to do the same with skrifa and harfrust.

> That said, perhaps this is standard operating procedure for text crates? I note that Harfrust defines its own `Setting` / `Tag` / `Language` / etc?

Agreed, this is unfortunate. We can (and should!) unify some of these harfrust types with read-fonts. But I think that is orthogonal to the purpose of this crate.

> I also like how this keeps `text_primitives` lean without any heavy dependencies, so maybe we do both?

Yes, my preference would be for this crate to remain minimal, providing just types for interop of higher level text layout libraries.

text_primitives/src/text.rs
dfrg: I'll add a +1 to this as well.

text_primitives/src/language.rs
dfrg: Generally agree with making the parsing a bit more strict. One thing I'd like to test is how the web handles malformed languages. If we set `lang="tr-Latin-TR"` do browsers still handle casing as Turkish?

Additionally, perhaps we should also return the remainder of the string even on a successful parse, allowing the user to process the remaining subtags. For example, `"tr-Latin-TR"` might result in `(Language("tr"), "Latin-TR")`

text_primitives/src/text.rs
waywardmonkeys: These are all exhaustive now.

I think this is a difference of opinion within the Linebender people over whether exhaustive or non-exhaustive is better. I have no interest in getting in the middle of it.

text_primitives/src/font.rs
waywardmonkeys: These are all renamed to `parse_css`.

I didn't add a `to_css`.

text_primitives/src/lib.rs
waywardmonkeys: That's not how we do `no_std` in any of our crates ... this way makes it much easier to maintain `no_std` support.

There's probably no actual usage of `std`, but we tend to have it just in case on our crates.

text_primitives/Cargo.toml
waywardmonkeys: This is boilerplate we have in other crates for forward compatibility. I've added a doc comment listing the "std" feature. I should've added "bytemuck" so I'll do that now.

parley/src/analysis/mod.rs
waywardmonkeys: They're gone now.

text_primitives/src/language.rs
waywardmonkeys: There's now a `parse_prefix` (matching how we do things in `color` roughly).

waywardmonkeys:
I _think_ that I've hit all of the feedback points.

No one has yet suggested a new name.

text_primitives/src/text.rs
taj-p: `BidiControl`, `BidiDirection`, `BidiOverride`, `GenericFamily` are still marked as `non_exhaustive`.

text_primitives/src/language.rs
```diff
@@ -0,0 +1,492 @@
...
+///
+/// This type captures only the `language` + optional `script` + optional `region` subtags from a
+/// BCP 47 language tag, normalized to common casing conventions:
+/// - language: lowercase (2–3 letters)
+/// - script: titlecase (4 letters)
+/// - region: uppercase (2 letters) or digits (3 digits)
+///
+/// Any additional subtags (variants, extensions, private use) are ignored, but the input must
+/// still be well-formed enough that `script`/`region` aren't silently dropped (for example, a
+/// region subtag after a variant is treated as an error).
```
taj-p: When I first read this, I thought it supported [this spec](https://datatracker.ietf.org/doc/html/rfc5646), but I think there are some intentional limitations that we discard.

Is something like this more accurate?


```suggestion
/// A compact, zero-allocation language tag.
///
/// Parses tags matching: `language [-script] [-region]`
///
/// This type captures only the `language` + optional `script` + optional `region` subtags from a
/// BCP 47 language tag. Parsing is case-insensitive; output is normalized to:
/// - language: 2-3 lowercase letters (e.g., `EN` → `en`)
/// - script: 4 titlecase letters (e.g., `LATN` → `Latn`)
/// - region: 2 uppercase letters or 3 digits (e.g., `us` → `US`, `419` → `419`)
///
/// Trailing subtags (variants, extensions, private use) are validated for BCP 47 conformance
/// but discarded. Extended language subtags (`extlang`) are not supported and will error.
```

text_primitives/src/language.rs
```diff
@@ -0,0 +1,492 @@
...
+
+    // optional script or region
+    let Some((start, end)) = next_part_bounds(s, &mut pos) else {
+        return Ok((out, ""));
+    };
+
+    let b = &bytes[start..end];
+    if b.len() == 4 {
+        if !b.iter().all(|c| c.is_ascii_alphabetic()) {
+            return Err(ParseLanguageError::InvalidScript);
```
taj-p, line 162: I think this isn't quite correct because the variant could start with a digit. 

i.e., this test fails:

```rs
    #[test]
    fn parse_4_char_digit_variant() {
        // BCP 47 variants can be 4 alphanumeric chars if they start with a digit.
        // These should be treated as variants (in remainder), not rejected as invalid scripts.
        let lang = Language::parse("de-1996").unwrap();
        assert_eq!(lang.as_str(), "de");
        assert_eq!(lang.script(), None);
        assert_eq!(lang.region(), None);

        // With script before the variant
        let lang = Language::parse("de-Latn-1996").unwrap();
        assert_eq!(lang.as_str(), "de-Latn");
        assert_eq!(lang.script(), Some("Latn"));

        // With region before the variant
        let lang = Language::parse("de-DE-1996").unwrap();
        assert_eq!(lang.as_str(), "de-DE");
        assert_eq!(lang.region(), Some("DE"));

        // parse_prefix should return the variant in the remainder
        let (lang, rest) = Language::parse_prefix("de-1996").unwrap();
        assert_eq!(lang.as_str(), "de");
        assert_eq!(rest, "1996");
    }
```

text_primitives/src/bidi.rs
```diff
@@ -0,0 +1,43 @@
...
+
+/// An inline bidirectional text control.
+///
+/// These map conceptually to Unicode bidi isolate/override controls. They are expressed as style
+/// properties rather than literal control characters.
+///
+/// For background on bidi behavior see UAX #9:
+/// <https://www.unicode.org/reports/tr9/>
+#[derive(Clone, Copy, Debug, PartialEq, Eq, Default)]
+#[non_exhaustive]
```
taj-p:
```suggestion
```

text_primitives/src/bidi.rs
```diff
@@ -0,0 +1,43 @@
...
+    None,
+    /// Isolate this span with the given direction.
+    Isolate(BidiDirection),
+    /// Override directional resolution within this span.
+    Override(BidiOverride),
+}
+
+/// Direction choice used by bidi controls.
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+#[non_exhaustive]
```
taj-p:
```suggestion
```

text_primitives/src/bidi.rs
```diff
@@ -0,0 +1,43 @@
...
+    Auto,
+    /// Left-to-right.
+    Ltr,
+    /// Right-to-left.
+    Rtl,
+}
+
+/// Direction choice used by bidi overrides.
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+#[non_exhaustive]
```
taj-p:
```suggestion
```

text_primitives/src/language.rs
```diff
@@ -0,0 +1,492 @@
...
+        };
+        out.bytes[0] = bytes[0];
+        out.bytes[1] = bytes[1];
+        out.bytes[2] = bytes[2];
+        out.len = len;
+        out
+    }
+}
+
+fn parse_language_prefix(s: &str) -> Result<(Language, &str), ParseLanguageError> {
```
taj-p, line 118: I had a rough read of the implementation. It looks good! That said, I haven't given this a super thorough read through - I'm relying on the tests, some more tests I wrote, and, if there are perf gains to be had, I imagine we can find them if this ever shows up in a perf profile.

taj-p (approved):
LGTM with comments addressed and @dfrg happy with high level direction 🙏 .

So good 🎉 🚀

taj-p:
As for the name, I don't mind `text_primitives`. Happy for someone else to weigh in. Considering it's not as yet a published crate, there might not be a lot of urgency (?). Other options:

- `text_types`
- `text_attrs`
- `text_props`
- ??

> No one has yet suggested a new name.

What's your preference @waywardmonkeys ?

text_primitives/src/language.rs
waywardmonkeys: I've used an edited form of this that tightens the language a bit more to match the actual implementation (since trailing subtags aren't validated for BCP 47 compliance, just structurally looked at).

text_primitives/src/language.rs
waywardmonkeys: Pushed a fix for that.

DJMcNab (approved):
I think that we can land this with just @taj-p's approval; I do agree it would be helpful to have confirmation from Chad that they don't want to see anything changed, but it's pretty clear from their comments in this thread and elsewhere (i.e. in #495) that they is at least not opposed to this direction.

As such, any tweaks from Chad can just as easily come in a follow-up (or, worst case, we can always revert). So to unblock follow-ups (#495), I think we should land it.

PR #501, opened 2026-01-05
text_primitives: improve OpenType settings CSS parsing

waywardmonkeys:
- Add structured parsing errors (kind + byte offset/span) for CSS OpenType settings strings.
- Rename list parsers to `parse_css_list`.
- Update Parley resolver to use the fallible CSS parser.
- Add unit tests covering success cases and key error offsets/spans.

### DIFF
--- a/parley/src/resolve/mod.rs
+++ b/parley/src/resolve/mod.rs
@@ -11,8 +11,8 @@ pub(crate) use range::RangedStyleBuilder;
 use alloc::{vec, vec::Vec};
 
 use super::style::{
-    Brush, FontFamily, FontFamilyName, FontFeature, FontSettings, FontStyle, FontVariation,
-    FontWeight, FontWidth, StyleProperty,
+    Brush, FontFamily, FontFamilyName, FontFeature, FontFeatures, FontStyle, FontVariation,
+    FontVariations, FontWeight, FontWidth, StyleProperty,
 };
 use crate::font::FontContext;
 use crate::style::TextStyle;
@@ -262,15 +262,15 @@ impl ResolveContext {
     /// Resolves font variation settings.
     pub(crate) fn resolve_variations(
         &mut self,
-        variations: &FontSettings<'_, FontVariation>,
+        variations: &FontVariations<'_>,
     ) -> Resolved<FontVariation> {
         match variations {
-            FontSettings::Source(source) => {
+            FontVariations::Source(source) => {
                 self.tmp_variations.clear();
                 self.tmp_variations
                     .extend(FontVariation::parse_css_list(source).map_while(Result::ok));
             }
-            FontSettings::List(settings) => {
+            FontVariations::List(settings) => {
                 self.tmp_variations.clear();
                 self.tmp_variations.extend_from_slice(settings);
             }
@@ -287,15 +287,15 @@ impl ResolveContext {
     /// Resolves font feature settings.
     pub(crate) fn resolve_features(
         &mut self,
-        features: &FontSettings<'_, FontFeature>,
+        features: &FontFeatures<'_>,
     ) -> Resolved<FontFeature> {
         match features {
-            FontSettings::Source(source) => {
+            FontFeatures::Source(source) => {
                 self.tmp_features.clear();
                 self.tmp_features
                     .extend(FontFeature::parse_css_list(source).map_while(Result::ok));
             }
-            FontSettings::List(settings) => {
+            FontFeatures::List(settings) => {
                 self.tmp_features.clear();
                 self.tmp_features.extend_from_slice(settings);
             }
--- a/text_primitives/src/tag.rs
+++ b/text_primitives/src/tag.rs
@@ -108,23 +108,21 @@ impl fmt::Display for ParseSettingsError {
 
 impl core::error::Error for ParseSettingsError {}
 
-/// A single OpenType setting (tag + value).
+/// OpenType font feature setting (tag + `u16` value).
 #[derive(Clone, Copy, PartialEq, Debug)]
-pub struct Setting<T> {
+pub struct FontFeature {
     /// The OpenType tag for this setting.
     pub tag: Tag,
-    /// The setting value.
-    pub value: T,
+    /// The feature value.
+    pub value: u16,
 }
 
-impl<T> Setting<T> {
-    /// Creates a new setting.
-    pub const fn new(tag: Tag, value: T) -> Self {
+impl FontFeature {
+    /// Creates a new feature setting.
+    pub const fn new(tag: Tag, value: u16) -> Self {
         Self { tag, value }
     }
-}
 
-impl Setting<u16> {
     /// Parses a comma-separated list of feature settings according to the CSS grammar.
     ///
     /// On success, yields a sequence of settings. On failure, yields a [`ParseSettingsError`].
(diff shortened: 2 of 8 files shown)


text_primitives/src/tag.rs
```diff
@@ -45,6 +45,69 @@ impl fmt::Display for Tag {
...
+/// Error returned when parsing OpenType settings source strings.
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+pub struct ParseSettingsError {
+    kind: ParseSettingsErrorKind,
+    at: usize,
+    span: Option<(usize, usize)>,
+}
+
+impl ParseSettingsError {
+    fn new(kind: ParseSettingsErrorKind, at: usize) -> Self {
```
taj-p:
```suggestion
    const fn new(kind: ParseSettingsErrorKind, at: usize) -> Self {
```

text_primitives/src/tag.rs
```diff
@@ -45,6 +45,69 @@ impl fmt::Display for Tag {
...
+        Self {
+            kind,
+            at,
+            span: None,
+        }
+    }
+
+    fn with_span(mut self, span: (usize, usize)) -> Self {
```
taj-p:
```suggestion
    const fn with_span(mut self, span: (usize, usize)) -> Self {
```

text_primitives/src/tag.rs
```diff
@@ -45,6 +45,69 @@ impl fmt::Display for Tag {
     }
 }
 
+/// Kinds of errors that can occur when parsing OpenType settings source strings.
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+#[non_exhaustive]
```
taj-p, line 50:
```suggestion
```

Was this intentional? Are error enums different than regular enums for which we want to be exhaustive?

text_primitives/src/tag.rs
```diff
@@ -45,6 +45,69 @@ impl fmt::Display for Tag {
...
+    /// A numeric value was out of range for the target type.
+    OutOfRange,
+}
+
+/// Error returned when parsing OpenType settings source strings.
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+pub struct ParseSettingsError {
+    kind: ParseSettingsErrorKind,
+    at: usize,
+    span: Option<(usize, usize)>,
```
taj-p, line 65: Usually I discourage the use of architecture-dependent types like `usize` to keep behaviour consistent across platforms, but I think it's fine for this type of struct. Curious what your thoughts are.

text_primitives/src/tag.rs
```diff
@@ -63,72 +126,119 @@ impl<T> Setting<T> {
...
             let ch = self.source[pos];
-            ch.is_ascii_whitespace() || ch == b','
-        } {
-            pos += 1;
+            if ch.is_ascii_whitespace() || ch == b',' {
+                pos += 1;
+            } else {
+                break;
+            }
         }
```
taj-p: I believe these lines allow an arbitrary number and order of ` ` and `,`. That's technically not allowed in the CSS spec IIUC, but I understand if you'd prefer to keep the leniency.

i.e., this test passes:
```rs
    #[test]
    fn parse_feature_settings_css_list_ok() {
        let parsed: Result<Vec<_>, _> =
            Setting::<u16>::parse_css_list(r#""liga" on,,, ,   ,,,    'kern', "dlig" off, "salt" 3,"#).collect();
        let settings = parsed.unwrap();

        assert_eq!(settings.len(), 4);
        assert_eq!(settings[0].tag, Tag::parse("liga").unwrap());
        assert_eq!(settings[0].value, 1);
        assert_eq!(settings[1].tag, Tag::parse("kern").unwrap());
        assert_eq!(settings[1].value, 1);
        assert_eq!(settings[2].tag, Tag::parse("dlig").unwrap());
        assert_eq!(settings[2].value, 0);
        assert_eq!(settings[3].tag, Tag::parse("salt").unwrap());
        assert_eq!(settings[3].value, 3);
    }
```

taj-p (approved):

text_primitives/src/tag.rs
waywardmonkeys: Yes! This is a pattern for extensible errors. They are reportable without the caller needing to know.

text_primitives/src/tag.rs
waywardmonkeys: Anything else would be super annoying in terms of casts.

text_primitives/src/tag.rs
waywardmonkeys: Example is Rust's `std::io::Error` which has an `ErrorKind` that is non-exhaustive: https://doc.rust-lang.org/std/io/struct.Error.html

text_primitives/src/tag.rs
waywardmonkeys: Better to start out lenient and make it less lenient if needed in the future. Working on a revision.

waywardmonkeys: Okay, I've pushed a commit that addresses this, adds more tests, and hardens some other stuff in parsing.

parley/src/resolve/mod.rs
```diff
@@ -268,7 +268,7 @@ impl ResolveContext {
             FontSettings::Source(source) => {
                 self.tmp_variations.clear();
                 self.tmp_variations
-                    .extend(FontVariation::parse_list(source));
+                    .extend(FontVariation::parse_css_list(source).map_while(Result::ok));
```
dfrg, line 271: Should this be `filter_map` instead? Seems useful to skip invalid entries when possible.

text_primitives/src/tag.rs
```diff
@@ -45,6 +45,69 @@ impl fmt::Display for Tag {
...
+        };
+        write!(f, "{msg} at byte {}", self.at)
+    }
+}
+
+impl core::error::Error for ParseSettingsError {}
+
 /// A single OpenType setting (tag + value).
 #[derive(Clone, Copy, PartialEq, Debug)]
 pub struct Setting<T> {
```
dfrg: What do you think about hiding this and just exposing concrete `FontVariation` and `FontFeature` types instead? I’m trying to think of a case where we want to handle these generically and none comes to mind.

parley/src/resolve/mod.rs
waywardmonkeys: I'll just give the Codex response here which I think I agree with:

> `filter_map(Result::ok)` only helps if the iterator can continue producing items after an error. Our `parse_css_list` iterators don’t recover: on the first `Err` they set `done = true` and terminate. In that world:
> 
> - `map_while(Result::ok)` = “take items until first error”
> - `filter_map(Result::ok)` = “drop the error, but there are no later items anyway”
> 
> So switching to `filter_map` wouldn’t actually “skip invalid entries when possible” today; it would just slightly change the signal of intent (and arguably mislead future readers into thinking we do recovery).
> 
> If the intent is “stop at first parse error, keep earlier valid ones” (matching the old `take_while(ok)` behavior), `map_while(Result::ok)` is the clearer expression.
> If the intent is “try to salvage later entries”, we’d need to change the parser to recover and keep iterating; then `filter_map` would make sense.

text_primitives/src/tag.rs
waywardmonkeys: I've done this as a separate commit in the series ... it ended up being sizable once I took it to the logical endpoint. Or what I think is the logical endpoint.

waywardmonkeys:
I've added the `FeatureSpec` stuff as a commit in this series ... but if there's a lot of work to be done to it, then I'm happy to pull it out and make it a separate PR (or potentially drop it).

PR #502, opened 2026-01-05
text_primitives: move CSS font-family into vocabulary layer

waywardmonkeys:
* Add text_primitives::FontFamily (CSS font-family value) and FontFamilyName (single entry), including CSS list parsing. 
* Parsing is fallible and returns structured errors with byte offsets/spans.
* Update Parley to use StyleProperty::FontFamily and TextStyle::font_family (renaming the old FontStack), and update tests/benches/examples accordingly.


text_primitives/src/font_family.rs
```diff
@@ -0,0 +1,305 @@
...
+/// Error returned when parsing CSS `font-family` strings.
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+pub struct ParseFontFamilyError {
+    kind: ParseFontFamilyErrorKind,
+    at: usize,
+    span: Option<(usize, usize)>,
+}
+
+impl ParseFontFamilyError {
+    fn new(kind: ParseFontFamilyErrorKind, at: usize) -> Self {
```
taj-p:
```suggestion
    const fn new(kind: ParseFontFamilyErrorKind, at: usize) -> Self {
```

text_primitives/src/font_family.rs
```diff
@@ -0,0 +1,305 @@
...
+        Self {
+            kind,
+            at,
+            span: None,
+        }
+    }
+
+    fn with_span(mut self, span: (usize, usize)) -> Self {
```
taj-p:
```suggestion
    const fn with_span(mut self, span: (usize, usize)) -> Self {
```

parley/src/tests/test_builders.rs
```diff
@@ -191,7 +193,9 @@ fn create_root_style() -> TextStyle<'static, ColorBrush> {
 ///
 /// The [`RangedBuilder`] version of [`create_root_style`].
 fn set_root_style(rb: &mut RangedBuilder<'_, ColorBrush>) {
-    rb.push_default(FontStack::from(FONT_STACK));
+    rb.push_default(StyleProperty::FontFamily(FontFamily::from(
+        FONT_FAMILY_LIST,
+    )));
```
taj-p: Should this and other places be:

```rs
    rb.push_default(FontFamily::from(FONT_FAMILY_LIST));
```

since `FontFamily` implements `StyleProperty`?

text_primitives/src/font_family.rs
```diff
@@ -0,0 +1,374 @@
...
+        let first = self.source[pos];
+        let mut start = pos;
+        if matches!(first, b'"' | b'\'') {
+            let quote = first;
+            let opening_quote = pos;
+            pos += 1;
+            start += 1;
+            while pos < self.len {
+                if self.source[pos] == quote {
+                    let name = core::str::from_utf8(self.source.get(start..pos)?).ok()?;
```
taj-p: The use of `?` here and on line 274 seem to be encoding invariants that will silently fail on violation. Shouldn't these always succeed and their failure will cause silent errors on internal logic errors?

taj-p (approved):
LGTM with comments addressed

text_primitives/src/font_family.rs
waywardmonkeys: Awarding you the “Silent None Slayer” 🔇∅⚔️ award for spotting this.

parley/src/tests/test_builders.rs
waywardmonkeys: Added this and similar to `FontFamilyName` at the same time:

```
+impl<'a> FontFamily<'a> {
+    /// Creates a `font-family` value consisting of a single named family.
+    pub const fn named(name: &'a str) -> Self {
+        Self::Single(FontFamilyName::named(name))
+    }
+}
```

PR #503, opened 2026-01-06
Expand test coverage

taj-p:
Parley is undergoing significant development. I'm a little worried that we won't capture regressions with the existing test suite. This PR increases the test coverage for all styles and does a little re-structuring of existing tests.

There is no source code change here. Only tests.

I think a follow up could split out `test_basic.rs` into files like `test_align.rs`, `test_inline_boxes.rs`, and so on. But, I think that can occur later.


parley/src/tests/utils/samples.rs
conor-93, line 1: 💯

conor-93 (approved):
This is a fantastic addition with really nice test grouping - thankyou! 

For ligatures breaking with letter spacing and word spacing not impacting justified text content width (thanks for identifying these!), should they be raised as issues? Looking through existing TODOs, I don't see any explicit mentions of text layout bugs like these.

taj-p:
Added issue in https://github.com/linebender/parley/issues/515. Great catch! Thanks Conor!

PR #505, opened 2026-01-10
text_primitives: add font-family into_owned helpers

waywardmonkeys:
Add `FontFamilyName::into_owned` and `FontFamily::into_owned` to convert borrowed font-family values into owned (`'static`) values for storage.

Includes unit tests for `Source`, `Single`, and `List` conversions.

### DIFF
--- a/text_primitives/src/font_family.rs
+++ b/text_primitives/src/font_family.rs
@@ -6,6 +6,7 @@
 extern crate alloc;
 
 use alloc::borrow::Cow;
+use alloc::vec::Vec;
 use core::fmt;
 
 use crate::GenericFamily;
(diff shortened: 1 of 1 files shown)


text_primitives/src/font_family.rs
```diff
@@ -144,6 +144,19 @@ impl<'a> FontFamilyName<'a> {
             done: false,
         }
     }
+
+    /// Returns an owned (`'static`) version of this font family name.
+    ///
+    /// This is useful when you want to store a `FontFamilyName` without borrowing from an input
+    /// string slice.
```
DJMcNab, line 152: There are arguments for saying that this will clone/alloc if the value is borrowed, although that's easily inferable.

text_primitives/src/font_family.rs
```diff
@@ -179,6 +192,25 @@ impl<'a> FontFamily<'a> {
     pub const fn named(name: &'a str) -> Self {
         Self::Single(FontFamilyName::named(name))
     }
+
+    /// Returns an owned (`'static`) version of this `font-family` value.
+    ///
+    /// This is useful when you want to store a `FontFamily` without borrowing from an input string
+    /// slice.
+    #[must_use]
```
DJMcNab, line 206: I'm not sure if it's good style to always give a message with the `must_use` attribute; I can't easily find anything online about it.
I think either of adding something or not would be fine here.

DJMcNab (approved):

text_primitives/src/font_family.rs
waywardmonkeys: I think it would be repetitive here. (And we have very few of those in our codebases.)

text_primitives/src/font_family.rs
waywardmonkeys: Added to the doc comment.

DJMcNab (approved):

PR #506, opened 2026-01-10
parley: drop `Source` OpenType settings, simplify settings types

waywardmonkeys:
* Remove `Source` variants and `From<&str>` for `FontFeatures` / `FontVariations`; Parley no longer parses CSS-like OpenType strings during resolution.
* Add a unit test demonstrating `text_primitives::{FontFeature, FontVariation}::parse_css_list` and passing the typed lists into Parley.
* Replace the now-single-variant enums with newtype wrappers around `Cow<[FontFeature]>` / `Cow<[FontVariation]>` for simpler, slice-like ergonomics.
* Add `StyleProperty` conversions from `&[FontFeature]`, `&[FontVariation]`, and `&[T; N]`; simplify resolution helpers to operate on slices and update tests.

### DIFF
--- a/parley/src/lib.rs
+++ b/parley/src/lib.rs
@@ -116,7 +116,6 @@ mod util;
 
 pub mod editing;
 pub mod layout;
-pub mod setting;
 pub mod style;
 
 #[cfg(test)]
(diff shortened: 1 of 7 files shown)


taj-p (commented):
I think this might be OBE? If not, please re-request after addressing Chad's comment and I can take a look 🙏

PR #508, opened 2026-01-12
text_primitives/fontique/parley: unify `Language` and `Script` typing

waywardmonkeys:
* Use `text_primitives::Language` as the shared locale/language-tag type across Fontique and Parley; `StyleProperty::Locale` now carries `Option<Language>` instead of `Option<&str>`, removing locale parsing from Parley’s resolve path.
* Add `text_primitives::Script` (ISO 15924) and update script handling, fallback, and shaping to use it end-to-end.
* Store typed `Language` in `fontique::FallbackKey` (and Android fallback tables) to avoid locale-string allocations; add `FallbackKey::locale_str()` for platform backends.
* Drop Parley’s direct `icu_locale_core` dependency and make Fontique’s `icu_locale_core` optional behind the `icu_properties` feature.


fontique/src/backend/android.rs
```diff
@@ -60,7 +59,7 @@ impl SystemFonts {
             );
         }
 
-        let mut locale_fallback = vec![];
+        let mut locale_fallback: Vec<(Language, FamilyId)> = vec![];
```
xorgy: This is completely unnecessary.

robertbastian:
This basically undoes #475. You now have a custom `Script` type again, that you need to parse into ICU4X's script type in `fontique`. Why does the ICU4X `Script` (and `LanguageIdentifier`) type not work for you?

PR #513, opened 2026-01-13
attributed_text: add validated `TextRange` and make it the primary API

waywardmonkeys:
* Add `TextRange` as a validated UTF-8 byte range (start <= end, in-bounds, char-boundary aligned) with docs + tests.
* Change `AttributedText::apply_attribute` to take `TextRange` (infallible).
* Add `AttributedText::apply_attribute_bytes(Range<usize>, ...) -> Result<_, Error>` for one-off validation.
* Deduplicate range validation logic and simplify tests to prefer `TextRange`.

### DIFF
--- a/attributed_text/src/text_range.rs
+++ b/attributed_text/src/text_range.rs
@@ -1,0 +1,190 @@
+// Copyright 2025 the Parley Authors
+// SPDX-License-Identifier: Apache-2.0 OR MIT
+
+use core::ops::Range;
+
+use crate::{Endpoint, Error, TextStorage};
+
+/// A validated byte range into a UTF-8 text buffer.
+///
+/// This is a convenience wrapper around `Range<usize>` that carries invariants useful for
+/// attributed text APIs:
+///
+/// - `start <= end`
+/// - `start` and `end` are within the text bounds
+/// - `start` and `end` lie on UTF-8 codepoint boundaries
+///
+/// **Why `TextRange`?**
+///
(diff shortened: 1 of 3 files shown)


attributed_text/src/text_range.rs
```diff
@@ -0,0 +1,187 @@
...
+/// ```
+#[derive(Clone, Copy, Debug, PartialEq, Eq, Hash)]
+pub struct TextRange {
+    start: usize,
+    end: usize,
+}
+
+impl TextRange {
+    /// Returns a validated `TextRange` for the provided text.
+    pub fn new<T: TextStorage>(text: &T, range: Range<usize>) -> Result<Self, Error> {
```
tomcur, line 50:
```suggestion
    #[inline]
    pub fn new<T: TextStorage>(text: &T, range: Range<usize>) -> Result<Self, Error> {
```

tomcur (approved):
This looks good. It might be ergonomic to have a method on `AttributedText` or `TextStorage` to get such a range (something like `AttributedText::text_range(&self, range: Range<usize>) -> Result<TextRange, Error>`), allowing user code to use text ranges without importing/naming the type.

In a sense `AttributedText::apply_attribute_bytes` is now safer to use than `AttributedText::apply_attribute`, as it checks the range explicitly, but the docs already mention effectively that and it's not really an issue.

attributed_text/src/text_range.rs
waywardmonkeys: There were a few other things that you've been inlined ... done that and force pushed.

PR #516, opened 2026-01-16
parley: add style-run builder; accept precomputed runs from `styled_text`

waywardmonkeys:
- Add `LayoutContext::style_run_builder` and a `StyleRunBuilder` API for constructing a Layout from contiguous, non-overlapping style runs, skipping Parley’s internal range-splitting step.
- Update `styled_text_parley` to lower StyledText’s computed runs directly into `StyleRunBuilder` using `TextStyle<'static, '_, B>`.
- Add a parity test ensuring `StyleRunBuilder` output matches `RangedBuilder` for equivalent styles.

### DIFF
--- a/parley/src/builder.rs
+++ b/parley/src/builder.rs
@@ -10,7 +10,7 @@ use super::style::{Brush, StyleProperty, TextStyle, WhiteSpaceCollapse};
 use super::layout::Layout;
 
 use alloc::string::String;
-use core::ops::RangeBounds;
+use core::ops::{Bound, Range, RangeBounds};
 
 use crate::inline_box::InlineBox;
 use crate::resolve::{ResolvedStyle, StyleRun, tree::ItemKind};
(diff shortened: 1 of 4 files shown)


waywardmonkeys:
This builds on top of PR #495 so only look at the right commit here ... but this shows a way to build a `Layout` going through less code and no resolution.

xStrom:
Why is that span example hard to represent? It's not clear to me, especially as there is no BiDi in that example.

waywardmonkeys:
@nicoburns This is probably just the first phase ... I think those sorts of things might change with time. There's a lot of stuff downstream from this.

nicoburns:
@waywardmonkeys Unrelated to my previous comment:

Is there a way to reuse styles with this API? For example, in `<p>first<b>second</b>third</p>`, the spans representing "first" and "third" would have the exact same set of styles. I believe the current `TreeBuilder` is taking advantage of this to simply refer to the already pushed Style rather than pushing a duplicate style. Although looking at `build_layout_into` it looks to me that this may result in incorrect "char_info` being set.

Basically, it might nice to have a layer of indirection between the style structs and the text ranges, so that more than one text range can point to the same style struct.

nicoburns:
> You mentioned it yourself, you add the paddings together. Your style flattener needs to know which properties are additive and which are overrides.

Handling it as part of flattening is interesting possibility that for some reason I had not considered. I'd need to think it through, but I think this would probably work.

> Of course for web compat you'll also get to deal with the juicy quirks of margin collapsing.

Luckily margin collapsing only applies to block-level elements, not inline-level elements, so we don't need to deal with this in Parley.

taj-p (commented):
I love the idea of enabling an API into Parley that bypasses style resolution - to pass precomputed runs. Please re-request review once you're ready

parley/src/builder.rs
```diff
@@ -73,6 +73,96 @@ impl<B: Brush> RangedBuilder<'_, B> {
...
+impl<B: Brush> StyleRunBuilder<'_, B> {
+    /// Reserves additional capacity for styles and runs.
+    ///
+    /// This is an optional optimization for callers that know counts
+    /// up front; call it before pushing styles and runs to reduce
+    /// reallocations.
+    pub fn reserve(&mut self, additional_styles: usize, additional_runs: usize) {
+        self.lcx.style_table.reserve(additional_styles);
+        self.lcx.style_runs.reserve(additional_runs);
+    }
```
xorgy, line 97: Might be interesting at some point to have some reuse, so that there's not a reallocation cycle every time... but I guess I won't know until I'm using it.

parley/src/lib.rs
```diff
@@ -7,15 +7,18 @@
 //! - [`FontContext`] and [`LayoutContext`] are resources which should be shared globally (or at coarse-grained boundaries).
 //!   - [`FontContext`] is database of fonts.
 //!   - [`LayoutContext`] is scratch space that allows for reuse of allocations between layouts.
-//! - [`RangedBuilder`] and [`TreeBuilder`] which are builders for creating a [`Layout`].
-//!     - [`RangedBuilder`] allows styles to be specified as a flat `Vec` of spans
-//!     - [`TreeBuilder`] allows styles to be specified as a tree of spans
+//! - Builders for creating a [`Layout`]:
+//!     - [`RangedBuilder`]: styles specified as a flat `Vec` of property spans
+//!     - [`TreeBuilder`]: styles specified as a tree of spans
```
xorgy: Might be nice to make the indent consistent with the other sections in here, if these lines are going to get touched, else put them back where they were.

waywardmonkeys: Moved them to the left.

parley/src/builder.rs
waywardmonkeys: These are reused already. This is just another knob.

nicoburns:
I'm currently unwell, so I only skimmed, but this looks pretty reasonable to me.

parley/src/builder.rs
```diff
@@ -73,6 +73,96 @@ impl<B: Brush> RangedBuilder<'_, B> {
...
+        assert!(style_index <= u16::MAX as usize, "too many styles");
+        self.lcx.style_table.push(resolved);
+        style_index as u16
+    }
+
+    /// Adds a style run referencing an entry from the style table.
+    ///
+    /// Runs must be contiguous and non-overlapping, and must cover
+    /// `0..text.len()` once all runs have been added.
+    pub fn push_style_run(&mut self, style_index: u16, range: impl RangeBounds<usize>) {
```
dfrg: If this takes a length rather than a range then we don’t have to check the lower bound and we remove one potential panic.

taj-p (commented):
I ran out of time today. But this definitely looks reasonable 👍🎉

PR #518, opened 2026-01-16
Upgrade windows crate to 0.62

nicoburns:
Upgrades the `windows` and `windows-core` crates to the latest 0.62 versions

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1305,8 +1305,8 @@ dependencies = [
  "roxmltree",
  "smallvec",
  "text_primitives",
- "windows 0.58.0",
- "windows-core 0.58.0",
+ "windows 0.62.2",
+ "windows-core 0.62.2",
  "yeslogic-fontconfig-sys",
 ]
 
@@ -5417,11 +5417,23 @@ version = "0.61.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "9babd3a767a4c1aef6900409f85f5d53ce2544ccdfaa86dad48c91782c6d6893"
 dependencies = [
- "windows-collections",
+ "windows-collections 0.2.0",
  "windows-core 0.61.2",
- "windows-future",
+ "windows-future 0.2.1",
  "windows-link 0.1.3",
- "windows-numerics",
+ "windows-numerics 0.2.0",
+]
+
+[[package]]
+name = "windows"
+version = "0.62.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "527fadee13e0c05939a6a05d5bd6eec6cd2e3dbd648b9f8e447c6518133d8580"
+dependencies = [
+ "windows-collections 0.3.2",
+ "windows-core 0.62.2",
+ "windows-future 0.3.2",
+ "windows-numerics 0.3.1",
 ]
 
 [[package]]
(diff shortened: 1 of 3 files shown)


fontique/src/backend/dwrite.rs
```diff
@@ -350,10 +350,10 @@ impl IDWriteTextAnalysisSource_Impl for TextSource_Impl<'_> {
         &self,
         textposition: u32,
         textlength: *mut u32,
-        numbersubstitution: *mut Option<IDWriteNumberSubstitution>,
+        numbersubstitution: OutRef<'_, IDWriteNumberSubstitution>,
     ) -> windows::core::Result<()> {
+        numbersubstitution.write(None)?;
```
nicoburns, line 355: Not 100% sure this is the correct way to translate this into the new API

xStrom: I looked a bit into this and I think the answer is yes, this is the way. However, the surrounding context seems quite arcane.

The [COM memory management rules] state that **all out pointer parameters (including pointer members of a caller-allocate callee-fill structure) must explicitly be set to NULL.** This is what the old code was presumably doing with `None` and this new `write` call replicates that.

I do find it scary that "safe" Rust code can be written so easily that results in uninitialized memory being read by the OS. I did not find any signs that `windows` adds any safety for this purpose, but it's possible I missed it. I did find various flavors of the following statement: *metadata doesn't currently include information about which fields require automatic initialization.* Meaning that the bindgen just doesn't know which parameters need extra safety vs. where it would ruin things.

It's not just the lack of well structured metadata either. Human readable docs are quite limited. [`GetNumberSubstitution` docs] have little content with no info about safety. The Rust bindings ([1], [2]) have literally zero documentation.

A few relevant links to this specific change:

* [windows-rs#3025](https://github.com/microsoft/windows-rs/pull/3025) added `OutRef`
* [windows-rs#3386](https://github.com/microsoft/windows-rs/pull/3386) started using it for COM interfaces
* [`OutRef` source](https://github.com/microsoft/windows-rs/blob/a7197b9d072112f0ab3559ef16db5ef5cd728bb5/crates/libs/core/src/out_ref.rs)

[COM memory management rules]: https://learn.microsoft.com/en-us/previous-versions/ms810016(v=msdn.10)?redirectedfrom=MSDN#memory-management-rules
[`GetNumberSubstitution` docs]: https://learn.microsoft.com/en-us/windows/win32/api/dwrite/nf-dwrite-idwritetextanalysissource-getnumbersubstitution
[1]: https://microsoft.github.io/windows-docs-rs/doc/windows/Win32/Graphics/DirectWrite/trait.IDWriteTextAnalysisSource_Impl.html#tymethod.GetNumberSubstitution
[2]: https://microsoft.github.io/windows-docs-rs/doc/windows/Win32/Graphics/DirectWrite/struct.IDWriteTextAnalysisSource.html#method.GetNumberSubstitution

nicoburns: Yeah, I also looked for docs on this and struggled to find any.

Cargo.lock
```diff
@@ -5417,11 +5417,23 @@ version = "0.61.3"
...
+name = "windows"
+version = "0.62.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "527fadee13e0c05939a6a05d5bd6eec6cd2e3dbd648b9f8e447c6518133d8580"
+dependencies = [
+ "windows-collections 0.3.2",
+ "windows-core 0.62.2",
+ "windows-future 0.3.2",
+ "windows-numerics 0.3.1",
 ]
```
T-256, line 5437: Now, here is 4th version of `windows` crate in the lockfile :)

nicoburns: I suspect it's Winit and Vello (via wgpu) dependencies pulling in old versions. The good news is that these are only in examples. The bad news is that Winit's release schedule is so slow that this may take a while to resolve.

xStrom: Sucks to have four of them, but the solution is for everyone to use the latest not to be stuck on an old version together.

PR #519, opened 2026-01-17
parley: split TextStyle lifetimes for family vs settings

waywardmonkeys:
Change `TextStyle` to take two lifetimes: one for `FontFamily` and one for OpenType settings (`FontVariations` / `FontFeatures`). This lets callers keep a long-lived font-family value while borrowing per-run settings without forcing both to share a single lifetime.

Update internal APIs and tests to use `TextStyle<'_, '_, B>`.


waywardmonkeys:
If we do something to change the lifetime structure of the font family, then this might be less necessary. But that also isn't currently clear...

PR #521, opened 2026-01-21
Fix double-scaling bug with non-hinted COLR glyphs

conor-93:
Copies https://github.com/linebender/vello/pull/1370


behdad:
Does this correctly handle non-uniform scales when component glyphs might be rotated, etc?

behdad:
> My experience with other COLRv1 renderers, in HarfBuzz, is that you need to undo the font transform and do apply the full component transform.

https://github.com/harfbuzz/harfbuzz/blob/6e37798a16c841da1c223ebf478526a84d9c6235/src/OT/Color/COLR/COLR.hh#L960-L966

PR #522, opened 2026-01-22
Upgrade to Vello 0.7, Sparse Strips 0.0.6

nicoburns:
A simple version bump.
This affects the vello/kurbo/peniko/ui-events versions used in tests, examples and `parley_draw` crate.


xorgy:
I think it makes sense also to update the ui-events and ui-events-winit to v0.3.0 in the example, since that aligns the Kurbo versions.

PR #526, opened 2026-01-26
Move most tests to new `parley_test` crate

valadaptive:
As discussed [on Zulip](https://xi.zulipchat.com/#narrow/channel/205635-parley/topic/Moving.20the.20test.20suite.20into.20its.20own.20crate/with/567755941), it's beneficial to move (most of) the integration test suite into its own crate.

This mirrors how Vello's tests are set up, and lets us share test infrastructure between different Parley crates. This is especially important as we begin to split up functionality by crate, and will allow us to test `parley_draw` in the future.

The text analysis and builder tests remain in the Parley crate proper, since they use a lot of private APIs. Everything else, and in particular every snapshot test, has been moved into `parley_tests`.

The split isn't completely clean. I had to make a couple cluster methods public (but `#[doc(hidden)]`) in Parley, since some common test assertions use them. I also duplicated the font context loading code between the `parley_tests` crate and the builder tests in Parley, since the latter can't be moved into `parley_tests`.

I also had to modify a couple tests and assertions to only use public APIs: `assert_eq_layout_data_alignments` has been replaced with a `assert_eq_layout_alignments` helper that doesn't reach into `LayoutData`, and the `spacing_causes_style_run_breaks` test is now a regular snapshot test (with some extra checks).

### DIFF
--- a/parley_tests/README.md
+++ b/parley_tests/README.md
@@ -4,9 +4,9 @@ Integration test suite for the Parley text layout library.
 
 ## Overview
 
-This crate contains the integration test suite for `parley`. It follows a similar
-architecture to `vello_sparse_tests` and uses snapshot testing to validate layout
-correctness.
+This crate contains the integration test suite for `parley`. It follows a similar architecture to
+[`vello_sparse_tests`](https://github.com/linebender/vello/tree/main/sparse_strips/vello_sparse_tests) and uses snapshot
+testing to validate layout correctness.
 
 ## Running Tests
 
@@ -41,7 +41,7 @@ cargo xtask review
 
 ## Detect dead snapshots
 
-The following command shows snapshots that are not used in any test. The command also allows to delete such snaphosts.
+The following command shows snapshots that are not used in any test. The command also allows to delete such snapshots.
 
 ```bash
 cargo xtask dead-snapshots


parley_tests/Cargo.toml
```diff
@@ -0,0 +1,30 @@
...
+repository.workspace = true
+publish = false
+autotests = false
+
+[[test]]
+name = "tests"
+path = "tests/mod.rs"
+
+[dependencies]
+# TODO: the cursor_ligature_selection test depends on system fonts.
```
taj-p: This TODO reads like a statement. Should we include that we want to remove the dependence on system fonts otherwise are tests aren't necessarily hermetic

parley_tests/tests/mod.rs
taj-p, line 1: Nice module docs 🎉

parley_tests/README.md
```diff
@@ -1,7 +1,17 @@
-# Usage
+# parley_tests
+
+Integration test suite for the Parley text layout library.
+
+## Overview
+
+This crate contains the integration test suite for `parley`. It follows a similar
+architecture to `vello_sparse_tests` and uses snapshot testing to validate layout
```
taj-p:
```suggestion
architecture to [`vello_sparse_tests`](https://github.com/linebender/vello/tree/main/sparse_strips/vello_sparse_tests) and uses snapshot testing to validate layout
```

parley_tests/Cargo.toml
valadaptive: Reworded the TODO.

parley_tests/tests/mod.rs
valadaptive: Mostly copied from `vello_sparse_tests` :sweat_smile:

PR #528, opened 2026-01-28
Convert `Glyph::y` to Y-down coordinate space

valadaptive:
The glyph `y` field uses font coordinates (Y-up). However, basically everything else (including a run's `baseline`) is in layout coordinates (Y-down).

All the existing examples except for `vello_cpu_render` use the un-positioned `glyphs` method, subtracting `glyph.y` instead of adding it. In essence, they are applying the layout-space transform themselves. However, the `positioned_glyphs` method (used by `vello_cpu_render`) does *not* perform this conversion, and returns nonsensical results (it's adding a Y-down baseline to a Y-up glyph coordinate).

One option (and the one I considered initially) was to flip the sign of `glyph.y` only within `positioned_glyphs`. However, `glyphs` and `positioned_glyphs` *both* return an iterator of `Glyph`s. It'd be really confusing if `Glyph::y` was Y-up when returned from `glyphs`, but Y-down when returned from `positioned_glyphs`.

I think the simpler option is to just flip `Glyph::y` to Y-down at construction time. This makes everything consistent, and allows API consumers to treat it the same way as `Glyph::x`. Right now, they have to know that they should subtract it instead.

This shouldn't be a huge breaking change, since it seems most glyphs have a Y of 0. In fact, it's likely that many API consumers are getting the coordinate transform wrong right now. I only noticed this when migrating the test suite to `parley_draw`, where some Arabic combining marks are incorrectly positioned because of this issue:

image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,30 @@ This release has an [MSRV] of 1.88.
 
 - Breaking change: `Glyph::y` is now in Y-down coordinate space instead of Y-up coordinate space. ([#528][] by [@valadaptive][])
 
+  **This does not change the API surface, but *will* change the behavior of existing code!** If you're iterating over non-positioned glyphs using the `GlyphRun::glyphs` method and positioning each glyph yourself, with code like:
+
+  ```rust
+  let run_y = glyph_run.baseline();
+  for glyph in glyph_run.glyphs() {
+    let glyph_y = run_y - glyph.y;
+    ...
+  }
+  ```
+
+  You'll need to update your code to *add* `glyph.y`, instead of subtracting it:
+
+  ```rust
+  let run_y = glyph_run.baseline();
+  for glyph in glyph_run.glyphs() {
+    let glyph_y = run_y + glyph.y;
+    ...
+  }
+  ```
+
+  If you instead use the `GlyphRun::positioned_glyphs` method, your code will not need to change.
+
+
+
 ## [0.7.0] - 2025-11-24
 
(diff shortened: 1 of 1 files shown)


CHANGELOG.md
```diff
@@ -15,6 +15,12 @@ You can find its changes [documented below](#060---2025-11-24).
 
 This release has an [MSRV] of 1.88.
 
+### Changed
+
+#### Parley
+
+- Breaking change: `Glyph::y` is now in Y-down coordinate space instead of Y-up coordinate space. ([#528][] by [@valadaptive][])
```
taj-p, line 22:
```suggestion
- ⚠️ BREAKING CHANGE ⚠️ : `Glyph::y` is now in Y-down coordinate space instead of Y-up coordinate space. ([#528][] by [@valadaptive][])
   - <TODO: Add short documentation on migration process>

```

Would it be possible to change the log to something like this?

nicoburns: Probably best to avoid emojis lest we trigger the anti-AI brigade?

conor-93: This would diverge from existing "Breaking change" entries in the changelog, so if we do this, let's update all entries for consistency. I think we can probably get away with emojis if they're not all over the codebase re: Nico's concern, but `BREAKING CHANGE` would be fine too.

valadaptive: I added a bit more to the changelog entry, warning people that they may need to update their code even though the API surface is unchanged. I think it should be sufficient; take a look.

PR #531, opened 2026-01-30
Fix max-content width being reset by mandatory line breaks

nicoburns:
When a mandatory line break occurs we should set the `running_max_width` to 0 and start counting `running_max_width` again. However, we should first accumulate the `running_max_width` into the overall `max_width`. Otherwise the width of the line we have just ended is not accounted for.

This was causing text in Blitz to wrap to the min-content width in cases where a paragraph ended with a newline as the max-content width was entirely zeroed out!

### DIFF
--- a/parley_tests/tests/issues.rs
+++ b/parley_tests/tests/issues.rs
@@ -98,7 +98,7 @@ And, finally, yet another sentence."#;
 fn max_context_with_mandatory_breaks() {
     let mut env = TestEnv::new(test_name!(), None);
 
-let text = "First line of text.
+    let text = "First line of text.
 Second line of text.
 Third line that ends with newlines\n\n";
 


parley_tests/tests/issues.rs
```diff
@@ -83,3 +83,34 @@ And, finally, yet another sentence."#;
...
+/// This was causing text in Blitz to wrap to the min-content width in cases
+/// where a paragraph ended with a newline as the max-content width was
+/// entirely zeroed out!
+#[test]
+fn max_context_with_mandatory_breaks() {
+    let mut env = TestEnv::new(test_name!(), None);
+
+    let text = "A sentence across two lines.
+And another sentence that breaks across, hopefully, three lines.
+And, finally, yet another sentence that ends with newlines\n\n";
```
taj-p:
```suggestion
let text = "First line of text.
Second line of text.
Third line that ends with newlines\n\n";
```

Could we do something like this instead - the text content is a little confusing because I believe we actually don't want any of these sentences to suffer mid-sentence breaks

PR #533, opened 2026-02-01
Add API for setting complex segmentation data

valadaptive:
This PR adds an API to Parley for loading locale-specific text analysis data at runtime: `LayoutContext::load_segmenter_models_auto` and `LayoutContext::load_segmenter_models_dictionary`.

The data itself is now generated by `parley_data_gen`, and exported as `static` byte slices from the `parley_data` crate if the `bundled-segmenter-models` feature is enabled. The intent is that if you're deploying a desktop application, you can just pass the models in from `parley_data` directly, but if you're deploying to the web, you can depend on `parley_data` at *build* time and export the model data into files to be dynamically loaded.

There are some required supporting changes:
- `parley_data` now generates dictionary providers too, even though the baked data doesn't provide any dictionaries and just errors out. This matches how it works with LSTMs.

- I noticed that the code becomes a lot cleaner, especially with some new abstractions, if `LineSegmenters` doesn't have to borrow itself mutably and uses `OnceCell` instead of `Option`.

  Right now, to juggle lifetimes properly, `analyze_text` moves the `LineSegmenters` struct *out* of the parent `LayoutContext`, and then moves it back *in* after analysis. I noticed that the `LineSegmenters` struct is 5KB large(!) so getting rid of those moves probably helps with performance too.

There's an opportunity for future improvement: right now, the data has to be loaded from a `&'static [u8]` or `Box<[u8]>`, since that's the interface that [icu_provider_blob exposes](https://docs.rs/icu_provider_blob/latest/icu_provider_blob/struct.BlobDataProvider.html). ~~We *could* change the API to use [linebender_resource_handle::Blob](https://github.com/linebender/raw_resource_handle), but it requires that type to be compatible with Yoke. [I did implement this](https://github.com/linebender/raw_resource_handle/pull/14), but it's not yet made it into a release.~~ This may be a bit trickier; icu_provider_blob has a private API that it doesn't expose...

`SegmenterModelData::from_blob` returns an `icu_provider::DataError`; not sure if we want to "leak" that error type publicly. Parley, as far as I can tell, does not define *any* error types of its own yet. Would an unwrap/expect be better here?


parley/src/analysis/mod.rs
```diff
@@ -36,44 +104,91 @@ pub(crate) struct AnalysisDataSources {
...
 }
 
 #[derive(Default)]
 struct LineSegmenters {
-    normal: Option<LineSegmenter>,
-    keep_all: Option<LineSegmenter>,
-    break_all: Option<LineSegmenter>,
+    normal: OnceCell<LineSegmenter>,
+    keep_all: OnceCell<LineSegmenter>,
+    break_all: OnceCell<LineSegmenter>,
```
conor-93: > getting rid of those moves probably helps with performance too.

I'm seeing a nice performance improvement on this PR!
```
Default Style - arabic 20 characters               [  10.5 us ...  10.1 us ]      -3.50%*
Default Style - latin 20 characters                [   4.7 us ...   4.5 us ]      -4.19%*
Default Style - japanese 20 characters             [   9.5 us ...   9.2 us ]      -3.94%*
Default Style - arabic 1 paragraph                 [  58.3 us ...  57.9 us ]      -0.58%
Default Style - latin 1 paragraph                  [  19.5 us ...  19.2 us ]      -1.63%
Default Style - japanese 1 paragraph               [  86.6 us ...  85.1 us ]      -1.69%*
Default Style - arabic 4 paragraph                 [ 245.2 us ... 245.4 us ]      +0.07%
Default Style - latin 4 paragraph                  [  72.4 us ...  72.4 us ]      +0.03%
Default Style - japanese 4 paragraph               [ 115.8 us ... 114.6 us ]      -0.99%
Styled - arabic 20 characters                      [  11.4 us ...  11.1 us ]      -2.48%*
Styled - latin 20 characters                       [   6.2 us ...   5.9 us ]      -5.24%*
Styled - japanese 20 characters                    [  10.6 us ...  10.1 us ]      -4.33%*
Styled - arabic 1 paragraph                        [  61.7 us ...  60.8 us ]      -1.48%
Styled - latin 1 paragraph                         [  23.6 us ...  23.0 us ]      -2.27%*
Styled - japanese 1 paragraph                      [  91.0 us ...  89.8 us ]      -1.27%
Styled - arabic 4 paragraph                        [ 264.5 us ... 260.1 us ]      -1.68%
Styled - latin 4 paragraph                         [ 122.0 us ... 122.8 us ]      +0.68%
Styled - japanese 4 paragraph                      [ 171.5 us ... 168.7 us ]      -1.66%
```

parley/src/context.rs
```diff
@@ -53,6 +53,73 @@ impl<B: Brush> LayoutContext<B> {
...
+        self.analysis_data_sources.load_segmenter_models(
+            models.into_iter().map(|model| model.provider).collect(),
+            crate::analysis::SegmenterMode::Auto,
+        );
+    }
+
+    /// Loads runtime dictionary segmenter models for word/line breaking.
+    ///
+    /// Unlike [`Self::load_segmenter_models_auto`] which uses LSTM for Southeast Asian scripts, this uses dictionary
+    /// data for all complex scripts. Dictionary models may be faster at runtime but require more data.
```
conor-93: super nit: I think it's more correct to refer to these as just 'dictionaries', they're not models in the sense that LSTMs are, more just lookup tables of known words. There's a few other usages of this phrasing this PR, too.
```suggestion
    /// Unlike [`Self::load_segmenter_models_auto`] which uses LSTM for Southeast Asian scripts, this uses dictionary
    /// data for all complex scripts. Dictionaries may be faster at runtime but require more data.
```

conor-93 (approved):
Wow, fantastic work!

parley_data/src/lib.rs
taj-p, line 1: This is great! I've yet to take a close look at the PR - I have time to do that tomorrow. But, one question I have is whether we're still planning on allowing the extraction of all unicode data from the binary so that it can be provided at runtime

valadaptive: > But, one question I have is whether we're still planning on allowing the extraction of all unicode data from the binary so that it can be provided at runtime

I can move in that direction if it turns out to be necessary, but the existing data (that is, everything but the LSTMs and dictionaries) is pretty small. The *source* files are ~50KB compressed. Moving those out of the library would require them to be provided before *any* text can be laid out, and might slow things down (the baked data is regular compiler-optimizable code; dynamically-loaded data would not be).

The tradeoff would become even worse if we decide to switch to Packtab, which (IIRC) produces smaller tables and faster lookups, but requires the compiler to be able to optimize (and inline!) the exact series of shifts/multiplies/etc used in each lookup.

parley/src/context.rs
valadaptive: Reworded the documentation a bit.

parley_tests/snapshots/mixed_script_line_break_with_no_models-0.png
taj-p, line 1: Did you mean to add a burmese font so we don't need to render tofu glyphs? 

I guess it's clear that the with models version is segmenting differently!

parley_tests/snapshots/mixed_script_line_break_with_multiple_models-0.png
taj-p, line 1: This looks correct per me playing around in Google. Does the line segmentation for the thai text look correct to you @waywardmonkeys ?

parley/src/analysis/mod.rs
```diff
@@ -84,14 +199,67 @@ impl AnalysisDataSources {
             script_short_name: PropertyNamesShort::<Script>::try_new_unstable(&PROVIDER).unwrap(),
             brackets: CodePointMapData::<BidiMirroringGlyph>::try_new_unstable(&PROVIDER).unwrap(),
             composite: CompositeProps,
+            #[cfg(feature = "runtime-segmenter-data")]
+            runtime_buffer_provider: None,
         }
     }
 
+    #[cfg(feature = "runtime-segmenter-data")]
+    pub(crate) fn load_segmenter_models(
```
taj-p, line 212: @robertbastian - I was wondering if you would be able to give this code a once over from the perspective of an ICU4X maintainer 🙏

parley/src/analysis/mod.rs
taj-p, line 1: How does a consumer know whether they should load in an external segmenter to correctly lay out some input text? 

If I understand the API currently, it requires the consumer to decide upfront - either via pre-scanning text prior to Parley or making assumptions based on the type of user and where they may be located.

Since there are only a small size of complex scripts needing this attention, I wonder whether we can use a bitset over a single byte to track this during analysis and provide it back to the caller in `Layout`.

parley/src/analysis/mod.rs
taj-p, line 212: My understanding is that, since this is a replace operation, consumers cannot easily append models. For example, if a user starts with latin, then adds Thai, then a few minutes later adds Burmese text, when we go to upload the Burmese text, we'll need to re-prepare the Thai model to `load_segmenter_models([thai, burmese])`.

I'm unsure whether we should support an `append` API and what that might look like. If it's too complex, happy to proceed as is and look at this another time.

parley_data/src/lib.rs
taj-p: > might slow things down (the baked data is regular compiler-optimizable code; dynamically-loaded data would not be)

I think `CodePointTrie` construction might be slower, but I'm not sure whether it would effect lookup. 

---

Nice! Yes, I validated that finding. It's about 50 kB gzipped (12% of total binary) and 240 kB uncompressed (17% of total binary) if the binary just lays out some text in Wasm:

```rs
use parley::{FontContext, LayoutContext, StyleProperty, GenericFamily, Layout, Alignment, AlignmentOptions};

#[derive(Clone, Copy, Default, Debug, PartialEq)]
struct Brush;

#[no_mangle]
pub extern "C" fn create_layout() -> usize {
    
    let text = "Hello World!";
    builder.push_default(GenericFamily::SansSerif);
    
    let mut layout: Layout<Brush> = builder.build(text);
    layout.break_all_lines(Some(200.0));
    layout.align(Some(200.0), Alignment::Start, AlignmentOptions::default());
    
    layout.width() as usize
}
``` 

For me, that represents 6% of my total binary quota 😅 . It's not nothing, but I agree it's probably not worth bikeshedding at this stage. I think the first port of call might be to enable ICU data sharing with HarfRust and if I had to choose between this and general Parley latency performance, I would choose the latter at this time.

parley_tests/snapshots/mixed_script_line_break_with_no_models-0.png
valadaptive: This test was added a bit later than the Thai one; I think a Burmese font wouldn't hurt.

parley/src/analysis/mod.rs
valadaptive: This might take a bit more work; I'll need to figure out how ICU4X itself determines characters' locales. Not sure if it's a matter of asking each individual model about the character in turn, or if there's an optimization it uses here.

parley/src/analysis/mod.rs
valadaptive: I assumed ICU4X treated things immutably, but it turns out there *is* a `push` method on `MultiForkByErrorProvider`. I've added an `append` API.

parley/src/analysis/mod.rs
robertbastian: ICU4X uses complex data if a code point has the [`SA` property](https://www.unicode.org/reports/tr14/#SA). Inside that code path we use the table from the spec to decide which model to use.

I don't think you or your callers should pre-scan the text to determine if complex breaking is required. If you do that, you need to have the data available *anyway*. Loading the data is very cheap (if you use compiled data, not what this PR is doing), *having* the data is the problem that some users might want to avoid.

parley_data_gen/src/lib.rs
robertbastian, line 1: issue: you're generating data without customizing it. this way you pay the cost of deserializing your data at runtime, with no benefit compared to compiled data

you should merge at least the first commit of https://github.com/linebender/parley/pull/482 before doing any more work with ICU4X data

robertbastian (commented):
From an ICU4X perspective this is overly complicated. I don't see a reason to use a buffer provider which has a deserialization cost at runtime. You can also hide the compiled data behind a Cargo feature.

https://github.com/linebender/parley/pull/482 should be merged before any more data provider work.

valadaptive:
> From an ICU4X perspective this is overly complicated. I don't see a reason to use a buffer provider which has a deserialization cost at runtime. You can also hide the compiled data behind a Cargo feature.

I believe the reason to separate the data is for web use cases. If the code changes more frequently than the ICU4X data, you can ship the two separately and serve the latter with a much longer cache life. That's what I understood from what @taj-p said, anyway.

parley/src/analysis/mod.rs
```diff
@@ -36,44 +104,91 @@ pub(crate) struct AnalysisDataSources {
...
+    #[cfg(feature = "runtime-segmenter-data")]
+    runtime_buffer_provider: Option<RuntimeBufferProvider>,
 }
 
 #[derive(Default)]
 struct LineSegmenters {
-    normal: Option<LineSegmenter>,
-    keep_all: Option<LineSegmenter>,
-    break_all: Option<LineSegmenter>,
+    normal: OnceCell<LineSegmenter>,
```
robertbastian: Introducing a `OnceLock` makes something that can be verified at compile time a runtime mutex lock. The move can be avoided differently, see #537

(edit: and with `OnceCell` you lose `Sync`, as you noticed)

parley/src/analysis/provider.rs
```diff
@@ -26,6 +26,8 @@ macro_rules! impl_data_provider {
         impl_segmenter_break_grapheme_cluster_v1!($provider);
         impl_segmenter_break_line_v1!($provider);
         impl_normalizer_nfc_v1!($provider);
+        impl_segmenter_dictionary_auto_v1!($provider);
+        impl_segmenter_dictionary_extended_v1!($provider);
```
robertbastian: you're adding erroring implementations that you only use with a `ForkByMarkerProvider`. this should work without these

parley/src/analysis/provider.rs
```diff
@@ -26,6 +26,8 @@ macro_rules! impl_data_provider {
         impl_segmenter_break_grapheme_cluster_v1!($provider);
         impl_segmenter_break_line_v1!($provider);
         impl_normalizer_nfc_v1!($provider);
+        impl_segmenter_dictionary_auto_v1!($provider);
```
robertbastian, line 29: these three contain no data. if you really need them (for bounds somehwere), consider not generating them and just adding empty `DataProvider` implementations

parley_data_gen/src/lib.rs
```diff
@@ -129,6 +132,50 @@ pub fn generate(out: std::path::PathBuf) {
...
+                LocaleFallbacker::new_without_data(),
+            )
+            .with_markers([
+                icu_segmenter::provider::SegmenterLstmAutoV1::INFO,
+                icu_segmenter::provider::SegmenterDictionaryAutoV1::INFO,
+                // SegmenterDictionaryAutoV1 only includes dictionary data for scripts with no LSTM model
+                icu_segmenter::provider::SegmenterDictionaryExtendedV1::INFO,
+                icu_segmenter::provider::SegmenterBreakWordV1::INFO,
+                icu_segmenter::provider::SegmenterBreakWordOverrideV1::INFO,
+                icu_segmenter::provider::SegmenterBreakLineV1::INFO,
```
robertbastian, line 166: you shouldn't need to add these three markers to the postcard data

parley_data_gen/src/lib.rs
```diff
@@ -129,6 +132,50 @@ pub fn generate(out: std::path::PathBuf) {
...
+            ])
+            .with_segmenter_models([(*model).to_string()])
+            .export(
+                &icu4x_source_provider,
+                BlobExporter::new_with_sink(Box::new(&mut blob)),
```
robertbastian: consider writing directly to a file

```suggestion
                BlobExporter::new_with_sink(Box::new(&mut BufWriter::new(File::create(&blob_path).unwrap()))),
```

parley_data/src/lib.rs
```diff
@@ -169,3 +177,92 @@ impl From<Properties> for u32 {
...
+/// Bundled segmenter models for languages requiring complex word breaking.
+///
+/// These are pre-generated postcard blobs that can be loaded at runtime.
+///
+/// You can also depend on `parley_data` in your build, and export these models to files that you can ship alongside
+/// your Parley-using application and load dynamically.
+#[cfg(feature = "bundled-segmenter-models")]
+pub mod bundled_models {
+    /// Thai LSTM model for word/line segmentation.
+    pub const THAI_LSTM: &[u8] = include_bytes!(
```
robertbastian, line 189: consider documenting the sizes of these so clients can make more informed decisions

parley_data/src/lib.rs
```diff
@@ -169,3 +177,92 @@ impl From<Properties> for u32 {
         value.0
     }
 }
+
+/// Bundled segmenter models for languages requiring complex word breaking.
+///
+/// These are pre-generated postcard blobs that can be loaded at runtime.
+///
+/// You can also depend on `parley_data` in your build, and export these models to files that you can ship alongside
```
robertbastian, line 185: I would just reexport this module in `parley`, so clients don't have to depend on `parley_data`

parley_data/Cargo.toml
```diff
@@ -12,6 +12,7 @@ publish = false
 default = ["baked"]
 baked = ["dep:zerovec"]
 datagen = []
+bundled-segmenter-models = []
```
robertbastian: what's the point of hiding these behind a feature?

parley/src/analysis/mod.rs
```diff
@@ -36,44 +104,91 @@ pub(crate) struct AnalysisDataSources {
...
+                        SegmenterMode::Dictionary => {
+                            LineSegmenter::try_new_dictionary_unstable(&combined, line_break_opts)
+                        }
+                    }
+                } else {
+                    LineSegmenter::try_new_auto_unstable(&PROVIDER, line_break_opts)
+                }
+                .expect("Failed to create LineSegmenter")
+            })
+            .as_borrowed()
```
robertbastian: you don't need to duplicate the whole function

```rust
                let line_break_opts = Self::make_line_break_opts(word_break_strength);
                #[cfg(feature = "runtime-segmenter-data")]
                if let Some(&RuntimeBufferProvider {
                    ref provider,
                    segmenter_mode,
                }) = runtime_buffer_provider
                {
                    let combined =
                        ForkByMarkerProvider::new(provider.as_deserializing(), &PROVIDER);
                    return match segmenter_mode {
                        SegmenterMode::Auto => {
                            LineSegmenter::try_new_auto_unstable(&combined, line_break_opts)
                        }
                        SegmenterMode::Dictionary => {
                            LineSegmenter::try_new_dictionary_unstable(&combined, line_break_opts)
                        }
                    }.expect("Failed to create LineSegmenter");
                }
                LineSegmenter::try_new_auto_unstable(&PROVIDER, line_break_opts)
                    .expect("Failed to create LineSegmenter")
```

parley/src/analysis/mod.rs
```diff
@@ -84,6 +199,73 @@ impl AnalysisDataSources {
...
+            SegmenterMode::Auto => {
+                WordSegmenter::try_new_auto_unstable(&combined, WordBreakOptions::default())
+            }
+            SegmenterMode::Dictionary => {
+                WordSegmenter::try_new_dictionary_unstable(&combined, WordBreakOptions::default())
+            }
+        }
+        .expect("Failed to create WordSegmenter with runtime models");
+
+        // Clear cached line segmenters; they will be lazily recreated with the new mode.
```
robertbastian, line 207: why not do the same for all segmenters?

parley/src/analysis/mod.rs
```diff
@@ -36,44 +104,91 @@ pub(crate) struct AnalysisDataSources {
...
+                    match segmenter_mode {
+                        SegmenterMode::Auto => {
+                            LineSegmenter::try_new_auto_unstable(&combined, line_break_opts)
+                        }
+                        SegmenterMode::Dictionary => {
+                            LineSegmenter::try_new_dictionary_unstable(&combined, line_break_opts)
+                        }
+                    }
+                } else {
+                    LineSegmenter::try_new_auto_unstable(&PROVIDER, line_break_opts)
```
robertbastian: this is why I'd really like to merge #482 before this

```suggestion
                    LineSegmenter::try_new_for_non_complex_scripts(&PROVIDER, line_break_opts)
```

parley/src/analysis/mod.rs
```diff
@@ -36,44 +104,91 @@ pub(crate) struct AnalysisDataSources {
...
+        line_break_opts
+    }
 }
 
 impl AnalysisDataSources {
     pub(crate) fn new() -> Self {
         Self {
             grapheme_segmenter: GraphemeClusterSegmenter::try_new_unstable(&PROVIDER).unwrap(),
-            word_segmenter: WordSegmenter::try_new_lstm_unstable(
+            word_segmenter: WordSegmenter::try_new_auto_unstable(
```
robertbastian:
```suggestion
            word_segmenter: WordSegmenter::try_new_for_non_complex_scripts(
```

parley/src/analysis/mod.rs
```diff
@@ -84,6 +199,73 @@ impl AnalysisDataSources {
...
+        .expect("Failed to create WordSegmenter with runtime models");
+
+        // Clear cached line segmenters; they will be lazily recreated with the new mode.
+        self.line_segmenters = LineSegmenters::default();
+    }
+
+    #[cfg(feature = "runtime-segmenter-data")]
+    pub(crate) fn load_segmenter_models(
+        &mut self,
+        providers: Vec<icu_provider_blob::BlobDataProvider>,
```
robertbastian, line 214: I'd move the `.provider` and `collect()` here, to encapsulate the logic more. then the `SegmenterModelData.provider` field can be private, and the fact that you collect into a `Vec` as well
```suggestion
        models: impl IntoIterator<Item = SegmenterModelData>,
```

parley/src/analysis/mod.rs
```diff
@@ -84,6 +199,73 @@ impl AnalysisDataSources {
...
+        provider: icu_provider_blob::BlobDataProvider,
+        mode: SegmenterMode,
+    ) {
+        match self.runtime_buffer_provider.as_mut() {
+            None => {
+                self.load_segmenter_models(alloc::vec![provider], mode);
+            }
+            Some(buffer_provider) => {
+                let cur_mode = buffer_provider.segmenter_mode;
+                assert_eq!(
```
robertbastian, line 242: this is an undocumented panic. the public append functions only talk about performance issues with duplicated models

I'm not sold on the use case of the append methods. when would a client not supply the models all at once?

parley/src/analysis/provider.rs
valadaptive: They are necessary for bounds. Since `impl_segmenter_lstm_auto_v1` is already empty and present before this PR, could I just make that change separately and land it beforehand?

parley_data_gen/src/lib.rs
valadaptive: This is necessary for trait bounds, otherwise I get errors like:

```
error[E0277]: the trait bound `BakedProvider: DataProvider<SegmenterDictionaryAutoV1>` is not satisfied
   --> parley/src/analysis/mod.rs:217:54
    |
217 |                 WordSegmenter::try_new_auto_unstable(&combined, WordBreakOptions::default())
    |                 ------------------------------------ ^^^^^^^^^ unsatisfied trait bound
    |                 |
    |                 required by a bound introduced by this call
    |
```

parley_data/src/lib.rs
valadaptive: Are their sizes expected to change? While exporting them from parley_data means they won't change without explicit regeneration, I'm worried the docs might drift out of sync with the actual data.

parley_data/Cargo.toml
valadaptive: I think I was concerned that they wouldn't be DCE'd? Probably not a big deal.

parley/src/analysis/mod.rs
valadaptive: The `RuntimeBufferProvider` struct contains types from `icu_provider_adapters` and `icu_provider_blob`, which are not added as dependencies unless the `runtime-segmenter-data` feature is enabled. That means we can't name the `RuntimeBufferProvider` type (and hence, cannot take it as a parameter even *without* using it) unless that feature is enabled.

There might be a better way to do this; maybe it'd be better to just `#[cfg]` out the *fields* on `RuntimeBufferProvider` if the feature is disabled?

parley_data/src/lib.rs
robertbastian: they might. you could generate the whole `bundled_models` module in `parley_data_gen`

parley/src/analysis/mod.rs
valadaptive: I believe the line segmenters and word segmenter are the only ones that use complex script data--correct me if I'm wrong. We eagerly reinitialize the word segmenter, and lazily reinitialize the line segmenters.

parley/src/analysis/mod.rs
robertbastian: just `cfg` the whole parameter

parley/src/analysis/mod.rs
valadaptive: The reason I collected into a `Vec` in the public methods on `Layout` is to avoid monomorphizing this function, which contains a lot more actual logic.

parley/src/analysis/mod.rs
robertbastian: > I believe the line segmenters and word segmenter are the only ones that use complex script data

yes. I meant why not treat the four segmenters that use complex script data the same

robertbastian: you should really do it eagerly, because if you have clients passing you arbitrary data, the constructors can fail and you want to return an error instead of panicking.

the non-custom constructors (`try_new_for_non_complex_scripts`) are infallible, so you can create these in `AnalysisDataSources::new()`. however the `unstable` constructors are pretty fallible with user-supplied data, so you should eagerly construct them in `load_segmenter_models` to return errors to the caller

parley/src/analysis/mod.rs
valadaptive: cc @taj-p

parley_data/src/lib.rs
```diff
@@ -184,58 +184,57 @@ impl From<Properties> for u32 {
 ///
 /// You can also depend on `parley_data` in your build, and export these models to files that you can ship alongside
 /// your Parley-using application and load dynamically.
-#[cfg(feature = "bundled-segmenter-models")]
 pub mod bundled_models {
     /// Thai LSTM model for word/line segmentation.
-    pub const THAI_LSTM: &[u8] = include_bytes!(
+    pub static THAI_LSTM: &[u8] = include_bytes!(
```
robertbastian: uhm I think const was better. it's a const reference, so while the reference might be copied around, the thing behind it (`'static`) won't be

parley/src/analysis/provider.rs
robertbastian: well, #482 does this

parley_data_gen/src/lib.rs
robertbastian: this error cannot be due to that. the compiler does not know which markers are in a postcard file

robertbastian:
I put some suggestions in https://github.com/robertbastian/parley/commit/b1c6c2821a95d2eacc37afc0e47169548115e123 including:
* propagating errors to the caller, which requires eager construction
* removing the append logic, so we don't have to store the provider anymore

parley/src/analysis/mod.rs
conor-93: > when would a client not supply the models all at once?

For our own web app, this would give us an opportunity to load in models as the user adds to their content, mostly to minimise memory usage (one of our leading concerns).

robertbastian: Even then that should be done through the other constructor. If you want to add two new models, calling the append function twice does unnecessary work.

PR #539, opened 2026-02-05
Add text-drawing benchmarks

valadaptive:
Adds a few different benchmark variants, to test the performance impact of underline drawing and caching.

For some reason, on my machine (Linux) I need to pass the `-p` (parallel) flag to tango when performing a comparison. Without it, the "new" code always comes back as ridiculously faster, even when it's not been changed at all. I can't find any information on why this would be.

### DIFF
--- a/parley_bench/src/draw.rs
+++ b/parley_bench/src/draw.rs
@@ -3,7 +3,7 @@
 
 //! # Draw Benchmarks
 //!
-//! Benchmarks for text rendering using parley_draw with vello_cpu.
+//! Benchmarks for text rendering using `parley_draw` with `vello_cpu`.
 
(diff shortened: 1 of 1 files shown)


valadaptive:
> I'm anticipating that we'll want to move `parley_draw` out of this repo at some point.

OK, what's the deal with this? I haven't been able to find any relevant Zulip threads about what on earth is actually in-scope for `parley_draw`. Is it just glue code for various renderer frontends? If we move it into its own repository, why can't we just move these benchmarks there too? Given that testing parley_draw will require a backend to drive text layout, won't the parley_draw test suite depend on parley proper?

I guess I just don't understand the motivation for *why* we want to move `parley_draw` into its own repo. What do we gain?

PR #541, opened 2026-02-06
Add method to convert non-shared `Collection`/`SourceCache` to shared variants

nicoburns:
This PR allows a `Collection` or `SourceCache` that was constructed as non-shared, to be later converted to be shared. This is useful when accepting `FontContext`/`Collection`/`SourceCache` provided by users (which may or may not be shared - as decided by the user) which you then want to "fan out" into thread-local shared clones.

Blitz is using the new `make_shared` method on `SourceCache` introduced in this PR. Blitz is not currently using the new corresponding `make_shared` method on `Collection`, but that method is included in this PR anyway for symmetry.

### DIFF
--- a/fontique/src/collection/mod.rs
+++ b/fontique/src/collection/mod.rs
@@ -829,7 +829,10 @@ impl Shared {
 }
 
 #[test]
-#[cfg(feature = "std")]
+#[cfg(all(
+    feature = "std",
+    any(target_os = "linux", target_os = "macos", target_os = "windows")
+))]
 fn make_shared_matches_local() {
     use crate::{Collection, CollectionOptions};
 


fontique/src/collection/mod.rs
```diff
@@ -254,6 +260,16 @@ impl Inner {
         }
     }
 
+    #[cfg(feature = "std")]
+    pub fn make_shared(&mut self) {
+        if self.shared.is_none() {
+            self.shared = Some(Arc::new(Shared {
+                data: Mutex::new(core::mem::take(&mut self.data)),
```
taj-p: I think there is a bug here. Do we need to increment the `self.shared_version` too or clone instead of take? Otherwise, local data seems to be lost on subsequent reads.

For example, this example code fails on the assert because all our reads are guarded through `sync_shared`. Since the version hasn't changed, we don't update our `self.data`.


```rs
use parley::fontique::{Collection, CollectionOptions};

fn main() {
    let mut collection = Collection::new(CollectionOptions {
        shared: false,
        system_fonts: false,
    });

    let font_dirs: Vec<std::path::PathBuf> = ["/System/Library/Fonts", "/Library/Fonts"]
        .iter()
        .map(std::path::PathBuf::from)
        .filter(|p| p.is_dir())
        .collect();

    if font_dirs.is_empty() {
        eprintln!("No system font directories found — cannot demonstrate the bug.");
        std::process::exit(1);
    }

    collection.load_fonts_from_paths(&font_dirs);

    let names_before: Vec<String> = collection.family_names().map(String::from).collect();

    collection.make_shared();

    let names_after: Vec<String> = collection.family_names().map(String::from).collect();

    assert_eq!(names_before.len(), names_after.len());
}
```

Image

PR #548, opened 2026-02-13
feat: Add glyph atlas cache with LRU eviction for vello spare strips renderers

grebmeg:
Introduces a glyph atlas cache that rasterizes glyphs once and reuses the bitmaps on subsequent frames, reducing redundant work for text-heavy scenes. The cache uses LRU age-based eviction, deterministic hashing for reproducible atlas packing, subpixel-quantized cache keys (supporting variable fonts and COLR glyphs), and a multi-page atlas backed by `vello_common::ImageCache`.

<p align="center">
vello_cpu_render
</p>

### DIFF
--- a/parley_tests/tests/draw.rs
+++ b/parley_tests/tests/draw.rs
@@ -119,7 +119,22 @@ fn draw_underline_descenders() {
 }
 
 /// Test bitmap (CBTF) emoji rendering across different hinting, per-glyph transform, and scale configurations.
+///
+/// Disabled: snapshots were generated on macOS (ARM/NEON) but fail on
+/// x86-64 (Ubuntu).
+///
+/// Proven:
+/// - PNG roundtrip (`into_png` → `from_png`) is lossless — 0 differing
+///   pixels when round-tripping the current render.
+///
+/// Hypothesis (not yet proven at the SIMD level):
+/// - The vello CPU renderer's bicubic image filter (`FilteredImagePainter<2>`)
+///   produces different edge-pixel values on ARM/NEON vs x86-64 (AVX2/SSE4.2)
+///   due to floating-point differences in the SIMD paths.
+///
+/// Snapshots need to be regenerated per platform.
 #[test]
+#[ignore]
 fn draw_bitmap_emoji() {
     let mut env = TestEnv::new(test_name!(), None);
     env.set_tolerance(5.0);


grebmeg:
Some tests are failing due to minor differences, I skimmed through the Kompari HTML report, and they’re very likely caused by recent changes in Vello that slightly affect the rendered output.

.github/workflows/ci.yml
```diff
@@ -13,7 +13,7 @@ env:
   # This should be limited to packages that are intended for publishing.
   RUST_MIN_VER_PKGS: "-p parley -p parley_core -p parley_draw -p fontique -p attributed_text"
   # List of features that depend on the standard library and will be excluded from no_std checks.
-  FEATURES_DEPENDING_ON_STD: "std,default,png,system"
+  FEATURES_DEPENDING_ON_STD: "std,default,png,system,vello_hybrid"
```
LaurenzV, line 16: This is because of guillotiere?

examples/common/src/lib.rs
```diff
@@ -0,0 +1,433 @@
...
+        Self {
+            text: "".to_owned(),
+            display_scale: 1.0,
+            quantize: true,
+            max_advance: Some(200.0),
+            foreground_color: Color::BLACK,
+            background_color: Color::WHITE,
+            padding: 20,
+            hint: true,
+            use_atlas_cache: true,
```
LaurenzV, line 74: If I set this to false I get a panic, is that intended?

```
=== Phase 1: Cold start — simple layout (populate cache) ===
  cache: len=0 hits=0 misses=0

=== Phase 2: Warm cache — simple layout (all hits) ===
  cache: len=0 hits=0 misses=0

=== Phase 3a: Rich layout (partial overlap, round 1) ===

thread 'main' (3649098) panicked at /Users/laurenz/.cargo/git/checkouts/vello-348d5f1672e79ca5/855760a/sparse_strips/vello_hybrid/src/schedule.rs:1073:30:
not implemented: Unsupported image source
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

examples/common/src/lib.rs
```diff
@@ -0,0 +1,433 @@
...
+    let mut path = PathBuf::from(crate_file);
+    if let Ok(canon) = std::fs::canonicalize(&path) {
+        path = canon;
+    }
+    for _ in 0..3 {
+        path.pop();
+    }
+    path.push("_output");
+    let _ = std::fs::create_dir(&path);
+    path
```
LaurenzV: Would it not be easier to use CARGO_MANIFEST_DIR somehow and then just append the folder chain? Not sure if I'm missing something.

examples/common/src/lib.rs
```diff
@@ -0,0 +1,433 @@
...
+    /// Disable accumulation mode and record the total accumulated duration.
+    pub fn finish_accumulating(&mut self, label: impl Into<String>) {
+        let label = label.into();
+        if let Some(total) = self.accumulators.remove(&label) {
+            self.timings.entry(label).or_default().push(total);
+        }
+    }
+
+    /// Print all collected timings, split into one-time and per-frame groups.
+    pub fn print_summary(&self) {
```
LaurenzV, line 371: Am I understanding correctly that we have a number of different frames which exercise different paths (different layouts, different caching behavior, etc.), and then we are summarizing the number of all frames across the different stages? If yes, isn't that a bit meaningless since each frame tests something different? So the numbers will just always vary wildly.

examples/vello_hybrid_render/src/main.rs
```diff
@@ -0,0 +1,653 @@
...
+    );
+    glyph_caches.glyph_atlas.clear_stats();
+}
+
+/// Zero out atlas regions on the GPU after eviction.
+///
+/// Uses `queue.write_texture` to write transparent pixels to each clear rect,
+/// preventing stale data from evicted glyphs from bleeding through when the
+/// slot is reused on a subsequent frame.
+fn clear_atlas_regions(queue: &wgpu::Queue, renderer: &Renderer, rects: &[PendingClearRect]) {
```
LaurenzV: Wouldn't it be possible to clear all regions in one pass? Is this just for simplicity for now?

fontique/src/backend/coretext.rs
```diff
@@ -114,6 +114,19 @@ fn scan_coretext_available_fonts() -> Option<scan::ScannedCollection> {
         }
     }
 
+    // CoreText may not enumerate all system fonts (e.g. the system UI font SFNS.ttf).
```
LaurenzV: This strikes me as something that should probably be submitted in a separate PR.

parley_dev/assets/fonts/noto_color_emoji/README.md
LaurenzV, line 1: I think it might be good to write what command you used to subset so that it can easily be rerun in the future if needed.

parley_draw/src/atlas/key.rs
```diff
@@ -0,0 +1,209 @@
...
+/// Unique identifier for a cached glyph bitmap.
+///
+/// Two glyphs with the same key are visually identical and can share
+/// the same cached bitmap. The key includes all parameters that affect
+/// the glyph's appearance.
+///
+/// All fields including `var_coords` are included in Hash/Eq, so glyphs
+/// with different variation settings are treated as distinct cache entries.
+#[derive(Clone, Debug)]
+pub struct GlyphCacheKey {
```
LaurenzV, line 36: I haven't checked this, do you think the size of the glyph cache key will have an impact on performance/memory footprint? Is it worth trying to reduce it?

LaurenzV: For example, some fields like `hinted` and `subpixel_x` could probably be bit-packed. For context_color, we could store it as a u32 instead of AlphaColor (which internally uses 4 f32).

parley_draw/src/atlas/key.rs
```diff
@@ -0,0 +1,209 @@
...
+)]
+#[inline]
+fn quantize_subpixel(frac: f32) -> u8 {
+    let normalized = frac.fract();
+    let normalized = if normalized < 0.0 {
+        normalized + 1.0
+    } else {
+        normalized
+    };
+    ((normalized * SUBPIXEL_BUCKETS as f32).round() as u8).min(SUBPIXEL_BUCKETS - 1)
```
LaurenzV, line 144: Is this correct? Shouldn't values >= 0.875 be snapped back to 0, instead of being rounded back to 3? See also:

Image

parley_draw/src/atlas/cache.rs
```diff
@@ -0,0 +1,699 @@
...
+/// Contains the cache entries, LRU tracking, pending uploads, and statistics.
+/// Does **not** own any pixel storage — that responsibility belongs to the
+/// concrete wrapper types ([`CpuGlyphAtlas`](crate::renderers::vello_cpu::CpuGlyphAtlas), [`GpuGlyphAtlas`](crate::renderers::vello_hybrid::GpuGlyphAtlas)).
+pub struct GlyphAtlas {
+    /// Eviction configuration.
+    eviction_config: GlyphCacheConfig,
+    /// Entries for non-variable fonts.
+    static_entries: FixedHashMap<GlyphCacheKey, GlyphCacheEntry>,
+    /// Entries for variable fonts, keyed by variation coordinates.
+    variable_entries: FixedHashMap<VarKey, FixedHashMap<GlyphCacheKey, GlyphCacheEntry>>,
```
LaurenzV, line 198: Revisiting by previous comment about the memory size of glyph cache key, since variable and static are separate anyway, can we maybe create two different glyph caches so that for fonts with no variations dont need to store the empty normalized coordinates?

(not blocking, just wondering)

LaurenzV: I guess it would complicate the `get` function though, so maybe not worth it...

parley_draw/src/atlas/cache.rs
```diff
@@ -0,0 +1,699 @@
...
+            match self
+                .variable_entries
+                .raw_entry_mut()
+                .from_key(&VarLookupKey(&key.var_coords))
+            {
+                RawEntryMut::Occupied(e) => e.into_mut(),
+                RawEntryMut::Vacant(e) => {
+                    e.insert(
+                        key.var_coords.clone(),
+                        FixedHashMap::with_hasher(FixedState::with_seed(0)),
```
LaurenzV: This seed seems to be reused a couple of times, maybe worth creating a constant?

parley_draw/src/atlas/cache.rs
```diff
@@ -0,0 +1,699 @@
...
+
+    /// Take all pending bitmap uploads, leaving the internal queue empty.
+    pub fn take_pending_uploads(&mut self) -> Vec<PendingBitmapUpload> {
+        core::mem::take(&mut self.pending_uploads)
+    }
+
+    /// Take all pending clear rects, leaving the internal queue empty.
+    pub fn take_pending_clear_rects(&mut self) -> Vec<PendingClearRect> {
+        core::mem::take(&mut self.pending_clear_rects)
+    }
```
LaurenzV: It seems like all usages of this method involve just iterating once over it. maybe we just return a draining iterator instead to keep the allocation around?

LaurenzV: (Same for other instances)

parley_draw/src/atlas/cache.rs
```diff
@@ -0,0 +1,699 @@
...
+            .get_or_insert_with(|| AtlasCommandRecorder::new(page_index, atlas_width, atlas_height))
+    }
+
+    /// Advance the frame counter and potentially evict old entries.
+    pub fn maintain(&mut self, image_cache: &mut ImageCache) {
+        self.tick();
+        let frames_since_eviction = self.serial.wrapping_sub(self.last_eviction_serial);
+        if frames_since_eviction < self.eviction_config.eviction_frequency {
+            return;
+        }
```
LaurenzV: Can this overflow? Does it make sense to use weapping sub here?

LaurenzV: Ah, because the ticker itelf also wraps, right?

parley_draw/src/atlas/cache.rs
```diff
@@ -0,0 +1,699 @@
...
+                    image_cache.deallocate(entry.atlas_slot.image_id);
+                    *entry_count = entry_count.saturating_sub(1);
+                    push_clear_rect_for_slot(pending_clear_rects, &entry.atlas_slot);
+                    false
+                } else {
+                    true
+                }
+            });
+            !entries.is_empty()
+        });
```
LaurenzV: Can this be deduplicated somehow?

parley_draw/src/atlas/cache.rs
```diff
@@ -0,0 +1,699 @@
...
+        height: slot.height + 2 * GLYPH_PADDING,
+    });
+}
+
+/// Statistics about cached glyphs.
+#[cfg(debug_assertions)]
+#[derive(Debug)]
+pub struct GlyphCacheStats {
+    /// Total number of cached glyph entries.
+    pub total_glyphs: usize,
```
LaurenzV: Can this not be derived from static_glyphs + variable_glypgs?

parley_draw/src/atlas/cache.rs
```diff
@@ -0,0 +1,699 @@
...
+
+    /// Log cache hit/miss statistics at info level.
+    pub fn log_hit_miss_stats(&self) {
+        let total = self.cache_hits + self.cache_misses;
+        let hit_rate = if total > 0 {
+            (self.cache_hits as f64 / total as f64) * 100.0
+        } else {
+            0.0
+        };
+        log::info!("=== Cache Hit/Miss Statistics ===");
```
LaurenzV: Is debug maybe a more appropriate level?

parley_draw/src/atlas/cache.rs
```diff
@@ -0,0 +1,699 @@
...
+    }
+
+    /// Returns all cached glyph keys (for debugging).
+    pub fn all_keys(&self) -> Vec<&GlyphCacheKey> {
+        let mut keys: Vec<_> = self.static_entries.keys().collect();
+        for entries in self.variable_entries.values() {
+            keys.extend(entries.keys());
+        }
+        keys
+    }
```
LaurenzV: Could return an iterator here by chaining the two I think?

parley_draw/src/atlas/cache.rs
```diff
@@ -0,0 +1,699 @@
...
+            .field("static_entries", &self.static_entries.len())
+            .field("variable_fonts", &self.variable_entries.len())
+            .field("serial", &self.serial)
+            .finish_non_exhaustive()
+    }
+}
+
+/// Save a pixmap to a PNG file (diagnostic utility).
+#[cfg(all(debug_assertions, feature = "png"))]
+pub fn save_pixmap_to_png(pixmap: &Pixmap, path: &std::path::Path) -> std::io::Result<()> {
```
LaurenzV: Mark as doc hidden as well maybe? Although it's possible that `debug_assertions` also hides it from normal docs, not sure.

parley_draw/src/renderers/vello_hybrid.rs
```diff
@@ -0,0 +1,445 @@
...
+    }
+}
+
+/// Maps COLR paint operations to Vello Hybrid draw calls.
+///
+/// `fill_solid` and `fill_gradient` fill the entire surface because COLR
+/// compositing relies on clip layers to restrict the painted region.
+impl ColrRenderer for Scene {
+    fn push_clip_layer(&mut self, clip: BezPath) {
+        self.push_layer(Some(&clip), None, None, None, None);
```
taj-p, line 361: Could we add a TODO or similar to use `push_clip_path` instead of `push_clip_layer` so that we can take advantage of Vello Hybrid fast paths? I understand this is more work (to track blend layers VS clip paths) in `colr.rs`, so let's not implement that now

parley_draw/src/atlas/cache.rs
```diff
@@ -0,0 +1,699 @@
...
+    /// Take all pending atlas command recorders (one per dirty page),
+    /// leaving the internal collection empty.
+    fn take_pending_atlas_commands(&mut self) -> Vec<AtlasCommandRecorder>;
+
+    /// Take all pending clear rects, leaving the internal queue empty.
+    ///
+    /// Each rect describes an atlas region that was freed during
+    /// [`maintain`](GlyphCache::maintain) and must be zeroed to transparent.
+    /// Drain these **after** calling `maintain`.
+    fn take_pending_clear_rects(&mut self) -> Vec<PendingClearRect>;
```
taj-p: `take_pending_atlas_commands` uses `mem::take`, which transfers ownership of the `SmallVec` and every inner `Vec<AtlasCommand>` to the caller. After replay the caller drops everything, so the heap allocations are lost every frame and re-allocated from scratch next frame.

Since consumers only need to iterate and drain the commands, we could avoid this by keeping the recorders in place and letting consumers borrow them. Something like:

```rs
pub fn replay_pending_atlas_commands(
    &mut self,
    mut f: impl FnMut(&mut AtlasCommandRecorder),
) {
    for slot in &mut self.pending_atlas_commands {
        if let Some(recorder) = slot.as_mut() {
            if !recorder.commands.is_empty() {
                f(recorder);
                recorder.commands.clear();
            }
        }
    }
}
```

Consumers change from:

```rs
for mut recorder in glyph_caches.glyph_atlas.take_pending_atlas_commands() {
    glyph_renderer.reset();
    replay_atlas_commands(&mut recorder.commands, glyph_renderer);
    // ...
}
```
to
```rs
glyph_caches.glyph_atlas.replay_pending_atlas_commands(|recorder| {
    glyph_renderer.reset();
    replay_atlas_commands(&recorder.commands, glyph_renderer);
    // ...
});
```

The benefit of this approach is that the command list for each `AtlasCommandRecorder` isn't freed each frame. I think that's beneficial for operations like zoom, where we need to insert into the cache every frame (since resolved font size will presumably change each frame).

parley_draw/src/renderers/vello_hybrid.rs
```diff
@@ -0,0 +1,445 @@
...
+        context_color: AlphaColor<Srgb>,
+        recorder: &mut AtlasCommandRecorder,
+        dst_x: u16,
+        dst_y: u16,
+    ) {
+        recorder.set_transform(Affine::translate((dst_x as f64, dst_y as f64)));
+
+        let mut colr_painter = ColrPainter::new(glyph, context_color, recorder);
+        colr_painter.paint();
+    }
```
taj-p: Do these methods need to live on the trait? Since they're duplicated between Vello CPU and Vello Hybrid and are only referenced by the free floating `render_*_via_cache`, could we move these methods to `vello_renderer.rs` as free floating functions that are referenced by `render_*_via_cache` to reduce duplication?

parley_draw/src/renderers/vello_hybrid.rs
```diff
@@ -0,0 +1,445 @@
...
+
+    fn queue_bitmap_upload_to_atlas(
+        glyph: &GlyphBitmap,
+        glyph_atlas: &mut GpuGlyphAtlas,
+        atlas_slot: AtlasSlot,
+    ) {
+        // Queue for GPU upload; the application drains pending uploads and calls
+        // Renderer::write_to_atlas before the main render pass.
+        glyph_atlas.push_pending_upload(atlas_slot.image_id, Arc::clone(&glyph.pixmap), atlas_slot);
+    }
```
taj-p: Does this need to live in the trait? Can we inline it in the single call site of `render_bitmap_via_cache`?

parley_draw/src/atlas/key.rs
```diff
@@ -0,0 +1,209 @@
...
+    /// Glyph index within the font.
+    pub glyph_id: u32,
+    /// Font size as f32 bits (exact match, no quantization).
+    pub size_bits: u32,
+    /// Whether hinting was applied.
+    pub hinted: bool,
+    /// Horizontal subpixel position (0 to SUBPIXEL_BUCKETS-1).
+    pub subpixel_x: u8,
+    /// Context color for COLR glyphs (packed RGBA). 0 for non-COLR glyphs.
+    pub context_color: AlphaColor<Srgb>,
```
taj-p, line 50: HashMap lookups are now a dominant part of our profile:

Image

Of this 13.4%, 6.7% is spent in doing `PartialEq`, where 5.2% is spent in `pack_color`:

Image

Could we "pre-pack" context cooler at construction time?

i.e.:

```rs
impl GlyphCacheKey {
    pub fn new(
        ...
    ) -> Self {
        Self {
            ...
            context_color,
            context_color_packed: pack_color(context_color),
            ...
        }
    }
}

impl Hash for GlyphCacheKey {
    #[inline]
    fn hash<H: Hasher>(&self, state: &mut H) {
        self.font_id.hash(state);
        self.font_index.hash(state);
        self.glyph_id.hash(state);
        self.size_bits.hash(state);
        self.hinted.hash(state);
        self.subpixel_x.hash(state);
         // Don't include context_color
        self.context_color_packed.hash(state);
        self.var_coords.hash(state);
    }
}

impl PartialEq for GlyphCacheKey {
    #[inline]
    fn eq(&self, other: &Self) -> bool {
        self.font_id == other.font_id
            && self.font_index == other.font_index
            && self.glyph_id == other.glyph_id
            && self.size_bits == other.size_bits
            && self.hinted == other.hinted
            && self.subpixel_x == other.subpixel_x
            // Don't include context_color
            && self.context_color_packed == other.context_color_packed
            && self.var_coords == other.var_coords
    }
}

```

taj-p: The above proposal ~halves lookup time:

Image

parley_draw/src/atlas/key.rs
```diff
@@ -0,0 +1,209 @@
...
+    #[inline]
+    fn eq(&self, other: &Self) -> bool {
+        self.font_id == other.font_id
+            && self.font_index == other.font_index
+            && self.glyph_id == other.glyph_id
+            && self.size_bits == other.size_bits
+            && self.hinted == other.hinted
+            && self.subpixel_x == other.subpixel_x
+            && pack_color(self.context_color) == pack_color(other.context_color)
+            && self.var_coords == other.var_coords
```
taj-p:
```suggestion
    }
}

impl PartialEq for GlyphCacheKey {
    #[inline]
    fn eq(&self, other: &Self) -> bool {
        self.font_id == other.font_id
            && self.font_index == other.font_index
            && self.glyph_id == other.glyph_id
            && self.size_bits == other.size_bits
            && self.hinted == other.hinted
            && self.subpixel_x == other.subpixel_x
            && pack_color(self.context_color) == pack_color(other.context_color)
```

I don't think we need to include `var_coords` in our cache key because we use a two level map structure in `GlyphAtlas` that already guarantees that all keys within a given inner map share the same var_coords. Perhaps just add documentation telling consumers that `var_coords` isn't part of `PartialEq` and `Hash` and that they're expected to use two level maps.

The reason why we want this: after https://github.com/linebender/parley/pull/548/changes#r2861350324, most time is spent evaluating `var_coords` in our atlas lookup:

Image

parley_draw/src/atlas/key.rs
```diff
@@ -0,0 +1,209 @@
...
+        self.var_coords.hash(state);
+    }
+}
+
+impl PartialEq for GlyphCacheKey {
+    #[inline]
+    fn eq(&self, other: &Self) -> bool {
+        self.font_id == other.font_id
+            && self.font_index == other.font_index
+            && self.glyph_id == other.glyph_id
```
taj-p: Font ID and font index are usually the same within a run. I recommend discriminating by `glyph_id` first (which is probably the most discriminating value) to short circuit this compound boolean:

i.e.:

```rs
self.glyph_id == other.glyph_id
&& self.font_id == other.font_id
&& ...
```

It might be worth reordering some of the other members, but I think `glyph_id` is likely the most obvious win.

taj-p (commented):
Looking good! Just flushing my comments before I go into meetings

parley_draw/src/glyph.rs
```diff
@@ -241,52 +449,122 @@ impl<'a, T: GlyphRenderer + 'a> GlyphRunBuilder<'a, T> {
...
+                    image_cache,
+                );
+                continue;
+            }
+
+            // ── Outline Glyphs ──────────────────────────────────────────
+            let Some(outline) = outlines.get(glyph_id) else {
+                // Glyph has no outline, bitmap, or COLR representation — skip.
+                continue;
             };
```
taj-p, line 572: Getting the outline from the font file is actually very expensive. It's showing up as a dominant part of my profile (rendering the same paragraph 1000 times):

Image

Can we restructure this to only perform the `outlines.get` on a cache miss? (The same comment applies to the colr and bitmap branches)

i.e.

```rs
            let transform = calculate_outline_transform(
                glyph,
                initial_transform,
                self.run.transform,
                hinting_instance,
            );

            // Sub-pixel x offset is quantized into the cache key so that glyphs
            // at different fractional positions get distinct cached bitmaps.
            let cache_key = self.atlas_cache_enabled.then(|| {
                let fractional_x = transform.translation().x.fract() as f32;
                GlyphCacheKey::new(
                    font_id,
                    font_index,
                    glyph.id,
                    ppem,
                    hinted,
                    fractional_x,
                    BLACK,
                    normalized_coords,
                )
            });

            if let Some(ref key) = cache_key {
                if let Some(cached_slot) = glyph_atlas.get(key) {
                    self.renderer.render_cached_glyph(
                        cached_slot,
                        transform,
                        CachedGlyphType::Outline,
                    );
                    continue;
                }
            }

            // Cache miss — fetch the outline from skrifa and build the path.
            let Some(outline) = outlines.get(glyph_id) else {
                continue;
            };
```

After this change, that huge block in our profile disappears! 🎉 

See the before (left) and after (right) profiles below

Image

parley_draw/src/glyph.rs
```diff
@@ -89,7 +116,7 @@ pub struct BitmapGlyph {
 /// Clients are supposed to first draw the glyph into an intermediate image texture/pixmap
 /// and then render that into the actual scene, in a similar fashion to
 /// bitmap glyphs.
-pub struct ColorGlyph<'a> {
+pub struct ColrGlyph<'a> {
```
taj-p: Since we renamed the other similar structs to use the a "Glyph" prefix, should we do the same here?

parley_draw/src/glyph.rs
```diff
@@ -107,33 +134,92 @@ pub struct ColorGlyph<'a> {
...
+    /// The `image_cache` parameter is the allocator used by `glyph_atlas` for atlas
+    /// region allocation — see [`fill_glyph`](Self::fill_glyph) for details.
+    ///
+    /// Outline and COLR glyph draw commands are recorded into the glyph atlas's
+    /// pending atlas commands queue for later replay into a glyph renderer.
+    fn stroke_glyph(
+        &mut self,
+        glyph: PreparedGlyph<'_>,
+        glyph_atlas: &mut C,
+        image_cache: &mut ImageCache,
```
taj-p, line 192: Is the plan to (sometime in the future) move these `vello_common` dependencies (and the shelf allocator) into a separate crate so that Glifo doesn't depend on `vello_common`?

parley_draw/src/atlas/key.rs
taj-p: I think we can optimise this further. Either:

- Each glyph in a run uses the same color, so we don't need to recalculate `pack_color` for each glyph. Just do it once per run. (Outline glyphs cache `pack_color(BLACK)` -> we don't even need to compute this twice for outline glyphs)
- Store this as an `Option` because outline glyphs always pass `BLACK` anyway.

The reason I think this is important: In my benchmark, the time spent in Parley Draw is almost 20% spent in `pack_color`

parley_draw/src/atlas/mod.rs
```diff
@@ -0,0 +1,34 @@
+// Copyright 2025 the Parley Authors
```
LaurenzV: Not 2026?

parley_draw/src/atlas/cache.rs
LaurenzV: While it's unlikely, does it still make sense to use a u64 here maybe?

parley_draw/src/renderers/vello_renderer.rs
```diff
@@ -0,0 +1,578 @@
...
+                    renderer,
+                    &glyph.path,
+                    transform,
+                    key,
+                    glyph_atlas,
+                    image_cache,
+                    tint_color,
+                ) {
+                    return;
+                }
```
LaurenzV: Isn't this the same code as for fill_glyphs? Where is the stroking happening here? 🤔

parley_draw/src/glyph.rs
```diff
@@ -241,52 +449,122 @@ impl<'a, T: GlyphRenderer + 'a> GlyphRunBuilder<'a, T> {
...
+                glyph,
+                initial_transform,
+                self.run.transform,
+                hinting_instance,
+            );
+
+            // Sub-pixel x offset is quantized into the cache key so that glyphs
+            // at different fractional positions get distinct cached bitmaps.
+            let cache_key = self.atlas_cache_enabled.then(|| {
+                let fractional_x = transform.translation().x.fract() as f32;
```
LaurenzV: Now that we encode the fractional translation in the subpixel offset, do we not need to update the transform to discard the fraction from there? Also, why are y fractions not handled here?

parley_draw/src/renderers/vello_renderer.rs
```diff
@@ -0,0 +1,578 @@
...
+        }
+    }
+}
+
+/// Try to render an outline glyph via the atlas cache.
+///
+/// On cache miss, allocates atlas space (the insert returns the per-page
+/// command recorder) and delegates rasterisation to the backend via
+/// [`GlyphAtlasBackend::render_outline_to_atlas`].
+fn render_outline_via_cache<B: GlyphAtlasBackend>(
```
LaurenzV: To me, this makes it sound like in case the glyph already is in the cache, weitwill simply draw the bitmap. But If I'm understanding correctly now, this method is only called if there was a cache miss in the first place? Maybe it should be renamed into something like `render_outline_and_insert_into_cache`? Also, the doc comment probably needs updating (as it suggests that there isn't always a cache miss), unless I'm understanding it wrong. Same for the other two functions below.

parley_draw/src/atlas/cache.rs
```diff
@@ -0,0 +1,699 @@
...
+///
+/// The hybrid (GPU) renderer samples atlas sub-images via `Extend::Pad`, which
+/// clamps out-of-bounds coordinates to the edge texel. Without at least 1px of
+/// transparent padding, strip-rasteriser overshoot at glyph boundaries would
+/// either duplicate the edge row/column or bleed in content from a neighbouring
+/// glyph allocation. 1px is sufficient: the overshoot is sub-pixel, and the
+/// transparent padding absorbs it.
+///
+/// TODO(atlas): Remove this padding once we have a way to handle overshoot.
+pub const GLYPH_PADDING: u16 = 1;
```
taj-p, line 49: This seems like good practice so that we could consider using a native bilinear sapper in hybrid. Is that right?

parley_draw/src/renderers/vello_cpu.rs
```diff
@@ -1,189 +1,632 @@
...
+            let mut path = std::path::PathBuf::from(env!("CARGO_MANIFEST_DIR"));
+            path.pop(); // up from parley_draw to workspace root
+            path.push("examples");
+            path.push("_output");
+            let _ = std::fs::create_dir_all(&path);
+            path.push(format!("vello_cpu_atlas_page_{i}.png"));
+            let _ = save_pixmap_to_png(pixmap, &path);
+        }
+    }
+}
```
LaurenzV: Can't say I'm a fan of having this here, even if it's only conditionally compiled in. 😄 But if others don't mind I suppose it's fine.

parley_draw/src/renderers/vello_cpu.rs
```diff
@@ -1,189 +1,632 @@
...
+
+        let state = renderer.take_current_state();
+
+        renderer.set_tint(tint);
+        renderer.set_transform(rect_transform);
+        renderer.set_paint(image);
+        renderer.set_paint_transform(paint_transform);
+        renderer.fill_rect(&area);
+
+        renderer.reset_tint();
```
LaurenzV, line 375: Shouldn't `restore_state` take care of this?

parley_draw/src/renderers/vello_cpu.rs
```diff
@@ -1,189 +1,632 @@
...
+        renderer.restore_state(state);
+    }
+
+    fn paint_transform(atlas_slot: &AtlasSlot) -> Affine {
+        // The CPU backend uses full-page pixmaps, so the paint origin must be
+        // shifted to the glyph's slot within the page. Contrast with the hybrid
+        // backend, which gets per-allocation origins from the image cache.
```
LaurenzV, line 382: Not sure I understand this, I thought vello_cpu and vello_hybrid work the same, where we have one (or multiple) huge atlas pages and glyphs are drawn into that at specific offsets. Why is there different behavior here?

parley_draw/src/atlas/cache.rs
```diff
@@ -0,0 +1,699 @@
...
+
+/// Configuration for glyph cache eviction behavior.
+#[derive(Clone, Debug)]
+pub struct GlyphCacheConfig {
```
taj-p, line 53: Is it possible to configure minimum font size (after which we use renderer sampling to interpolate smaller sizes) and a maximum size (after which we always directly render)?

Cargo.toml
```diff
@@ -106,3 +110,12 @@ clippy.negative_feature_names = "warn"
...
 clippy.wildcard_dependencies = "warn"
 # END LINEBENDER LINT SET
+
+[patch.crates-io]
+vello_common = { git = "https://github.com/linebender/vello.git", branch = "gemberg/glyph-cache-rect-allocator" }
+vello_cpu = { git = "https://github.com/linebender/vello.git", branch = "gemberg/glyph-cache-rect-allocator" }
+vello_hybrid = { git = "https://github.com/linebender/vello.git", branch = "gemberg/glyph-cache-rect-allocator" }
+
+[profile.profiling]
+inherits = "release"
```
LaurenzV, line 120: strip = "none" might also be worth adding!

parley_draw/src/atlas/cache.rs
```diff
@@ -0,0 +1,699 @@
...
+            max_entry_age: 64,
+            eviction_frequency: 64,
+        }
+    }
+}
+
+/// Common interface for glyph atlas caches.
+///
+/// Both the CPU and hybrid (GPU) renderers use this trait to interact with the
+/// glyph cache. Shared rendering code in `vello_renderer` is generic over this
```
taj-p: Since `vello_renderer` is an optional feature (and the vision it to support arbitrary renderers) with glifo, I recommend keeping those details out of the common interfaces

parley_draw/src/renderers/vello_hybrid.rs
```diff
@@ -0,0 +1,445 @@
...
+
+        let state = renderer.take_current_state();
+
+        renderer.set_tint(tint);
+        renderer.set_transform(rect_transform);
+        renderer.set_paint(image);
+        renderer.set_paint_transform(paint_transform);
+        renderer.fill_rect(&area);
+
+        renderer.reset_tint();
```
LaurenzV, line 292: Same as in vello_cpu

parley_draw/src/renderers/vello_hybrid.rs
```diff
@@ -0,0 +1,445 @@
...
+        renderer.restore_state(state);
+    }
+
+    fn paint_transform(_atlas_slot: &AtlasSlot) -> Affine {
+        // The image cache resolves the slot to the allocation origin, but the
+        // actual glyph bitmap is inset by GLYPH_PADDING on each side (a guard
+        // band required by Extend::Pad sampling). Shift paint to compensate.
+        let padding = GLYPH_PADDING as f64;
+        Affine::translate((-padding, -padding))
```
LaurenzV, line 301: Do we not need glyph padding for vello_cpu?

parley_draw/src/renderers/vello_renderer.rs
taj-p: This function name and documentation is a little bit misleading. It looks like we always allocate the glyph into the atlas and render the glyph from the atlas to screen unconditionally - the check for whether the glyph is in the cache happens upstream.

parley_draw/src/glyph.rs
```diff
@@ -52,11 +58,26 @@ pub struct Glyph {
...
+#[derive(Debug, Clone, Copy)]
+pub enum CachedGlyphType {
+    /// An outline glyph cached in the atlas.
+    Outline,
+    /// A bitmap glyph cached in the atlas.
+    Bitmap,
+    /// A COLR glyph cached in the atlas.
+    /// The `Rect` parameter contains the fractional area dimensions
+    /// to preserve sub-pixel accuracy during rendering.
+    Colr(Rect),
```
LaurenzV, line 91: Why do Colr and Outline have different behavior here? Because we don't use subpixel offsets here, since they are not as critical for emojis?

parley_draw/src/renderers/vello_renderer.rs
```diff
@@ -0,0 +1,578 @@
...
+
+    let area = glyph.area;
+
+    let Some((dst_x, dst_y, atlas_slot, recorder)) =
+        glyph_atlas.insert(image_cache, cache_key.clone(), raster_metrics)
+    else {
+        return CacheResult::AtlasFull;
+    };
+
+    B::render_colr_to_atlas(glyph, cache_key.context_color, recorder, dst_x, dst_y);
```
taj-p:
```suggestion
    let context_color = cache_key.context_color;
    let Some((dst_x, dst_y, atlas_slot, recorder)) =
        glyph_atlas.insert(image_cache, cache_key, raster_metrics)
    else {
        return CacheResult::AtlasFull;
    };

    B::render_colr_to_atlas(glyph, context_color, recorder, dst_x, dst_y);
```

We can avoid the clone like this

parley_draw/src/renderers/vello_renderer.rs
```diff
@@ -0,0 +1,578 @@
...
+
+/// Try to render a bitmap glyph via the atlas cache.
+///
+/// On cache miss, delegates atlas population to the backend via
+/// [`GlyphAtlasBackend::queue_bitmap_upload_to_atlas`].
+fn render_bitmap_via_cache<B: GlyphAtlasBackend>(
+    renderer: &mut B::Renderer,
+    glyph: &GlyphBitmap,
+    transform: Affine,
+    cache_key: &GlyphCacheKey,
```
taj-p:
```suggestion
    cache_key: GlyphCacheKey,
```

We can pass an owned value here (to be honest with callers that we need an owned value, rather than hiding the unconditional clone) and then take the `Option<GlyphCacheKey>` in the caller to avoid cloning the value altogether:

```rs
            if let Some(key) = cache_key.take() {
                if let CacheResult::CachedAndRendered = render_bitmap_via_cache::<B>(
                    renderer,
                    &glyph,
                    transform,
                    key,
                    glyph_atlas,
                    image_cache,
                ) {
                    return;
                }
            }
```

We should we able to do this for all the glyph types

parley_draw/src/renderers/vello_renderer.rs
```diff
@@ -0,0 +1,578 @@
...
+        return CacheResult::NotAxisAligned;
+    }
+
+    let bounds = path.bounding_box();
+    let raster_metrics = calculate_raster_metrics(&bounds);
+
+    let subpixel_offset = subpixel_offset(cache_key.subpixel_x);
+
+    let Some((dst_x, dst_y, atlas_slot, recorder)) =
+        glyph_atlas.insert(image_cache, cache_key.clone(), raster_metrics)
```
taj-p: See below comment on how to avoid this clone

parley_draw/src/atlas/cache.rs
```diff
@@ -0,0 +1,699 @@
...
+    ///
+    /// Returns `(dst_x, dst_y, atlas_slot, recorder)` if successful, `None`
+    /// if allocation failed (e.g., atlas is full).  The returned recorder
+    /// accumulates draw commands for the atlas page the glyph was placed on.
+    fn insert(
+        &mut self,
+        image_cache: &mut ImageCache,
+        key: GlyphCacheKey,
+        raster_metrics: RasterMetrics,
+    ) -> Option<(u16, u16, AtlasSlot, &mut AtlasCommandRecorder)>;
```
taj-p, line 95: It's unfortunate that this form of API necessitates a redundant hash map lookup on cache miss. I.e., on cache miss, we perform another hash map operation to insert (being unable to re-use the `RawEntry` or similar). 

We could precompute the hash once and use something like `HashMap::raw_entry_mut().from_key_hashed_nocheck(...)` but let's maybe leave that as a follow up

parley_draw/src/atlas/cache.rs
taj-p: See my other comment. It actually shows up in profiles so it's worth addressing. I suggest simply ignoring it in the hashing implementation as described but document the contract that we expect a 2 level cache.

parley_draw/src/atlas/cache.rs
```diff
@@ -0,0 +1,699 @@
...
+    /// Total cached glyph count (across all maps).
+    entry_count: usize,
+    /// Bitmap glyphs awaiting GPU upload.
+    pending_uploads: Vec<PendingBitmapUpload>,
+    /// Atlas regions that must be cleared to transparent before compositing.
+    pending_clear_rects: Vec<PendingClearRect>,
+    /// Outline and COLR glyph commands awaiting replay, indexed by atlas page.
+    /// Uses `SmallVec` with inline capacity of 1 because most applications use
+    /// a single atlas page; the common case avoids heap allocation entirely.
+    pending_atlas_commands: SmallVec<[Option<AtlasCommandRecorder>; 1]>,
```
taj-p, line 212: Is there a way to configure a maximum value for this?

parley_draw/src/atlas/commands.rs
```diff
@@ -0,0 +1,173 @@
+// Copyright 2025 the Parley Authors
```
taj-p:
```suggestion
// Copyright 2026 the Parley Authors
```

 Can you update the files to use 2026?

parley_draw/src/atlas/commands.rs
```diff
@@ -0,0 +1,173 @@
...
+use alloc::vec::Vec;
+
+use crate::color::{AlphaColor, Srgb};
+use crate::kurbo::{Affine, BezPath, Rect};
+use crate::peniko::{BlendMode, Gradient};
+
+/// Paint type for atlas commands.
+///
+/// Covers the subset of paint types used when rendering glyphs to the atlas:
+/// solid colours (outlines, COLR solid fills) and gradients (COLR gradient fills).
```
taj-p: Can you please move this documentation onto the enum members instead at the top level

parley_draw/src/atlas/commands.rs
```diff
@@ -0,0 +1,173 @@
...
+        self.commands.push(AtlasCommand::SetTransform(t));
+    }
+
+    /// Set the current paint (accepts `AlphaColor<Srgb>` or `Gradient`).
+    pub fn set_paint(&mut self, paint: impl Into<AtlasPaint>) {
+        self.commands.push(AtlasCommand::SetPaint(paint.into()));
+    }
+
+    /// Set the paint transform.
+    pub fn set_paint_transform(&mut self, t: Affine) {
```
taj-p, line 130: Can we inline these (and the other method of this struct)?

parley_draw/src/colr.rs
```diff
@@ -401,3 +398,41 @@ pub(crate) fn convert_bounding_box(rect: BoundingBox<f32>) -> Rect {
         f64::from(rect.y_max),
     )
 }
+
+impl ColrRenderer for AtlasCommandRecorder {
```
taj-p, line 402: Should we inline these?

parley_draw/src/glyph.rs
```diff
@@ -241,52 +449,122 @@ impl<'a, T: GlyphRenderer + 'a> GlyphRunBuilder<'a, T> {
...
+                // context color; variation coords are irrelevant for fixed strikes.
+                let cache_key = self.atlas_cache_enabled.then(|| GlyphCacheKey {
+                    font_id,
+                    font_index,
+                    glyph_id: glyph.id,
+                    size_bits: bitmap_ppem.to_bits(),
+                    hinted: false,
+                    subpixel_x: 0,
+                    context_color: BLACK,
+                    var_coords: SmallVec::new(),
```
taj-p, line 534: Depending on `var_coords` for the run, this could allocate every time. I think we either should consider:
1. removing `var_coords` from the key and rely on the 2 level map structure (which encodes the `var_coords` anyway); or
2. we should use the `Equivalent` trait to create borrowed keys (shown below) and convert it to an owned key on cache misses only - see `parley/src/lru_cache.rs` for an example implementation


```rs
#[derive(Clone, Copy)]
pub struct GlyphCacheKeyRef<'a> {
    pub font_id: u64,
    pub font_index: u32,
    pub glyph_id: u32,
    pub size_bits: u32,
    pub hinted: bool,
    pub subpixel_x: u8,
    pub context_color_packed: u32,
    pub var_coords: &'a [NormalizedCoord],
}
```

parley_draw/src/glyph.rs
```diff
@@ -222,14 +359,85 @@ impl<'a, T: GlyphRenderer + 'a> GlyphRunBuilder<'a, T> {
...
+            Style::Stroke => GlyphRenderer::<C>::stroke_glyph,
         };
 
         for glyph in glyphs {
-            let bitmap_data = bitmaps
-                .glyph_for_size(Size::new(self.run.font_size), GlyphId::new(glyph.id))
+            let glyph_id = GlyphId::new(glyph.id);
+
+            // ── COLR Glyphs ───────────────────────────────────────────
+            if let Some(color_glyph) = color_glyphs.get(glyph_id) {
```
taj-p, line 435: It's expensive to do the `color_glyphs` and `bitmaps.glyph_for_size` lookups. We can't ignore the priority of checking colr, then bitmap, then outline, but we could speculatively check for an outline glyph in the cache before performing the colr / bitmap lookups. 

`calculate_outline_transform` is just arithmetic on affine coefficients. It has no font table parsing so computing it early is essentially free.

If it's not in the cache, we drop through to checking for colr and bitmap (and rendering the outline). But, I think it's worth greedily checking for outline glyphs in the cache to avoid this work (since 99% of the time we will be rendering outline glyphs).

parley_draw/src/glyph.rs
taj-p: I think the fractional removal is handled in `render_outline_glyph_from_atlas`

parley_tests/tests/util/renderer.rs
taj-p, line 1: Could we please add some tests for bitmap and colr emojis since we've added in 2 emoji fonts?

parley_draw/src/renderers/vello_cpu.rs
```diff
@@ -1,189 +1,632 @@
...
+    /// One `Pixmap` per atlas page, grown on demand. Wrapped in `Arc` so
+    /// callers can cheaply share a page with a render context.
+    pixmaps: Vec<Arc<Pixmap>>,
+    /// Width of each atlas page in pixels.
+    page_width: u16,
+    /// Height of each atlas page in pixels.
+    page_height: u16,
+}
+
+impl CpuGlyphAtlas {
```
taj-p, line 54: Can we please make these methods inline?

parley_draw/src/renderers/vello_cpu.rs
```diff
@@ -1,189 +1,632 @@
...
+            .field("page_width", &self.page_width)
+            .field("page_height", &self.page_height)
+            .finish()
+    }
+}
+
+/// Delegates to the inner [`GlyphAtlas`], additionally growing the `pixmaps`
+/// vector when `insert` opens a new atlas page.
+impl GlyphCache for CpuGlyphAtlas {
+    fn get(&mut self, key: &GlyphCacheKey) -> Option<AtlasSlot> {
```
taj-p, line 130:
```suggestion
    #[inline(always)]
    fn get(&mut self, key: &GlyphCacheKey) -> Option<AtlasSlot> {
```

parley_draw/src/renderers/vello_cpu.rs
```diff
@@ -1,189 +1,632 @@
...
+        self.inner.cache_hits()
+    }
+
+    fn cache_misses(&self) -> u64 {
+        self.inner.cache_misses()
+    }
+
+    fn clear_stats(&mut self) {
+        self.inner.clear_stats();
+    }
```
taj-p, line 228: Could we please make these methods inline?

parley_draw/src/renderers/vello_cpu.rs
taj-p: We could create a `renderers/debug` folder and put these implementations there?

parley_draw/src/renderers/vello_cpu.rs
```diff
@@ -1,189 +1,632 @@
...
     }
 
     fn pop_layer(&mut self) {
         Self::pop_layer(self);
     }
 }
+
+/// Allows recorded [`AtlasCommand`](crate::atlas::commands::AtlasCommand)s
+/// to be replayed into a CPU [`RenderContext`].
+impl AtlasReplayTarget for RenderContext {
```
taj-p, line 475: Should we make these inline?

parley_draw/src/renderers/vello_hybrid.rs
```diff
@@ -0,0 +1,445 @@
...
+    pub fn with_config(eviction_config: GlyphCacheConfig) -> Self {
+        Self {
+            inner: GlyphAtlas::with_config(eviction_config),
+        }
+    }
+}
+
+/// Thin delegation to the inner [`GlyphAtlas`]. No page-level pixel storage
+/// to manage here — the GPU owns atlas textures.
+impl GlyphCache for GpuGlyphAtlas {
```
taj-p, line 65: Should we make some of these methods inline?

parley_draw/src/renderers/vello_hybrid.rs
```diff
@@ -0,0 +1,445 @@
...
+            hinting_cache: HintCache::default(),
+            underline_exclusions: Vec::new(),
+            glyph_atlas: GpuGlyphAtlas::with_config(eviction_config),
+        }
+    }
+}
+
+/// Bridges Parley's [`GlyphRenderer`] trait to the shared
+/// [`vello_renderer`] cache orchestration for the hybrid backend.
+impl GlyphRenderer<GpuGlyphAtlas> for Scene {
```
taj-p, line 177: Could we please make some of these methods inline?

taj-p (commented):
Nice! This looks _really_ good! Some comments

.github/workflows/ci.yml
grebmeg: Originally it was because of `guillotiere`, but "now" that guillotiere is `no_std`-compatible that should not be required.

examples/common/src/lib.rs
grebmeg: Yes, this is expected for now — `vello_hybrid` doesn't currently support uploading bitmap pixmaps directly, so without the atlas cache the bitmap emoji path hits an unimplemented code path. This is being addressed in linebender/vello#1459. With the atlas cache enabled, bitmaps are rasterized into the atlas first, avoiding this.

examples/common/src/lib.rs
grebmeg: Unfortunately `CARGO_MANIFEST_DIR` gives the crate's manifest directory (e.g. `examples/vello_cpu_render/`), not the workspace root, and there's no stable `CARGO_WORKSPACE_DIR` equivalent. So we can't just append a path from the root. Instead I switched to `env!("CARGO_MANIFEST_DIR")` and pop one level to reach `examples/`, which is simpler and more robust than the old `file!()` + canonicalize + pop-3 approach.

examples/common/src/lib.rs
grebmeg: The aggregated stats are intentional, they give an amortized view across the full cache lifecycle (cold, warm, partial overlap, etc.), which is useful for comparing overall cached vs uncached performance. Each phase also prints its own per-phase stats. This is a development/benchmarking aid and the mixed summary is helpful for getting a quick read on the overall picture. Happy to revisit the output format later if needed.

examples/vello_hybrid_render/src/main.rs
grebmeg: Yes, this is just for simplicity for now. The proper approach would be to use `vello_hybrid`'s GPU clear shader to clear all rects in a single render pass, but that requires exposing a batched clear API on `vello_hybrid`'s Renderer. I'll do that as a follow-up in Vello. For now I've at least reused a single zeroed buffer across rects to avoid the per-rect allocation.

fontique/src/backend/coretext.rs
grebmeg: Good point, split this out into #568.

parley_dev/assets/fonts/noto_color_emoji/README.md
grebmeg: This is a 1:1 copy from the Vello repo — I didn't do the subsetting myself. Do you have the command handy from when you created it? Happy to add it here (and maybe in Vello too).

parley_draw/src/atlas/key.rs
grebmeg: Good call! The main bottleneck turned out to be `pack_color` being called on every hash/eq check (as @taj-p identified in profiling). I'll pre-pack the color to `u32` at construction time. The `hinted`/`subpixel_x` bit-packing would save a few bytes but probably isn't worth the readability cost — not sure yet. The `var_coords` removal from `Hash`/`Eq` (per @taj-p's suggestion) is the other big win, since the two-level map already guarantees that.

parley_draw/src/atlas/key.rs
grebmeg: Good catch! Values in `[0.875, 1.0)` are indeed geometrically closer to `0.0` (next pixel) than to `0.75`. However, snapping to bucket `0` here without also bumping the integer pixel coordinate by `1` would actually make things worse.

For example, a glyph at `x = 5.92` splits into integer `5`, fractional `0.92`. If we snap the fractional part to bucket `0` (offset `0.0`) but keep integer position `5`, the glyph renders at `5.0` instead of `~5.92` — that's `~0.92px` off. To get the correct result (`6.0`), we'd need to also increment the integer coordinate by `1`, which means `quantize_subpixel` would need to return both a bucket and a carry bit, and the caller would need to apply it.

With the current clamp to bucket `3` (offset `0.75`), the glyph renders at `5.75` — only `0.17px` off. The worst case error is `0.125px`, which is visually negligible. I'll add a comment explaining this trade-off.

parley_draw/src/atlas/cache.rs
grebmeg: Good point! The two-level map structure already separates static and variable fonts. As @taj-p suggested, I'll exclude `var_coords` from the `Hash impl` and document the contract that the outer map provides the variation-coordinate discrimination.

parley_draw/src/atlas/cache.rs
grebmeg: Done, extracted a `HASH_SEED` constant and two compile-time const maps (`EMPTY_GLYPH_MAP`, `EMPTY_VAR_MAP`) to replace the repeated `FixedHashMap::with_hasher(FixedState::with_seed(0))` calls.

parley_draw/src/atlas/cache.rs
grebmeg: Switched `take_pending_uploads`, `take_pending_clear_rects`, and `take_pending_atlas_commands` to `drain_*` variants that preserve the internal allocation across frames. The hybrid example uses the drain iterators directly; the CPU example still needs to collect into a temporary `Vec` because the loop bodies re-borrow the atlas via `page_pixmap_mut()` — left a TODO about considering another approach there.

parley_draw/src/atlas/cache.rs
grebmeg: Good call! Updated all serial/frame counters (`serial`, `last_eviction_serial`, `max_entry_age`, `eviction_frequency`) to `u64`. With `u64` wrapping is no longer a practical concern (~2.4 billion years at 240 FPS), so I also simplified the arithmetic to plain subtraction/addition.

parley_draw/src/atlas/cache.rs
grebmeg: Good call! Extracted the eviction check into a shared `should_retain` closure used by both `static_entries` and `variable_entries` retain calls.

parley_draw/src/atlas/cache.rs
grebmeg: Replaced the redundant `total_glyphs` field with a `total_glyphs()` method that derives it from `static_glyphs` + `variable_glyphs`.

parley_draw/src/atlas/cache.rs
grebmeg: Maybe yeah, I don't mind either way. Changed all logging in `log_hit_miss_stats`, `log_atlas_stats`, and `log_keys_grouped` to debug level.

parley_draw/src/atlas/cache.rs
grebmeg: Thanks! Changed to return `impl Iterator` by chaining the two key iterators, avoiding the allocation.

parley_draw/src/atlas/cache.rs
grebmeg: Good idea, added `#[doc(hidden)]`.

parley_draw/src/renderers/vello_hybrid.rs
grebmeg: Added a TODO comment.

parley_draw/src/atlas/cache.rs
grebmeg: Good idea! Switched to a `replay_pending_atlas_commands` callback API that keeps recorders in place and only clears their command `Vec`s after replay, preserving capacity across frames. For the CPU backend, added a `replay_pending_atlas_commands_with_pixmaps` helper that splits the borrow to give the closure access to both the recorder and the page pixmaps.

parley_draw/src/renderers/vello_hybrid.rs
grebmeg: Done! These were originally different per-backend (before the `AtlasCommandRecorder` abstraction), but became identical once both backends started recording through the same recorder. Moved all three (`render_outline_to_atlas`, `render_colr_to_atlas`, `queue_bitmap_upload_to_atlas`) out of the trait into free functions in `vello_renderer.rs`

parley_draw/src/renderers/vello_hybrid.rs
grebmeg: Inlined `queue_bitmap_upload_to_atlas` into its single call site in `render_bitmap_via_cache` — it was a one-liner so the extra indirection wasn't worth it.

parley_draw/src/atlas/key.rs
grebmeg: Applied: added pre-packed `context_color_packed: u32` field (computed once at construction), use it in `Hash`/`PartialEq`, and hoisted `pack_color` out of the per-glyph loop (once per run). Skipped for now the `Option` approach, with a pre-packed `u32` the hot-path cost is just an integer comparison regardless of glyph type.

parley_draw/src/atlas/key.rs
grebmeg: Done. Removed `var_coords` from `Hash`/`PartialEq` and documented the two-level map contract.

parley_draw/src/atlas/key.rs
grebmeg: Reordered. `glyph_id` and `subpixel_x` are now checked first in `PartialEq` for early short-circuit.

parley_draw/src/glyph.rs
grebmeg: Nice! Thanks! It's my bad, I missted that somehow. Done. Moved `outlines.get()` below the cache hit check so it's only called on cache misses.

parley_draw/src/glyph.rs
grebmeg: Oh yeah, of course! Renamed `ColrGlyph` → `GlyphColr` for consistency with the other structs.

parley_draw/src/glyph.rs
grebmeg: Yes, the long-term plan is to extract shared types (color, basic geometry) and the allocator into their own crate(s) so glifo doesn't depend on `vello_common` directly. For now the coupling is pragmatic: those types are stable and we'd just be re-exporting the same things. Happy to track this as a follow-up issue if that would be helpful?

parley_draw/src/renderers/vello_renderer.rs
grebmeg: Good catch, thanks! This is actually already like [that on main](https://github.com/linebender/parley/blob/bbc4e9979b3f9ba0e14c03c5bd5b24783495763b/parley_draw/src/renderers/vello_cpu.rs#L137) — I just didn't notice. Split `render_outline_directly` into `fill_outline_directly` and `stroke_outline_directly`, with the stroke variant now correctly calling stroke_path.

parley_draw/src/glyph.rs
grebmeg: Good question! The fractional removal at render time is handled by `render_outline_glyph_from_atlas` which uses `tx.floor()`/`ty.floor()` — so the full transform is passed through but only the integer part is used for placement.

For `y` fractions: when hinting is enabled, the `y` translation is already rounded in `calculate_outline_transform`. When hinting is off, if I'm not missing something, vertical sub-pixel shifts are less perceptually impactful for horizontal text, and adding a `y` dimension would multiply cache entries. Added a comment to clarify this.

parley_draw/src/renderers/vello_renderer.rs
grebmeg: Addressed in the reply to Laurenz's comment above — renamed all three functions and updated the docs.

grebmeg: Agreed, I went through multiple iterations on these functions and the current naming was a leftover from an earlier one. Renamed to `insert_and_render_outline` / `insert_and_render_bitmap` / `insert_and_render_colr` and updated the doc comments to clarify they always insert + render, with the cache-hit check happening upstream.

parley_draw/src/atlas/cache.rs
grebmeg: Yes, that's right — the `1px` padding absorbs strip-rasteriser overshoot today, and would also give us room to switch to native bilinear sampling in hybrid down the road. Updated the doc comment to mention this.

parley_draw/src/renderers/vello_cpu.rs
grebmeg: Good call! Moved all the debug/diagnostic impl blocks (`save_atlas_pages`, stats, key inspection) along with `save_pixmap_to_png` into a new `renderers/debug/` folder as suggested.

parley_draw/src/renderers/vello_cpu.rs
grebmeg: Good question! Unfortunately `tint` is currently not part of `vello_cpu`'s `RenderState`: `take_current_state`/`restore_state` only covers `transform`, `fill_rule`, `stroke`, `paint`, and `paint_transform`. So the explicit `reset_tint()` is needed to avoid leaking the tint into subsequent draw calls. But we'll upstream a fix to include `tint` in `RenderState`, but that'll be a separate change in vello.

parley_draw/src/renderers/vello_cpu.rs
grebmeg: Both backends do use atlas pages with glyphs at specific offsets. The difference is in how the paint transform is set up when reading back from the atlas:

- vello_cpu: The `ImageSource` references the **entire** page pixmap, so we need to shift the paint origin to the glyph's `(x, y)` position within the page.

- vello_hybrid: Vello's image cache resolves each allocation to its own origin, so we only need to compensate for the `GLYPH_PADDING` inset.

The underlying atlas structure is the same — it's just the coordinate system for the paint source that differs between the two rendering paths.

parley_draw/src/atlas/cache.rs
grebmeg: Added `max_cached_font_size` to `GlyphCacheConfig` (default `128 ppem`) — glyphs above this threshold now bypass the atlas and render directly each frame.

For `min_cached_font_size`: this would involve caching at the clamped size and applying a downscale transform at composite time, plus adjustments to hinting, metrics, and subpixel offsets across all three glyph branches. I think it's a good idea but a bit too involved for this PR. I'll track it as a follow-up. Happy to hear what you think!

Cargo.toml
grebmeg: Added `strip = "none"` to make sure symbols are preserved for profiling.

parley_draw/src/atlas/cache.rs
grebmeg: Removed the `vello_renderer` reference from the `GlyphCache` trait doc comment so the common interface stays renderer-agnostic.

parley_draw/src/renderers/vello_hybrid.rs
grebmeg: Same answer as above — tint isn't part of Vello's `RenderState`, so the explicit `reset_tint()` is needed here.

parley_draw/src/renderers/vello_hybrid.rs
grebmeg: The vello_cpu doesn't need an explicit `GLYPH_PADDING` offset in paint_transform because `atlas_slot.x` / `atlas_slot.y` are already inset by `GLYPH_PADDING` from the allocation origin (see `GlyphAtlas::insert_entry`). So the translate to `(-slot.x, -slot.y)` already moves past the padding. The hybrid backend's image cache resolves to the allocation origin (before padding), so it needs the explicit `(-padding, -padding)` shift.

parley_draw/src/glyph.rs
grebmeg: `Colr` glyphs use fractional `scaled_bbox` dimensions for their area (computed from the glyph's actual bounding box), while outline glyphs use integer `atlas_slot.width`/`height` since they snap to pixel boundaries. The `Rect` carries these fractional bounds so the rendered quad matches the original glyph dimensions rather than the padded integer atlas allocation, which avoids subtle scaling artifacts.

parley_draw/src/renderers/vello_renderer.rs
grebmeg: Oh, nice one! Thanks, can't believe I missed such an easy one. Done!

parley_draw/src/renderers/vello_renderer.rs
grebmeg: Done. `insert_and_render_outline` and `insert_and_render_bitmap` now take an owned `GlyphCacheKey`, and callers use `.take()` to move out of the `Option`. This eliminates the hidden clones on both paths.

parley_draw/src/renderers/vello_renderer.rs
grebmeg: Done. All three `insert_and_render_*` functions take `cache_key: GlyphCacheKey` by value, and callers use `.take()` from the `Option` to move the key in without cloning.

parley_draw/src/atlas/cache.rs
grebmeg: Agreed, the double lookup on miss is a known trade-off of the current get/insert split. I'll keep `raw_entry_mut` with precomputed hashes in mind as a follow-up optimisation. Thanks for flagging it.

parley_draw/src/atlas/cache.rs
grebmeg: Yes, the underlying `AtlasConfig` (from `vello_common`) already has a `max_atlases` field that caps the number of atlas pages. When the limit is reached, `allocate()` fails and the glyph is drawn directly without caching. This is configured on the `ImageCache` side rather than `GlyphCacheConfig`.

parley_draw/src/atlas/commands.rs
grebmeg: Updated to 2026 across all new files in the PR.

parley_draw/src/atlas/commands.rs
grebmeg: Done, moved the detail onto the enum members and trimmed the top-level doc.

parley_draw/src/atlas/commands.rs
grebmeg: Done, added `#[inline]` to all `AtlasCommandRecorder` recording methods.

parley_draw/src/colr.rs
grebmeg: Done, added `#[inline]` to all `ColrRenderer` methods on `AtlasCommandRecorder`.

parley_draw/src/glyph.rs
grebmeg: Great insight! Added a speculative outline cache probe at the top of the per-glyph loop: the transform and key are pure arithmetic so we compute them once upfront, check the cache, and on a hit skip `color_glyphs.get()`/`bitmaps.glyph_for_size()` entirely. On a miss both are reused in the outline branch, so no duplicate work.

parley_draw/src/renderers/vello_cpu.rs
grebmeg: Done

parley_draw/src/renderers/vello_cpu.rs
grebmeg: Done

parley_draw/src/renderers/vello_cpu.rs
grebmeg: Done

parley_draw/src/renderers/vello_cpu.rs
grebmeg: Done

parley_draw/src/renderers/vello_hybrid.rs
grebmeg: Done

parley_draw/src/renderers/vello_hybrid.rs
grebmeg: Done

parley_tests/tests/util/renderer.rs
grebmeg: Done! Added `draw_bitmap_emoji` and `draw_colr_emoji` tests

parley_dev/assets/fonts/noto_color_emoji/README.md
LaurenzV: Ah, whoops, maybe I forgot it as well. :D I guess let's ignore in this case (I don't remember it either, probably used fonttools).

parley_draw/src/atlas/key.rs
LaurenzV: Ah, right. I think it would still be good to have a proper solution for this in the future, but doesn't have to be in this PR. Maybe a TODO?

fontique/src/backend/coretext.rs
```diff
@@ -114,24 +112,26 @@ fn scan_coretext_available_fonts() -> Option<scan::ScannedCollection> {
         }
     }
 
+    // Apple hides certain fonts from CTFontCollection (notably SFNS.ttf, the San Francisco
+    // system UI font). Scanning Library/Fonts directories catches what CoreText omits.
+    paths.extend(library_font_dirs());
+
```
LaurenzV: I thought I left a comment here already but seems like I didn't, this seems like something worth addressing in a separate PR and circumvent it for now if possible? I'm not familiar with CoreText so unfortunately can't really comment much.

parley_draw/src/atlas/cache.rs
```diff
@@ -0,0 +1,696 @@
...
+
+    /// Drain all pending bitmap uploads, keeping the allocation for reuse.
+    pub fn drain_pending_uploads(&mut self) -> vec::Drain<'_, PendingBitmapUpload> {
+        self.pending_uploads.drain(..)
+    }
+
+    /// Drain all pending clear rects, keeping the allocation for reuse.
+    pub fn drain_pending_clear_rects(&mut self) -> vec::Drain<'_, PendingClearRect> {
+        self.pending_clear_rects.drain(..)
+    }
```
LaurenzV: Might be nicer to just return impl Iterator here, but up to you. 🙂

parley_draw/src/glyph.rs
taj-p: > Happy to track this as a follow-up issue if that would be helpful?
However you like 🙏 !

parley_draw/src/atlas/cache.rs
taj-p: > I think it's a good idea but a bit too involved for this PR. I'll track it as a follow-up. Happy to hear what you think!

SGTM!

fontique/src/backend/coretext.rs
taj-p: Looks like it's being addressed in https://github.com/linebender/parley/pull/568. I'll leave it to @grebmeg to decide on whether to wait for that to go in before this or revert the changes to this file

taj-p (approved):
LGTM 🎉 ! Let's goooo!!! I didn't re-read all the code line-by-line again, but skimmed through the comments and their resolutions. All seems absolutely BRILLIANT!

fontique/src/backend/coretext.rs
grebmeg: Yeah, I addressed it there and rebased this branch onto that branch in #568, but I forgot to change the target branch from `main` here. Sorry about that. I think I’ll revert that change here and go ahead with merging without it.

parley_draw/src/atlas/cache.rs
grebmeg: Done, thank you! 🙏

PR #551, opened 2026-02-14
Implement CSS `text-indent` support

### LINKED ISSUE #545: Support `text-indent`
Parley doesn't currently have a way to indent the first line of a paragraph (like CSS `text-indent`).

I tried working around this by inserting a zero-height `InlineBox` at the start of the paragraph as a spacer. This mostly works, but the line breaker unconditionally marks a break opportunity after every inline box:

https://github.com/linebender/parley/blob/df7cedc073e0c5b445dc5bd043f95f730b68dba4/parley/src/layout/line_break.rs#L257-L258

With these settings:

```rust
builder.push_default(StyleProperty::OverflowWrap(OverflowWrap::Anywhere));
builder.push_default(StyleProperty::WordBreak(WordBreakStrength::BreakAll));
```

Filling a line with `,` characters causes the `InlineBox` to be orphaned on a line by itself — the text moves to the next line at `x=0` with no indent. (This doesn't happen with `A` characters.)

A native `text_indent` parameter would avoid this problem. Negative values could give hanging indents, useful for list markers.

devunt:
Closes #545

## Summary

Adds CSS `text-indent` support to parley. The indent is treated as a margin on the start edge of the line box (per CSS spec), reducing available width during line breaking and offsetting content during alignment.

### Design decisions

- **Layout-level property, not `StyleProperty`** — `text-indent` is a block-level CSS property, so it follows the same pattern as `Alignment`: stored in `LayoutData`, configured via a method on `Layout`.
- **API mirrors `Layout::align()`** — `Layout::set_text_indent(amount, options)` separates the core value from behavioral options.
- **Per-line indent stored in `LineData`** — computed once during line breaking, reused during alignment without recalculation.
- **`IndentOptions` in `layout` module** — since text-indent is a layout-level property (not a per-span style), `IndentOptions` lives in `layout/mod.rs` alongside `Alignment`, `AlignmentOptions`, and `ContentWidths`.

### Features

- **Positive indent** — first line offset inward, available width reduced.
- **Negative indent** — first line protrudes beyond the start edge (overflow, matching CSS behavior where negative indent doesn't affect box size).
- **`each-line` keyword** — indent applies after every hard line break (`BreakReason::Explicit`), not just the first line.
- **`hanging` keyword** — inverts which lines are indented: continuation lines instead of first line(s).
- **Alignment integration** — works correctly with `Start`, `Center`, `Right`, and `Justify` alignments. Indent acts as a start-edge margin, and alignment operates within the remaining space.

### API

```rust
layout.set_text_indent(50.0, IndentOptions::default());
layout.set_text_indent(50.0, IndentOptions { each_line: true, ..Default::default() });
layout.set_text_indent(50.0, IndentOptions { hanging: true, ..Default::default() });
layout.break_all_lines(Some(300.0));
layout.align(None, Alignment::Center, AlignmentOptions::default());
```

### How it works

**Line breaking (`line_break.rs`):**
- Before processing each line, determines whether the line should be indented based on `each_line` and `hanging` options.
- Subtracts the indent from `max_advance` to reduce available width for the indented line.
- Stores the computed per-line indent in `LineData::indent`.
- Width calculation in `Drop` adds positive indent to line advance (negative indent causes overflow, not box growth).

**Alignment (`alignment.rs`):**
- For LTR text, sets `offset = indent` (previously always `0`).
- Subtracts indent from `free_space` so alignment algorithms (center, end, justify) operate within the reduced space.
- RTL trailing whitespace handling is preserved.

## Changes

| File | Change |
|------|--------|
| `parley/src/layout/mod.rs` | Add `IndentOptions` struct |
| `parley/src/layout/data.rs` | Add `indent_amount`/`indent_options` to `LayoutData`, `indent` to `LineData` |
| `parley/src/layout/layout.rs` | Add `Layout::set_text_indent()` public API |
| `parley/src/layout/line_break.rs` | Indent calculation in `break_next`, propagation through `try_commit_line`, width accounting in `Drop` |
| `parley/src/layout/alignment.rs` | Indent-aware offset and free space calculation |
| `parley_tests/tests/text_indent.rs` | 22 test cases with snapshot verification |

## Test plan

- [x] `cargo check -p parley` — compiles
- [x] `cargo test -p parley` — 49 passed (existing unit tests unaffected)
- [x] `cargo clippy -p parley -p parley_tests` — no warnings
- [x] `cargo test -p parley_tests` — 125 passed (103 existing + 22 new)

### New test cases

| Test | What it verifies |
|------|--------------------|
| `text_indent_basic` | Positive indent on wrapping text, only first line indented |
| `text_indent_no_wrap` | Single line with indent |
| `text_indent_each_line` | Indent after every hard break (`\n`) |
| `text_indent_hanging` | Continuation lines indented, first line not |
| `text_indent_hanging_each_line` | Combined hanging + each-line |
| `text_indent_negative` | Negative indent (first line protrudes) |
| `text_indent_negative_hanging` | Negative indent on continuation lines |
| `text_indent_center_alignment` | Indent + center alignment |
| `text_indent_right_alignment` | Indent + right alignment |
| `text_indent_justify` | Indent + justified alignment |
| `text_indent_zero` | Zero indent preserves existing behavior |
| `text_indent_line_breaking` | Indent causes additional line breaks |
| `text_indent_rtl_basic` | RTL text with positive indent |
| `text_indent_rtl_hanging` | RTL hanging indent |
| `text_indent_rtl_each_line` | RTL each-line indent |
| `text_indent_rtl_negative` | RTL negative indent |
| `text_indent_rtl_center_alignment` | RTL indent + center alignment |
| `text_indent_rtl_justify` | RTL indent + justified alignment |
| `text_indent_cursor_geometry` | Cursor position accounts for indent |
| `text_indent_hit_testing` | Hit testing in indent area and after indent |
| `text_indent_negative_hit_testing` | Hit testing in negative indent overflow area |
| `text_indent_selection_geometry` | Selection geometry accounts for indent |

### DIFF
--- a/parley/src/layout/layout.rs
+++ b/parley/src/layout/layout.rs
@@ -117,7 +117,7 @@ impl<B: Brush> Layout<B> {
     ///
     /// This must be called before [`Layout::break_all_lines`] or [`Layout::break_lines`],
     /// and before [`Layout::align`].
-    pub fn indent(&mut self, amount: f32, options: IndentOptions) {
+    pub fn set_text_indent(&mut self, amount: f32, options: IndentOptions) {
         self.data.indent_amount = amount;
         self.data.indent_options = options;
     }
--- a/parley_tests/tests/text_indent.rs
+++ b/parley_tests/tests/text_indent.rs
@@ -17,7 +17,7 @@ fn build_indented_layout(
 ) -> parley::Layout<ColorBrush> {
     let builder = env.ranged_builder(text);
     let mut layout = builder.build(text);
-    layout.indent(indent_amount, indent_options);
+    layout.set_text_indent(indent_amount, indent_options);
     layout.break_all_lines(Some(wrap_width));
     layout.align(None, alignment, AlignmentOptions::default());
     layout


parley/src/layout/alignment.rs
```diff
@@ -100,17 +100,19 @@ fn align_impl<B: Brush, const UNDO_JUSTIFICATION: bool>(
 
     // Apply alignment to line items
     for line in &mut layout.lines {
-        line.metrics.offset = 0.;
+        let indent = line.indent;
 
         if is_rtl {
             // In RTL text, trailing whitespace is on the left. As we hang that whitespace, offset
             // the line to the left.
             line.metrics.offset = -line.metrics.trailing_whitespace;
```
nicoburns, line 109: Pretty sure this should be `-line.metrics_trailing_whitespace - indent` for the RTL case. The text is indented from the right for RTL text.

parley/src/layout/line_break.rs
```diff
@@ -203,6 +203,27 @@ impl<'a, B: Brush> BreakLines<'a, B> {
...
+                    || self.lines.lines.last().map(|l| l.break_reason)
+                        == Some(BreakReason::Explicit)
+            } else {
+                self.lines.lines.is_empty()
+            };
+            if self.layout.data.indent_options.hanging {
+                !is_scope_line
+            } else {
+                is_scope_line
+            }
```
nicoburns:
```suggestion
            is_scope_line ^ self.layout.data.indent_options.hanging
```

nicoburns (changes requested):
This mostly looks good to me. But we also need to make sure the RTL case is working correctly. Can we get tests for RTL case? And I think the fix I have suggested for the offset may be required.

We should also check that hit testing / cursor placement / selection works correctly with indented text (including text that overflows the bounds of it's box due to negative text indent).

parley/src/layout/alignment.rs
devunt: I initially applied this suggestion, but it actually causes the text to overflow beyond the left boundary of the box. The issue is that `indent` ends up being subtracted twice — once here in the initial offset, and once in the `free_space` calculation on line 114:

```rust
let free_space = layout.alignment_width - indent - line.metrics.advance
    + line.metrics.trailing_whitespace;
```

For RTL Start alignment (right-aligned), the final offset becomes:

- With `-indent`: `offset = -tw - indent + (aw - indent - adv + tw)` = `aw - 2*indent - adv` ❌
- Without `-indent`: `offset = -tw + (aw - indent - adv + tw)` = `aw - indent - adv` ✓

The second form correctly places the text's right edge at `alignment_width - indent`, which is the intended indent from the right. This is symmetric with the LTR case, where the initial `offset = indent` works because Start/Left alignment doesn't add `free_space`.

I've added RTL snapshot tests and they visually confirm the text stays within bounds without this change.

parley/src/layout/line_break.rs
devunt: Nice trick!

parley/src/layout/layout.rs
```diff
@@ -108,6 +109,19 @@ impl<B: Brush> Layout<B> {
...
 
+    /// Sets the text-indent for the layout.
+    ///
+    /// The indent is applied as a margin on the start edge of indented lines, reducing the
+    /// available width for line breaking and offsetting content during alignment. Negative
+    /// values cause the line to protrude beyond the start edge.
+    ///
+    /// This must be called before [`Layout::break_all_lines`] or [`Layout::break_lines`],
+    /// and before [`Layout::align`].
+    pub fn indent(&mut self, amount: f32, options: IndentOptions) {
```
nicoburns: Could we name this `set_text_indent`. Because unlike the `align` method it doesn't actually apply the indent to the text (it just sets the style).

nicoburns (approved):
I have one suggestion around method naming. Otherwise this LGTM now. But I'll give other maintainers a chance to review before merging.

I've tested that hit testing / mouse selection logic works correctly in Blitz.

PR #553, opened 2026-02-17
attributed_text: add attribute segments with reusable workspace

waywardmonkeys:
Add a new attribute-segmentation API to `attributed_text`:
`AttributeSegmentsWorkspace`, `AttributeSegments`, and `ActiveSpans`.

This computes contiguous, non-overlapping ranges and exposes active
attribute spans in application order for each yielded segment.

Implementation details:
- CSR-style start/end event buffers over sorted boundaries
- allocation-free `next()`
- workspace-backed reuse across repeated segmentations
- active span tracking by attribute index
- `ActiveSpans::iter` / `iter_rev` return `ExactSizeIterator`

Behavior notes:
- inputs beyond `u32` bounds panic
- empty-range attributes are not active but can still split segments

### DIFF
--- a/attributed_text/src/attribute_segments.rs
+++ b/attributed_text/src/attribute_segments.rs
@@ -1,0 +1,502 @@
+// Copyright 2026 the Parley Authors
+// SPDX-License-Identifier: Apache-2.0 OR MIT
+
+//! Attribute-based segmentation for [`AttributedText`].
+//!
+//! Given an [`AttributedText`] with overlapping attribute spans, this module produces
+//! non-overlapping, contiguous segments and provides a view of spans active over each segment.
+
+use alloc::vec::Vec;
+use core::fmt::Debug;
+use core::ops::Range;
+
+use crate::AttributedText;
+use crate::TextStorage;
+
+fn build_segment_state<T: Debug + TextStorage, Attr: Debug>(
+    attributed: &AttributedText<T, Attr>,
+    workspace: &mut AttributeSegmentsWorkspace,
+) {
+    let len = attributed.len();
+    debug_assert!(
+        len <= u32::MAX as usize,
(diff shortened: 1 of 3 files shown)


attributed_text/src/sweep_line.rs
```diff
@@ -0,0 +1,604 @@
...
+            .boundaries
+            .binary_search(&end_u32)
+            .expect("attribute boundary end should be in boundary list");
+        if start_boundary == end_boundary {
+            continue;
+        }
+
+        workspace.span_entries.push(SpanEntry {
+            attr_index: u32::try_from(attr_index).expect("attribute index overflow"),
+        });
```
tomcur: If I'm reading this module right, `Workspace::span_entries` is just a vector of the attribute indices (i.e., just a vector of `0..attributes_len`). It can probably be dropped.

attributed_text/src/sweep_line.rs
```diff
@@ -0,0 +1,604 @@
...
+    active_ids: &'s [u32],
+    span_entries: &'s [SpanEntry],
+    attributed: &'a AttributedText<T, Attr>,
+}
+
+impl<'s, 'a, T: Debug + TextStorage, Attr: Debug> ActiveSpans<'s, 'a, T, Attr> {
+    /// Iterate over the active spans in application order (ascending span id).
+    ///
+    /// Each item is `(&Range<usize>, &Attr)`.
+    pub fn iter(&self) -> impl Iterator<Item = (&'a Range<usize>, &'a Attr)> + '_ {
```
tomcur: This could be `ExactSizeIterator` (same for `iter_rev`).

```suggestion
    pub fn iter(&self) -> impl ExactSizeIterator<Item = (&'a Range<usize>, &'a Attr)> + '_ {
```

PR #559, opened 2026-02-18
parley: make builders emit indexed style runs directly, with tree style-id reuse

waywardmonkeys:
Refactors Parley’s style-lowering pipeline so builder outputs are consumed in a single indexed representation (style_table + style_runs) and removes redundant intermediate style state.

## End state vs main

- `TreeBuilder` and `RangedBuilder` both lower directly into indexed style data used by analysis/shaping/layout.
- The old extra `LayoutContext::styles` intermediate state is removed.
- `TreeStyleBuilder` now reuses style IDs when returning to previously active tree nodes (for example "A <b>B</b> C" reuses the same outer style entry for A and C).
  - `RangedStyleBuilder` lowers directly to style table/runs and uses reusable scratch storage to avoid per-build temporary allocations.
  - Builder regression assertions were updated to compare canonicalized style indexing (semantic equality, not incidental duplicate style slots).

## Motivation

- Clarify architecture: one shared downstream representation for style-aware pipeline stages.
- Reduce duplicate/implicit state and keep style ownership/lowering responsibilities at builder boundaries.
- Enable style sharing where structure already provides it (tree push/pop), without forcing a global interning pass on all callers.

### DIFF
--- a/parley/src/context.rs
+++ b/parley/src/context.rs
@@ -8,7 +8,7 @@ use alloc::{vec, vec::Vec};
 use super::FontContext;
 use super::builder::RangedBuilder;
 use super::resolve::tree::TreeStyleBuilder;
-use super::resolve::{RangedStyle, RangedStyleBuilder, ResolveContext, ResolvedStyle, StyleRun};
+use super::resolve::{RangedStyleBuilder, ResolveContext, ResolvedStyle, StyleRun};
 use super::style::{Brush, TextStyle};
 
 use crate::analysis::{AnalysisDataSources, CharInfo};
@@ -22,7 +22,6 @@ use crate::shape::ShapeContext;
 /// This type is designed to be a global resource with only one per-application (or per-thread).
 pub struct LayoutContext<B: Brush = [u8; 4]> {
     pub(crate) rcx: ResolveContext,
-    pub(crate) styles: Vec<RangedStyle<B>>,
     pub(crate) style_table: Vec<ResolvedStyle<B>>,
     pub(crate) style_runs: Vec<StyleRun>,
     pub(crate) inline_boxes: Vec<InlineBox>,
@@ -44,7 +43,6 @@ impl<B: Brush> LayoutContext<B> {
     pub fn new() -> Self {
         Self {
             rcx: ResolveContext::default(),
-            styles: vec![],
             style_table: vec![],
             style_runs: vec![],
             inline_boxes: vec![],
@@ -152,7 +150,6 @@ impl<B: Brush> LayoutContext<B> {
 
     fn begin(&mut self) {
         self.rcx.clear();
-        self.styles.clear();
         self.style_table.clear();
         self.style_runs.clear();
         self.inline_boxes.clear();


waywardmonkeys:
This can be viewed on  a commit by commit basis. This is groundwork for improved style sharing with #516.

waywardmonkeys:
@nicoburns You might like the changes here to the tree builder.

parley/src/builder.rs
```diff
@@ -155,6 +160,18 @@ fn build_into_layout<B: Brush>(
...
+        lcx.style_table.push(ResolvedStyle::default());
+        lcx.style_runs.push(StyleRun {
+            style_index: 0,
+            range: 0..0,
+        });
+    }
+    assert!(
+        text.is_empty() || !lcx.style_runs.is_empty(),
+        "non-empty text requires at least one style run"
+    );
```
xStrom, line 173: I believe the requirement is general and we just happen to auto-satisfy it for empty text, just before this assert as it happens.

```suggestion
    assert!(
        !lcx.style_runs.is_empty(),
        "at least one style run is required"
    );
```

parley/src/analysis/mod.rs
```diff
@@ -298,24 +307,18 @@ pub(crate) fn analyze_text<B: Brush>(lcx: &mut LayoutContext<B>, mut text: &str)
...
 
     // Line boundaries (word break naming refers to the line boundary determination config).
     //
     // This breaks text into sequences with similar line boundary config (part of style
     // information). If this config is consistent for all text, we use a fast path through this.
-    let Some((first_style, rest)) = lcx.styles.split_first() else {
-        panic!("No style info");
+    let Some((first_style_run, rest_runs)) = lcx.style_runs.split_first() else {
+        return;
     };
```
xStrom: Why remove this panic? With this PR there is an assert before calling `analyze_text` anyway, so this code path never executes. However, if with future code changes we somehow end up here, is a silent return really the correct behavior?

parley/src/tests/utils/asserts.rs
```diff
@@ -3,10 +3,43 @@
...
+            .iter()
+            .position(|existing| existing == style)
+        {
+            remap.push(index as u16);
+        } else {
+            let index = canonical_styles.len() as u16;
+            canonical_styles.push(style.clone());
+            remap.push(index);
+        }
+    }
```
xStrom, line 26: So right now only the `TreeBuilder` deduplicates styles and `RangedBuilder` still generates duplicate styles?

I can understand canonicalizing the sort order of styles, however deduplication is beyond what I would consider canonicalization and is closer to hiding issues from the test.

I know there's also #516 in flight etc, so I'd like to better understand what the state is here and what the plans are. Because conceptually I think both builders could be efficient with style reuse. Maybe not as part of this PR, but then I'd like to at least document that this `canonicalize_layout_data` is a temporary hack to work around that issue.

parley/src/analysis/mod.rs
waywardmonkeys: The entry points take care of this now. (Part of making some stuff not require mutating styles.)

Could make it an `unreachable!()`.

parley/src/tests/utils/asserts.rs
waywardmonkeys: The tree-builder does not de-duplicate strictly speaking. It re-uses from the same nesting level as you pop / push.

The ranged builder does not do that ... I had changes where I used a HashMap to actually de-duplicate and the overhead came out to roughly zero ... but I didn't want to introduce that level of change here (and a canonicalization would still be required for this test because the de-duplicated structure would still be different).

The test still ensures that the end result of the built structure has the same styles now, so the goal of the test is being met.

People who want the best scenario will end up using the future `StyleRunBuilder` which allows the caller to feed things through and handle de-duplication on their own.

#516 is my own PR and that's what led to me submitting this one ... and some other stuff that isn't ready at all where I am working on a model of `styled_text` which *also* de-duplicates a lot of stuff at the higher levels.

parley/src/builder.rs
waywardmonkeys: There used to be something where we had to have non-empty text as well. I don't want to change that type of existing behavior as part of this.

parley/src/analysis/mod.rs
xStrom: Even `unreachable!` would be better than `return` I think, yes. Though I'd prefer the more descriptive panic message, or even just an `unwrap()` over the `unreachable!`.

parley/src/builder.rs
xStrom: My suggested code change does not change any behavior. It merely simplifies the assert and makes the panic message more generic.

There is a check just before this assert that guarantees that empty text has a style run. If that guarantee were to be removed, then my simplified assert would catch that, while the version in this PR would let it incorrectly through, as the next line calls `analyze_text` which expects a style run to exist.

waywardmonkeys: Okay, applying this change in the next push.

parley/src/analysis/mod.rs
waywardmonkeys: Resolving this in the next push as well.

PR #566, opened 2026-02-27
Fix font fallback by computing map_len in fill_cluster_in_place

devunt:
`char_cluster.map_len` is hardcoded to `0`, which causes `Mapper::map()` to short-circuit and return `1.0` (complete match) for every font candidate. This effectively disables all font fallback, as the first candidate always wins regardless of actual glyph coverage. For example, emoji sequences are shaped with the primary text font, producing `.notdef` glyphs instead of falling back to an emoji font.

This fix computes `map_len` as the count of characters where `contributes_to_shaping` is true, matching the semantics expected by `Mapper::map()`.

### DIFF
--- a/parley/src/shape/mod.rs
+++ b/parley/src/shape/mod.rs
@@ -235,6 +235,7 @@ fn fill_cluster_in_place(
 
     let mut force_normalize = false;
     let mut is_emoji_or_pictograph = false;
+    let mut map_len: u8 = 0;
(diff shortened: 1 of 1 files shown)


conor-93 (commented):
Thank you for finding this! 

One suggestion: We can avoid the extra iteration if we count `map_len` in the loop we build `char_cluster.chars` in.

PR #568, opened 2026-03-01
fix: combine CoreText enumeration with Library/Fonts directory scan

grebmeg:
Follow-up to #536 which introduced CoreText-based font enumeration.

That PR replaced the Library/Fonts directory scan with `CTFontCollection::from_available_fonts`, keeping the directory scan only as a fallback for when CoreText fails entirely. However, Apple hides certain fonts from `CTFontCollection`, notably `SFNS.ttf` (San Francisco system UI font), so relying solely on CoreText still leaves gaps.

This PR merges both approaches into a single `scan_system_fonts` function. This ensures full coverage.

### DIFF
--- a/fontique/src/backend/coretext.rs
+++ b/fontique/src/backend/coretext.rs
@@ -8,6 +8,7 @@ use super::{
 use alloc::format;
 use alloc::string::ToString;
 use alloc::sync::Arc;
+use alloc::vec::Vec;
 use core::ptr::{null, null_mut};
 use hashbrown::{HashMap, HashSet};
 use objc2_core_foundation::{
@@ -19,7 +20,7 @@ use objc2_core_text::{
 use objc2_foundation::{
     NSSearchPathDirectory, NSSearchPathDomainMask, NSSearchPathForDirectoriesInDomains,
 };
-use std::path::PathBuf;
+use std::path::{Path, PathBuf};
 
 const DEFAULT_GENERIC_FAMILIES: &[(GenericFamily, &[&str])] = &[
     (GenericFamily::Serif, &["Times", "Times New Roman"]),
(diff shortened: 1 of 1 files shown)


grebmeg:
> The CoreText-based font enumeration has `HashSet`-based path deduplication (I think this is deduplicating the paths of individual font files, not just directories). Could we include the fonts from the `library_font_dirs` in this deduplication? It seems highly likely that a lot of the fonts from `/Library/Fonts` will be duplicates of those from the CoreText enumeration, and I think we should try to avoid processing them twice.
> 
> Also, could we run numbers on the performance (time to enumerate and process fonts) of this PR vs. main?

Good catch! You were right that the previous `library_font_dirs()` was inserting directory paths into the `HashSet` of file paths, so there was zero deduplication happening between the two sources.

Fixed by replacing `library_font_dirs()` with `library_font_files()`, which walks each `Library/Fonts` directory recursively and collects individual file paths. These get inserted into the same `HashSet<PathBuf>`, so any file already discovered by `CoreText` is automatically deduplicated. On my OS, out of 377 files found in `Library/Fonts` dirs, 211 were already known to `CoreText` and deduplicated, while 166 genuinely new files were discovered, including SF system UI fonts (SFNS.ttf, SFCompact.ttf, NewYork.ttf), ~100 Noto Sans fonts, and ~20 STIX math fonts.

Performance numbers (`scan_system_fonts()`, --release, 5 warm runs on Apple Silicon):
||Average|Range
---|---|---
main|	~36 ms|	33–38ms|
this PR|	~47ms|	44–49ms|
delta|	+11ms (+31%)|	|

The extra cost comes from walking the `Library/Fonts` directories to enumerate individual files for `HashSet` deduplication. This is a one-time startup cost.

PR #569, opened 2026-03-03
Add PackTab for `CompositeProps`

taj-p:
## Intent

Adds https://github.com/behdad/packtab.rs for generating `CompositeProps` behind a feature flag.

## Results

I'm seeing roughly 115 kB savings on binary size (building `vello_cpu_render` with and without PackTab): 

image

Overall layout performance seems a touch faster:

image

Tested via:

```shell
cargo export target/benchmarks -- bench --bench=main
cargo bench -q --bench=main --features parley_bench/packtab -- compare target/benchmarks/main -p --time 5
```

These results seem unsurprising considering https://github.com/taj-p/parley/pull/4.

Happy to remove the feature flagging and release to main directly. Also happy to keep it defensively behind a flag for a few weeks.

cc @behdad

### DIFF
--- a/parley_data_gen/src/lib.rs
+++ b/parley_data_gen/src/lib.rs
@@ -75,7 +75,7 @@ pub fn generate(out: std::path::PathBuf, config: &Config) {
     code.push('\n');
     write!(
         code,
-        "#[inline]\npub fn composite_get(cp: u32) -> u32 {{\n    {namespace}_get(cp as usize)\n}}\n"
+        "#[allow(missing_docs, reason = \"packtab generated code\")]\n#[inline]\npub fn composite_get(cp: u32) -> u32 {{\n    {namespace}_get(cp as usize)\n}}\n"
     )
     .unwrap();
 
@@ -87,21 +87,6 @@ pub fn generate(out: std::path::PathBuf, config: &Config) {
         config.compression, config.unsafe_access
     )
     .unwrap();
-    writeln!(&mut file).unwrap();
-    writeln!(&mut file, "#![allow(").unwrap();
-    for lint in [
-        "unsafe_code",
-        "trivial_numeric_casts",
-        "missing_docs",
-        "clippy::allow_attributes_without_reason",
-        "clippy::unseparated_literal_suffix",
-        "clippy::double_parens",
-        "clippy::unnecessary_cast",
-    ] {
-        writeln!(&mut file, "    {lint},").unwrap();
-    }
-    writeln!(&mut file, "    reason = \"packtab generated code\"").unwrap();
-    writeln!(&mut file, ")]").unwrap();
     writeln!(&mut file).unwrap();
     write!(&mut file, "{code}").unwrap();
 }
--- a/parley_data/src/generated/mod.rs
+++ b/parley_data/src/generated/mod.rs
@@ -3,17 +3,6 @@
 
 //! Backing data for composite properties (PackTab, compression=5, unsafe=false)
 
-#![allow(
-    unsafe_code,
-    trivial_numeric_casts,
-    missing_docs,
-    clippy::allow_attributes_without_reason,
-    clippy::unseparated_literal_suffix,
-    clippy::double_parens,
-    clippy::unnecessary_cast,
-    reason = "packtab generated code"
-)]
-
 #[allow(
     dead_code,
     non_upper_case_globals,
@@ -1005,6 +994,7 @@ pub(crate) fn composite_packtab_get(u: usize) -> u32 {
     }
 }
 
+#[allow(missing_docs, reason = "packtab generated code")]
 #[inline]
 pub fn composite_get(cp: u32) -> u32 {
     composite_packtab_get(cp as usize)


parley_data/src/generated_packtab/mod.rs
```diff
@@ -0,0 +1,990 @@
...
+    0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 78, 0, 0, 0, 79, 80, 81, 82, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3,
+    3, 3, 83, 84, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 85, 86, 87, 0, 0, 0, 0, 88, 0, 0,
+    0, 0, 0, 0, 0, 89, 90, 91, 92, 93, 94, 95, 96, 0, 97, 98, 99, 100, 101, 102, 103, 104, 105,
+    106, 107, 108, 109, 110, 111, 112, 113, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2,
+    2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2,
+    2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2,
+    2, 114, 2, 2, 2, 2, 2, 2, 2, 2, 115, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 116, 2, 2, 2, 2, 2, 2, 2, 2,
+    2, 2, 2, 2, 2, 117, 2, 118, 0, 0, 0, 0, 2, 119, 0, 4, 2, 2, 2, 2, 2, 2, 2, 2, 2, 120, 2, 2, 2,
+    2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 121, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
+    0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
```
nicoburns: This data looks very repetitive, which seems weird for compressed data...

.github/workflows/ci.yml
```diff
@@ -395,9 +395,15 @@ jobs:
         with:
           save-if: ${{ github.event_name != 'merge_group' }}
 
-      - name: regenerate unicode data
+      - name: regenerate unicode data (CodePointTrie)
         run: cargo run --locked -p parley_data_gen -- ./parley_data/src/generated
 
+      - name: regenerate unicode data (PackTab)
+        run: cargo run --locked -p parley_data_gen -- ./parley_data/src/generated_packtab --packtab --compression=5
```
nicoburns: Why level 5? Why not level 9? Is it because higher compression levels also affect decompression speed?

parley_data_gen/src/lib.rs
```diff
@@ -18,59 +18,121 @@ use std::io::{BufWriter, Write};
...
+        },
+    );
+
+    if !code.ends_with('\n') {
+        code.push('\n');
+    }
+    code.push('\n');
+    code.push_str(&format!(
+        "#[inline]\npub fn composite_get(cp: u32) -> u32 {{\n    {namespace}_get(cp as usize)\n}}\n"
+    ));
```
nicoburns: Nit: you should be able to `write!(code, ...)` (or `writeln`) directlty into `code`.

parley_data/src/generated_packtab/mod.rs
behdad: PackTab just converts the data to a branchless multi-level lookup table. In this case, the shape of the table (from the comment in the generated code) is: `[2^8,2^5,2^3,2^1]`. That how it is broken down. You still get lots of repetition because eg. all different blocks of 32 numbers are encoded separately.

.github/workflows/ci.yml
behdad: Correct.

In this case, it's unlikely that level 9 generate any different result. There's level 10 that gives you absolute smallest data size with whatever speed comes with it. The level numbers from 1 to 9 tune the heuristic to pick which solution in the tradeoff space (number of lookups vs data size) to pick.

parley_data/src/generated_packtab/mod.rs
nicoburns: Interesting. So presumably it would be possible to further decrease the binary size impact by running the packtab'd data through a general purpose compression algorithm (LZ4, gzip, etc), at the cost of slightly higher RAM usage and a small one-time runtime cost to decompress the data.

behdad: If that kind of minimization is desired, I'm also curious to see what compression=10 does.

nicoburns: If my understanding is correct then the cost of `compression=10` would be paid for every lookup? Whereas if the was compressed, it could be uncompressed into the `compression=5` (or whatever) format at a one-time cost?

behdad: You are correct. I just became curious.

If the primary concern is woff serving, isn't that handled by compression in the transport layer?

taj-p: > If that kind of minimization is desired, I'm also curious to see what compression=10 does.

Performance seems a bit worse and the binary decreased in size by only 128 bytes.

image

> If my understanding is correct then the cost of compression=10 would be paid for every lookup? Whereas if the was compressed, it could be uncompressed into the compression=5 (or whatever) format at a one-time cost?

Great idea! I'll create an issue after this is merged documenting it.

.github/workflows/ci.yml
taj-p: > it's unlikely that level 9 generate any different result

Behdad is right! Levels 5 through 9 produced the same result

parley_data_gen/src/lib.rs
taj-p: Thank you! Updated in https://github.com/linebender/parley/pull/569/commits/6d5e2a83c84dd595bb19603216b7c8c6850ca4b5

taj-p:
> > Is there some of those attributes that packtab itself should generate? We already generate some.
> 
> Oh I meant the rust attributes. I think you call them lints:
> 
> ```
> for lint in [
>         "unsafe_code",
>         "trivial_numeric_casts",
>         "missing_docs",
>         "clippy::allow_attributes_without_reason",
>         "clippy::unseparated_literal_suffix",
>         "clippy::double_parens",
>         "clippy::unnecessary_cast",
>     ] 
> ```


Ah, yes. I think it's probably better for PackTab to generate these.

> As for separate tables or combined, I think if you can look up once, this approach is better in general. Can you tell me which properties you are packing?

- script
- gc
- gib
- bidi
- is_emoji_or_pictograph (`isEmoji || isExtendedPictograph`)
- is_variation_selector
- is_region_indicator
- is_mandatory_linebreak (LineBreak is MandatoryBreak, CarriageReturn, LineFeed, or NextLine)

I think we should also use PackTab for CodePointMap/Trie data here:

https://github.com/linebender/parley/blob/6d5e2a83c84dd595bb19603216b7c8c6850ca4b5/parley/src/analysis/mod.rs#L29-L95

taj-p:
Done in https://github.com/linebender/parley/pull/569/commits/5bf8d2fdb68cb7f23eed12703ec6e13a5124b356! Nice thanks @behdad 🙏 🎉

PR #573, opened 2026-03-12
Extract transform scales into font size whenever possible

LaurenzV:
Based on #572. I don't like having this hack here for too long, but I also don't want to make any larger changes until Alex can look into it properly, so this is the easiest thing we can do for now.

Not entirely sure why the snapshot had to be updated, it's possible that skrifa has different behavior depending on what font size is passed, even if there is no hinting. But it looks harmless.


grebmeg (approved):
I agree with @taj-p that this is an improvement over the current state. LGTM as a temporary fix.

The only update I’d suggest for now is `is_axis_aligned` in `vello_renderer`. At the moment it only checks whether skew is applied, and if it is, it falls back to direct rendering. For now, it might make sense to render any scaled or skewed glyphs directly, except for non-hinted, uniformly scaled glyphs with no skew, where the `font_size` absorbs the scale.

As a next step, I agree we should think about a more principled approach. One option is to include the skew and scale components in the cache key. The downsides are that the key becomes larger, performance will slow down, and the cache could grow much faster than necessary. We might need to quantize the values to mitigate this.

grebmeg:
> Doesn't this mean now that we only update the total_transform (which i though was supposed to be the concatentation of run_transform + glyph_transform), but the two transforms themselves remain unchanged, thus making them out-of-sync?

If we treat (rename) `total_transform` as something like `outline_transform` (or give it a clearer name that reflects how transform and scale are applied) and absorb the `scale` into `font_size`, it may stop looking like an out-of-sync issue. That way we wouldn’t need to think of `run_transform` + `glyph_transform` as producing `total_transform`.

From what I can see, `hinted_size` was introduced, which seems to represent a size that can absorb the scale. In `draw_glyphs`, `unhinted_transform` is calculated as the multiplication of `run_transform` and `glyph_transform`.

I also looked into the `draw_glyphs` to check whether we can bake the scale into `font_size` and keep the remaining transforms as-is, and whether that would affect rendering. It seems mostly fine, except that we currently don’t disable caching for skewed or non-uniformly scaled glyphs, so those still get cached. That said, the current transforms and sizes are quite hard to follow, with multiple code paths across different glyph types, so improving this would definitely help.

These changes were introduced in #542, maybe @valadaptive can help clarify how best to reconcile this?

PR #575, opened 2026-03-17
Add `break_line_with_next`

taj-p:
We have been using this `break_line_with_next` breaking algorithm in production for a couple months now. The use case is that sometimes the consumer knows how many characters they want to break each line at (rather than specifying a width by some pixel dimension). (It's also significantly faster than the current implementation for obvious reasons - no need to calculate and bookkeep intermediary line length state). 

Happy to change the name to something like `break_line_with_max_chars` or similar.

See https://xi.zulipchat.com/#narrow/channel/205635-parley/topic/Floated.20boxes/near/564353468


parley/src/layout/line_break.rs
```diff
@@ -504,6 +653,26 @@ impl<'a, B: Brush> BreakLines<'a, B> {
         }
     }
 
+    #[inline]
+    fn resolve_indent(&mut self) -> f32 {
```
nicoburns, line 657: Does this need to be `&mut`?

parley_tests/tests/line_break.rs
```diff
@@ -0,0 +1,386 @@
...
+    breaker.break_next_with_length(2);
+    breaker.break_next_with_length(2);
+    breaker.break_next_with_length(2);
+    breaker.finish();
+
+    assert_eq!(layout.len(), 3, "Expected 3 lines");
+    env.check_layout_snapshot(&layout);
+}
+
+/// This test verifies that breaking in the middle of a ligature produces valid glyphs.
```
nicoburns, line 141: > This test verifies that breaking in the middle of a ligature produces valid glyphs.

It sounds like it actually verifies that breaking in the middle of a ligature *doesn't* produce valid glyphs? In the other line breaker we just break early in leiu of reshaping. I guess you have reasons for not wanting to do that?

parley_tests/tests/line_break.rs
```diff
@@ -0,0 +1,386 @@
...
+//! Line breaking tests.
+//!
+//! These tests validate `break_next_with_length`, which breaks lines by character count
+//! rather than advance width.
+
+use crate::test_name;
+use crate::util::TestEnv;
+use parley::{Alignment, AlignmentOptions, InlineBox, PositionedLayoutItem, StyleProperty};
+
+#[test]
```
nicoburns, line 13: I'd love to see a test that includes some emoji. They seems like a case that might easily break if not handled correctly.

parley/src/layout/line_break.rs
```diff
@@ -461,6 +448,168 @@ impl<'a, B: Brush> BreakLines<'a, B> {
...
+    /// Each inline box also counts as 1 character.
+    /// Ligature components each count separately (matching character count).
+    ///
+    /// Unlike `break_next`, this method does not respect normal line break opportunities and
+    /// will break exactly when the character limit is reached. It does not break on newlines, for example.
+    pub fn break_next_with_length(&mut self, max_chars: u32) -> Option<()> {
+        if self.done {
+            return None;
+        }
+        self.prev_state = Some(self.state.clone());
```
nicoburns, line 464: Is this necessary? I would expect this line breaker to never need to revert?

parley/src/layout/line_break.rs
```diff
@@ -461,6 +448,168 @@ impl<'a, B: Brush> BreakLines<'a, B> {
...
+                )
+            };
+        }
+
+        let item_count = self.layout.data.items.len();
+        while self.state.item_idx < item_count {
+            let item = &self.layout.data.items[self.state.item_idx];
+
+            match item.kind {
+                LayoutItemKind::InlineBox => {
```
nicoburns, line 490: Does supporting inline boxes in this mode make sense? (are you using them?) I guess if you carefully control the size of the boxes it would work decently well..

parley/src/layout/line_break.rs
```diff
@@ -461,6 +448,168 @@ impl<'a, B: Brush> BreakLines<'a, B> {
...
+                        }
+
+                        let whitespace = cluster.info().whitespace();
+                        let is_newline = whitespace == Whitespace::Newline;
+                        let is_space = whitespace.is_space_or_nbsp();
+                        let advance = cluster.advance();
+
+                        // Compute the x position.
+                        // Newlines don't contribute to line width (matching break_next behavior).
+                        let next_x = if is_newline {
```
nicoburns, line 554: Would it make sense to strip out newlines before passing the string into Parley? That's what Blitz is doing. I guess there's a general conversation to be had about whitespace collapsing.

parley/src/layout/line_break.rs
taj-p: No! Oops!!

parley_tests/tests/line_break.rs
taj-p: Yup! My bad on this phrasing.

> In the other line breaker we just break early in leiu of reshaping.

👀 ! Maybe not for long! There is work being planned to make this more correct - to reshape after breaking if it is required (for Arabic / ligatures / and so on)!

> I guess you have reasons for not wanting to do that?

We actually don't like this behaviour - we want to reshape after breaking, but that isn't currently supported

parley/src/layout/line_break.rs
taj-p: I'm not using them. Happy to change this behaviour to something else?

parley/src/layout/line_break.rs
taj-p: Yes, I'm not sure TBH! We might need to talk about it

parley/src/layout/line_break.rs
nicoburns: I'd be tempted to ignore them entirely in this mode. But I guess it's also harmless.

parley/src/layout/line_break.rs
nicoburns: I'm surprised the linter didn't pick this one up

PR #576, opened 2026-03-17
Fix LRU cache epoch not incremented on hits, breaking eviction

taj-p:
There was a bug in my implementation. We need to increment the epoch on cache hits, not just cache misses. Otherwise, we may not evict the LRU.

### DIFF
--- a/parley/src/lru_cache.rs
+++ b/parley/src/lru_cache.rs
@@ -38,23 +38,14 @@ impl<ID, T> LruCache<ID, T> {
     where
         K: Equivalent<ID> + Into<ID>,
     {
-        match self.find_entry(id, make_data) {
-            (true, index) => {
-                self.epoch += 1;
-                let entry = &mut self.entries[index];
-                entry.epoch = self.epoch;
-                &entry.data
-            }
-            (false, index) => {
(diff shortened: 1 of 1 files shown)


nicoburns (approved):
Good catch. This PR also wins the award for the best test I've seen in while! :)

parley/src/lru_cache.rs
conor-93, line 46: These arms seem identical now, could this be simplified to:
```rust
pub(crate) fn entry<K>(&mut self, id: K, make_data: impl FnOnce() -> T) -> &T
where
    K: Equivalent<ID> + Into<ID>,
{
    let (_, index) = self.find_entry(id, make_data);
    self.epoch += 1;
    let entry = &mut self.entries[index];
    entry.epoch = self.epoch;
    &entry.data
}
```

And if so, `find_entry` would only need to return `usize`.

conor-93 (approved):
Nice, thank you! One comment

parley/src/lru_cache.rs
taj-p: Amazing! Thank you! Implemented in c9d95ae5a2c8bc3bb73e12bfd041d3d76b13f3f7

PR #579, opened 2026-03-18
implement vertical align

jkelleyrtp:


parley/src/layout/line.rs
```diff
@@ -253,12 +253,23 @@ impl<'a, B: Brush> Iterator for GlyphRunIter<'a, B> {
...
+                    // Get baseline_offset from the corresponding LineItemData
+                    let abs_item_idx = self.line.data.item_range.start + self.item_index;
+                    let baseline_offset = self
+                        .line
+                        .layout
+                        .data
+                        .line_items
+                        .get(abs_item_idx)
+                        .map(|li| li.baseline_offset)
+                        .unwrap_or(0.0);
```
nicoburns: This probably ought to be a method on `Line` or `Layout`?

parley/src/inline_box.rs
```diff
@@ -1,6 +1,8 @@
 // Copyright 2024 the Parley Authors
 // SPDX-License-Identifier: Apache-2.0 OR MIT
 
+use crate::VerticalAlign;
+
 /// A box to be laid out inline with text
 #[derive(PartialEq, Debug, Clone)]
 pub struct InlineBox {
```
nicoburns, line 8: Strictly speaking, this is orthogonal to this PR (but it may be useful to implement it here anyway, and it could also land before this PR), but:

InlineBox ought to gain a first_baseline: Option<f32> field. If that is Some then the inline box ought to be aligned using that as it's baseline. If it is None then it would fallback to aligning using the full height of the box (aligning to the bottom of the box), which is the current behaviour.

In Blitz, we'd source this value from Taffy and set it here before running text layout.

parley/src/layout/line_break.rs
```diff
@@ -693,63 +693,174 @@ impl<'a, B: Brush> BreakLines<'a, B> {
...
+
+                    // Skip trailing whitespace for metrics contribution
+                    if is_trailing_ws {
+                        continue;
+                    }
+
+                    // Contribute to line metrics considering the offset
+                    if !matches!(vertical_align, VerticalAlign::Top | VerticalAlign::Bottom) {
+                        let effective_ascent = (run.metrics.ascent - offset).max(0.0);
+                        let effective_descent = (run.metrics.descent + offset).max(0.0);
```
nicoburns, line 976: Not 100% sure, but I think `offset` should be in the same direction for both of these calculations.

parley/src/layout/line_break.rs
```diff
@@ -693,63 +693,174 @@ impl<'a, B: Brush> BreakLines<'a, B> {
...
+                        VerticalAlign::TextTop => -(line.metrics.ascent - item.height),
+                        VerticalAlign::TextBottom => line.metrics.descent,
+                        // Top/Bottom deferred to pass 2
+                        VerticalAlign::Top | VerticalAlign::Bottom => 0.0,
+                    };
+                    line_item.baseline_offset = offset;
+
+                    // Contribute to line metrics considering the offset.
+                    // Inline boxes sit with bottom at baseline by default,
+                    // so effective_ascent = height - offset, effective_descent = offset
```
nicoburns: We need to add a `first_baseline: Option<f32>` field to inline boxes so in the case that the inline box internally contains text, we can align with the first line of text's baseline rather than the bottom of the box.

parley/src/layout/line_break.rs
```diff
@@ -693,63 +693,174 @@ impl<'a, B: Brush> BreakLines<'a, B> {
...
+                        VerticalAlign::Middle => {
+                            let x_height = run
+                                .metrics
+                                .x_height
+                                .unwrap_or(run.metrics.ascent * 0.5);
+                            -(x_height * 0.5)
+                        }
+                        VerticalAlign::Length(v) => -v,
+                        VerticalAlign::TextTop => -(line.metrics.ascent - run.metrics.ascent),
+                        VerticalAlign::TextBottom => line.metrics.descent - run.metrics.descent,
```
nicoburns: We haven't computed `line.metrics.ascent` here yet? We only finish computing that at the end of this loop.

parley/src/layout/line_break.rs
```diff
@@ -693,63 +693,174 @@ impl<'a, B: Brush> BreakLines<'a, B> {
...
+                    if is_trailing_ws {
+                        continue;
+                    }
+
+                    // Contribute to line metrics considering the offset
+                    if !matches!(vertical_align, VerticalAlign::Top | VerticalAlign::Bottom) {
+                        let effective_ascent = (run.metrics.ascent - offset).max(0.0);
+                        let effective_descent = (run.metrics.descent + offset).max(0.0);
+                        line.metrics.ascent = line.metrics.ascent.max(effective_ascent);
+                        line.metrics.descent = line.metrics.descent.max(effective_descent);
```
nicoburns, line 978: I don't think this is sufficient in the general case. As text spans form a tree structure, you can end up with nested offsets (imagine a superscript of a superscript). See https://github.com/linebender/parley/issues/291

Perhaps this is landable like this though? I guess we'd need to test how it performs in practice.

parley/src/layout/line_break.rs
```diff
@@ -693,63 +693,174 @@ impl<'a, B: Brush> BreakLines<'a, B> {
...
+                    // Compute baseline_offset (positive = down)
+                    let offset = match vertical_align {
+                        VerticalAlign::Baseline => 0.0,
+                        VerticalAlign::Sub => run.metrics.descent,
+                        VerticalAlign::Super => -(run.metrics.ascent * 0.4),
+                        VerticalAlign::Middle => {
+                            let x_height = run
+                                .metrics
+                                .x_height
+                                .unwrap_or(run.metrics.ascent * 0.5);
```
nicoburns: CSS spec suggests a fallback of `font_size * 0.5` https://www.w3.org/TR/css-inline-3/#baseline-synthesis-fonts

parley/src/layout/line_break.rs
```diff
@@ -693,63 +693,174 @@ impl<'a, B: Brush> BreakLines<'a, B> {
...
                     let run = &self.layout.data.runs[line_item.index];
-                    line.metrics.ascent = line.metrics.ascent.max(run.metrics.ascent);
-                    line.metrics.descent = line.metrics.descent.max(run.metrics.descent);
 
-                    // Mark us as having seen non-whitespace content on this line
+                    // Compute baseline_offset (positive = down)
+                    let offset = match vertical_align {
+                        VerticalAlign::Baseline => 0.0,
+                        VerticalAlign::Sub => run.metrics.descent,
+                        VerticalAlign::Super => -(run.metrics.ascent * 0.4),
```
nicoburns: These really ought to be font-specific values taken from the font's `BASE` table (in `Layout::push_run`). And added to the run metrics like `x_height` currently is.

nicoburns (changes requested):
Some initial feedback

nicoburns:
I think this needs to be split into a few separate PRs:

1. Adding the new `AlignmentBaseline` and `BaselineShift` styles (we may want to skip `BaselineSource` initially)
2. Implementing the vertical alignment algorithm in `line_break.rs` (initial version should probably skip the "strut metrics")
3. Extracting additional baseline metrics from the font.
4. Adding baseline support to inline boxes
5. Root level metrics (aka strut metrics) - this bit is currently adding a lot of complexity, and probably needs some design work. I suggest we try and land the other stuff first.


1, 2, and potentially 3 could potentially all be in the first PR.

nicoburns:
This may be useful for exactly matching browser https://github.com/w3c/csswg-drafts/issues/5225

nicoburns:
Superceded by https://github.com/linebender/parley/pull/766

PR #581, opened 2026-03-24
fix: Prevent glyph cache collisions between outline, colr, and bitmap entries

grebmeg:
### Problem
COLR and bitmap glyph cache entries use `subpixel_x: 0`, `hinted: false`, and default/black context colors — all of which are valid values for an unhinted outline glyph rendered at the first subpixel bucket with a black context color. When such an outline glyph shares the same font, glyph id, size, and variation coordinates with a COLR or bitmap glyph, their cache keys become identical, causing one to silently overwrite the other and producing incorrect rendering.

### Root Cause
The `subpixel_x` field in `GlyphCacheKey` was being used for two unrelated purposes: subpixel positioning for outlines and a hardcoded zero for non-outline glyph types. Since `quantize_subpixel` can return 0, and COLR/bitmap entries also set `hinted: false` and use black as the context color, which are both plausible values for outline glyphs, there is nothing left in the key to distinguish glyph types. The key spaces overlap whenever all other fields (font id, font index, glyph id, size, variation coords) match.

### Solution
Introduce two sentinel constants: `SUBPIXEL_COLR` (= `SUBPIXEL_BUCKETS`) and `SUBPIXEL_BITMAP` (= `SUBPIXEL_BUCKETS + 1`), that sit just above the valid quantization range. COLR and bitmap cache entries now use these sentinels instead of 0, guaranteeing distinct keys for every glyph type regardless of the hinting, color, or subpixel values used by outline entries.

### DIFF
--- a/parley_draw/src/glyph.rs
+++ b/parley_draw/src/glyph.rs
@@ -349,7 +349,6 @@ impl<'a, 'b, Glyphs: Iterator<Item = Glyph> + Clone, C: GlyphCache>
 
                 // COLR glyphs are never hinted and have no sub-pixel offset;
                 // context_color is part of the key because it affects painted layers.
-                // SUBPIXEL_COLR prevents collisions with outline keys (see key.rs).
                 let cache_key = cache_enabled.then(|| GlyphCacheKey {
                     font_id,
                     font_index,
@@ -427,7 +426,6 @@ impl<'a, 'b, Glyphs: Iterator<Item = Glyph> + Clone, C: GlyphCache>
 
                 // Bitmaps are not hinted and have no sub-pixel offset or
                 // context color; variation coords are irrelevant for fixed strikes.
-                // SUBPIXEL_BITMAP prevents collisions with outline/COLR keys (see key.rs).
                 let cache_key = cache_enabled.then(|| GlyphCacheKey {
                     font_id,
                     font_index,


parley_draw/src/glyph.rs
```diff
@@ -349,13 +349,14 @@ impl<'a, 'b, Glyphs: Iterator<Item = Glyph> + Clone, C: GlyphCache>
 
                 // COLR glyphs are never hinted and have no sub-pixel offset;
                 // context_color is part of the key because it affects painted layers.
+                // SUBPIXEL_COLR prevents collisions with outline keys (see key.rs).
```
LaurenzV: Up to you, but I think we have enough documentation in other places already which makes this comment a bit superfluous.

parley_draw/src/glyph.rs
```diff
@@ -426,13 +427,14 @@ impl<'a, 'b, Glyphs: Iterator<Item = Glyph> + Clone, C: GlyphCache>
 
                 // Bitmaps are not hinted and have no sub-pixel offset or
                 // context color; variation coords are irrelevant for fixed strikes.
+                // SUBPIXEL_BITMAP prevents collisions with outline/COLR keys (see key.rs).
```
LaurenzV: Same here.

parley_draw/src/glyph.rs
grebmeg: Make sense to me! 

Done, removed both.

PR #586, opened 2026-03-26
Move to `glifo` and remove `parley_draw`

laurenz-canva:
Accompanying vello commit: https://github.com/linebender/vello/pull/1539

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -50,7 +50,7 @@ parley = { version = "0.8.0", default-features = false, path = "parley" }
 parley_data = { version = "0.8.0", path = "parley_data", default-features = false }
 parley_data_gen = { path = "parley_data_gen" }
 parley_dev = { default-features = false, path = "parley_dev" }
-glifo = { git = "https://github.com/linebender/vello", rev = "9b070fd", default-features = false }
+glifo = { git = "https://github.com/linebender/vello", rev = "5796226", default-features = false }
 peniko = { version = "0.6.0", default-features = false }
 read-fonts = { version = "0.37.0", default-features = false }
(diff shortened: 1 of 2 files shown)


Cargo.toml
```diff
@@ -109,9 +108,10 @@ clippy.wildcard_dependencies = "warn"
 # END LINEBENDER LINT SET
 
 [patch.crates-io]
-vello_common = { git = "https://github.com/linebender/vello.git", rev = "190f5a2b77ca2433f7ec5781b2855a7b4625b8f4" }
-vello_cpu = { git = "https://github.com/linebender/vello.git", rev = "190f5a2b77ca2433f7ec5781b2855a7b4625b8f4" }
-vello_hybrid = { git = "https://github.com/linebender/vello.git", rev = "190f5a2b77ca2433f7ec5781b2855a7b4625b8f4" }
+vello_common = { git = "https://github.com/linebender/vello.git", rev = "de56b48" }
+vello_cpu = { git = "https://github.com/linebender/vello.git", rev = "de56b48" }
+vello_hybrid = { git = "https://github.com/linebender/vello.git", rev = "de56b48" }
+
```
xStrom: Given that you updated these to 0.0.7 already, should this patch even remain? Perhaps it can be removed.

nicoburns: I think we'll need to depend on the git version of vello, otherwise it won't match the version of vello that `glifo` is depending on. But it could probably be a regular git dep rather than a patch (as we won't be publishing any of the crates with this dependency).

LaurenzV: As you want! Just did it to keep it consistent with how it's currently done.

PR #595, opened 2026-03-29
editing: move text input to document-space composition

waywardmonkeys:
Rework `PlainEditor`’s text-input/composition model so host integrations can operate in document UTF-8 byte space instead of emulating Parley’s old raw preedit internals.

This changes the internal composition model from a single hidden-preedit span to two distinct states:

* hidden preedit text, which is excluded from `PlainEditor::text()`
* visible composing regions over existing document text, which remain part of `PlainEditor::text()`

On top of that model, add a host-facing editing surface on `PlainEditorDriver` for:

* setting document selection
* setting composing regions
* insert/replace with optional replacement ranges
* atomic composition updates
* surrounding delete in document space
* explicit composition commit/clear
* semantic delete-to-edge helpers for future edit-command integration

Move encoded-offset conversion over visible editor text into Parley itself via `SplitString::to_utf8_range` and `TextIndexEncoding`, so adapters do not need to allocate or duplicate range-conversion logic.

Also:

* remove the legacy mutating compose APIs
* replace `raw_compose` with `composition`
* rename `ime_cursor_area` to `text_input_area`
* migrate `examples/vello_editor` to the `ui-events` text-input path

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,17 +20,17 @@ This release has an [MSRV] of 1.88.
 
 #### Parley
 
-- `PlainEditorDriver` now exposes UTF-8 document-range editing helpers for host-side text input and composition handling, plus delete-to-edge helpers for semantic edit-command integration.
-- `PlainEditorDriver::commit_composition` to explicitly commit an active composition.
-- `SplitString::to_utf8_range` and `TextIndexEncoding` to convert encoded offsets over visible editor text without allocating.
+- `PlainEditorDriver` now exposes UTF-8 document-range editing helpers for host-side text input and composition handling, plus delete-to-edge helpers for semantic edit-command integration. ([#595][] by [@waywardmonkeys][])
+- `PlainEditorDriver::commit_composition` to explicitly commit an active composition. ([#595][] by [@waywardmonkeys][])
+- `SplitString::to_utf8_range` and `TextIndexEncoding` to convert encoded offsets over visible editor text without allocating. ([#595][] by [@waywardmonkeys][])
 
 ### Changed
 
 #### Parley
 
-- Breaking change: removed legacy `PlainEditorDriver` composition mutation methods `set_compose`, `set_compose_byte_range`, `clear_compose`, and `finish_compose` in favor of the document-space editing API.
-- Breaking change: `PlainEditor::raw_compose` has been replaced by `PlainEditor::composition`, which exposes composing text and its document offset instead of raw-buffer byte ranges.
-- Breaking change: `PlainEditor::ime_cursor_area` has been renamed to `PlainEditor::text_input_area`.
+- Breaking change: removed legacy `PlainEditorDriver` composition mutation methods `set_compose`, `set_compose_byte_range`, `clear_compose`, and `finish_compose` in favor of the document-space editing API. ([#595][] by [@waywardmonkeys][])
+- Breaking change: `PlainEditor::raw_compose` has been replaced by `PlainEditor::composition`, which exposes composing text and its document offset instead of raw-buffer byte ranges. ([#595][] by [@waywardmonkeys][])
+- Breaking change: `PlainEditor::ime_cursor_area` has been renamed to `PlainEditor::text_input_area`. ([#595][] by [@waywardmonkeys][])
 
 ## [0.8.0] - 2026-03-27
 
@@ -578,6 +578,7 @@ This release has an [MSRV][] of 1.70.
 [#578]: https://github.com/linebender/parley/pull/578
 [#589]: https://github.com/linebender/parley/pull/589
 [#594]: https://github.com/linebender/parley/pull/594
+[#595]: https://github.com/linebender/parley/pull/595
 
 [Unreleased]: https://github.com/linebender/parley/compare/v0.8.0...HEAD
 [0.8.0]: https://github.com/linebender/parley/compare/v0.7.0...v0.8.0
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,20 @@ This release has an [MSRV] of 1.88.
 
 - `FallbackKey` now implements `From<Script>` for convenience. ([#594][] by [@xStrom][])
 
+#### Parley
+
+- `PlainEditorDriver` now exposes UTF-8 document-range editing helpers for host-side text input and composition handling, plus delete-to-edge helpers for semantic edit-command integration.
+- `PlainEditorDriver::commit_composition` to explicitly commit an active composition.
+- `SplitString::to_utf8_range` and `TextIndexEncoding` to convert encoded offsets over visible editor text without allocating.
+
+### Changed
+
+#### Parley
+
+- Breaking change: removed legacy `PlainEditorDriver` composition mutation methods `set_compose`, `set_compose_byte_range`, `clear_compose`, and `finish_compose` in favor of the document-space editing API.
+- Breaking change: `PlainEditor::raw_compose` has been replaced by `PlainEditor::composition`, which exposes composing text and its document offset instead of raw-buffer byte ranges.
+- Breaking change: `PlainEditor::ime_cursor_area` has been renamed to `PlainEditor::text_input_area`.
+
 ## [0.8.0] - 2026-03-27
 
 This release has an [MSRV] of 1.88.
--- a/parley/src/editing/editor.rs
+++ b/parley/src/editing/editor.rs
@@ -319,7 +319,11 @@ where
         }
     }
 
-    /// Delete the selection or to the start of the physical line.
+    /// Delete the current selection, or delete backward to the start of the
+    /// current physical line when the selection is collapsed.
+    ///
+    /// This corresponds to semantic editing commands such as AppKit's
+    /// `deleteToBeginningOfLine:`.
     pub fn delete_to_line_start(&mut self) {
         self.refresh_layout();
         if self.editor.selection.is_collapsed() {
@@ -335,7 +339,11 @@ where
         }
     }
 
-    /// Delete the selection or to the end of the physical line.
+    /// Delete the current selection, or delete forward to the end of the
+    /// current physical line when the selection is collapsed.
+    ///
+    /// This corresponds to semantic editing commands such as AppKit's
+    /// `deleteToEndOfLine:`.
     pub fn delete_to_line_end(&mut self) {
         self.refresh_layout();
         if self.editor.selection.is_collapsed() {
@@ -351,7 +359,8 @@ where
         }
     }
 
-    /// Delete the selection or to the beginning of the document.
+    /// Delete the current selection, or delete backward to the beginning of
+    /// the document when the selection is collapsed.
     pub fn delete_to_text_start(&mut self) {
         if self.editor.selection.is_collapsed() {
             let end = self.editor.selection.focus().index();
@@ -362,7 +371,8 @@ where
         }
     }
 
-    /// Delete the selection or to the end of the document.
+    /// Delete the current selection, or delete forward to the end of the
+    /// document when the selection is collapsed.
     pub fn delete_to_text_end(&mut self) {
         if self.editor.selection.is_collapsed() {
             let start = self.editor.selection.focus().index();
@@ -375,19 +385,27 @@ where
     }
 
     /// Insert a newline at the current selection.
+    ///
+    /// This is useful for hosts that surface newline insertion as a semantic
+    /// editing action rather than ordinary text input.
     pub fn insert_newline(&mut self) {
         self.insert_or_replace_selection("\n");
     }
 
     /// Insert a horizontal tab at the current selection.
+    ///
+    /// This is useful for hosts that surface tab insertion as a semantic
+    /// editing action rather than ordinary text input.
     pub fn insert_tab(&mut self) {
         self.insert_or_replace_selection("\t");
     }
 
     /// Set the document selection using UTF-8 byte offsets over [`PlainEditor::text`].
     ///
-    /// Returns `false` if the provided range is reversed, out of bounds, or
-    /// does not land on character boundaries in the document text.
+    /// Returns `false` and leaves the editor unchanged if the provided range is
+    /// reversed, out of bounds, or does not land on UTF-8 scalar-value
+    /// boundaries in the document text. This does not promise grapheme-cluster
+    /// aware selection behavior.
     pub fn set_document_selection(&mut self, range: Range<usize>) -> bool {
         let Some(range) = self.editor.document_range_to_raw_range(range) else {
             return false;
@@ -403,8 +421,12 @@ where
 
     /// Set the composing region using UTF-8 byte offsets over [`PlainEditor::text`].
     ///
-    /// Returns `false` if the provided range is reversed, out of bounds, or
-    /// does not land on character boundaries in the document text.
+    /// Any existing composition, whether hidden preedit or a visible composing
+    /// region, is replaced by this new visible composing region.
+    ///
+    /// Returns `false` and leaves the editor unchanged if the provided range is
+    /// reversed, out of bounds, or does not land on UTF-8 scalar-value
+    /// boundaries in the document text.
     pub fn set_composing_region(&mut self, range: Range<usize>) -> bool {
         let Some(range) = self.editor.document_range_to_raw_range(range) else {
             return false;
@@ -424,7 +446,8 @@ where
     /// [`PlainEditor::text`]. The `selection_in_inserted_text` range is
     /// expressed in UTF-8 byte offsets over `text`.
     ///
-    /// Returns `false` if either range is invalid.
+    /// Returns `false` and leaves the editor unchanged if either range is
+    /// invalid.
     pub fn insert_or_replace(
         &mut self,
         text: &str,
@@ -473,9 +496,14 @@ where
     /// [`PlainEditor::text`]. The `selection_in_composition` range is
     /// expressed in UTF-8 byte offsets over `text`.
     ///
+    /// If `replacement` is provided, that document range is replaced. Otherwise
+    /// the current composition range is replaced if one exists; if not, the
+    /// current selection is replaced.
+    ///
     /// If `text` is empty, this clears any active composition.
     ///
-    /// Returns `false` if either range is invalid.
+    /// Returns `false` and leaves the editor unchanged if either range is
+    /// invalid.
     pub fn update_composition(
         &mut self,
         text: &str,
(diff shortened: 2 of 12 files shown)


parley/src/editing/editor.rs
```diff
@@ -39,48 +33,30 @@ impl Generation {
...
+/// This text may either be transient preedit text excluded from
+/// [`PlainEditor::text`] or an existing document range that is currently marked
+/// as composing.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub struct Composition<'source> {
+    /// The composing text.
+    pub text: &'source str,
+    /// The UTF-8 byte offset in [`PlainEditor::text`] where the composition is
+    /// currently located.
+    pub document_offset: usize,
```
DJMcNab, line 47: Would it make sense to expose whether or not this `Composition` is transient here? I don't have a concrete case where it would be useful, but it feels a bit weird to have "this may be one or the other, and we don't tell you which one".

parley/src/editing/editor.rs
```diff
@@ -343,88 +319,280 @@ where
         }
     }
 
-    // --- MARK: IME ---
-    /// Set the IME preedit composing text.
+    /// Delete the selection or to the start of the physical line.
+    pub fn delete_to_line_start(&mut self) {
```
DJMcNab: This method name and docs combination have some dissonance. I see that it is the behaviour from doing <kbd>⌘</kbd><kbd>⌫</kbd>, at least on this mac. I presume that there's an equivalent macOS event, which we should probably link to?
I unfortunately am struggling to come up with a concise name for "delete to the start of the line, unless the user has a selection in which case ignore that command and clear their selection instead"

parley/src/editing/editor.rs
```diff
@@ -343,88 +319,280 @@ where
         }
     }
 
-    // --- MARK: IME ---
-    /// Set the IME preedit composing text.
+    /// Delete the selection or to the start of the physical line.
```
DJMcNab: I like the term physical line here; I'd previously been using "soft line" for it.

parley/src/editing/editor.rs
DJMcNab: If there's a commonly implemented shortcut for this on Windows/Linux, even if it isn't interpreted by the system, we should probably mention this here.

parley/src/editing/editor.rs
```diff
@@ -343,88 +319,280 @@ where
...
+
+    /// Insert a newline at the current selection.
+    pub fn insert_newline(&mut self) {
+        self.insert_or_replace_selection("\n");
+    }
+
+    /// Insert a horizontal tab at the current selection.
+    pub fn insert_tab(&mut self) {
+        self.insert_or_replace_selection("\t");
+    }
```
DJMcNab, line 401: These methods are required just because enter and tab are special keys which are treated as buttons rather than text input, right?

parley/src/editing/editor.rs
```diff
@@ -343,88 +319,280 @@ where
...
+
+    /// Set the document selection using UTF-8 byte offsets over [`PlainEditor::text`].
     ///
-    /// This starts composing. Composing is reset by calling [`clear_compose`](Self::clear_compose).
-    /// Alternatively, the preedit text can be committed by calling [`finish_compose`](Self::finish_compose).
+    /// Returns `false` if the provided range is reversed, out of bounds, or
+    /// does not land on character boundaries in the document text.
```
DJMcNab: I presume this is utf-8 character boundaries. If so, we should specify behaviour around landing inside grapheme clusters (or whatever the right term for "reasonable selection boundaries" is).

parley/src/editing/editor.rs
```diff
@@ -343,88 +319,280 @@ where
...
+        true
+    }
+
+    /// Set the composing region using UTF-8 byte offsets over [`PlainEditor::text`].
     ///
-    /// The selection and preedit region can be manipulated independently while composing
-    /// is active.
+    /// Returns `false` if the provided range is reversed, out of bounds, or
+    /// does not land on character boundaries in the document text.
+    pub fn set_composing_region(&mut self, range: Range<usize>) -> bool {
```
DJMcNab, line 430: What happens to the old pre-edit style composing region if there is one?

parley/src/editing/editor.rs
```diff
@@ -343,88 +319,280 @@ where
...
+    /// The `replacement` range is expressed in UTF-8 byte offsets over
+    /// [`PlainEditor::text`]. The `selection_in_inserted_text` range is
+    /// expressed in UTF-8 byte offsets over `text`.
     ///
-    /// The selection is updated based on `cursor`, which contains the byte offsets relative to the
-    /// start of the preedit text. If `cursor` is `None`, the selection and caret are hidden.
-    pub fn set_compose(&mut self, text: &str, cursor: Option<(usize, usize)>) {
-        debug_assert!(!text.is_empty());
-        debug_assert!(cursor.map(|cursor| cursor.1 <= text.len()).unwrap_or(true));
+    /// Returns `false` if either range is invalid.
```
DJMcNab: (and does nothing?)

parley/src/editing/editor.rs
```diff
@@ -343,88 +319,280 @@ where
...
+    /// Apply a composition update atomically.
+    ///
+    /// The `replacement` range is expressed in UTF-8 byte offsets over
+    /// [`PlainEditor::text`]. The `selection_in_composition` range is
+    /// expressed in UTF-8 byte offsets over `text`.
+    ///
+    /// If `text` is empty, this clears any active composition.
+    ///
+    /// Returns `false` if either range is invalid.
+    pub fn update_composition(
```
DJMcNab, line 507: I'm finding this method a bit hard to understand, even reading the implementation. In particular, the case where the replacement range is some but the text is empty is hard for me to wrap my head around.

This is basically for handling the case where you get a new provided preedit, right?

parley/src/editing/editor.rs
```diff
@@ -343,88 +319,280 @@ where
...
+        self.editor.compose = Some(ActiveComposition {
+            range: start..start + text.len(),
+            kind: CompositionKind::HiddenPreedit,
+        });
+        self.editor.show_cursor = selection_in_composition.is_some();
         self.update_layout();
 
-        // Select the location indicated by the IME. If `cursor` is none, collapse the selection to
-        // a caret at the start of the preedit text. As `self.editor.show_cursor` is `false`, it
-        // won't show up.
```
DJMcNab, line 385: I think that at least the last sentence of this comment is still valuable.

parley/src/editing/editor.rs
```diff
@@ -343,88 +319,280 @@ where
...
+            self.editor.cursor_at(start + selection.start),
+            self.editor.cursor_at(start + selection.end),
         ));
+        true
     }
 
-    /// Set the preedit range to a range of byte indices.
-    /// This leaves the selection and cursor unchanged.
+    /// Delete the requested number of UTF-8 bytes surrounding the current
+    /// document selection.
```
DJMcNab, line 553: This also deletes the current selection, right?

parley/src/editing/editor.rs
```diff
@@ -343,88 +319,280 @@ where
...
+        match compose.kind {
+            CompositionKind::HiddenPreedit => {
+                self.editor.buffer.replace_range(compose.range.clone(), "");
+                self.update_layout();
+                self.editor
+                    .set_selection(self.editor.cursor_at(compose.range.start).into());
+            }
+            CompositionKind::VisibleRegion => {
+                self.update_layout();
+            }
```
DJMcNab, line 610: I see a case on macOS where a (seemingly) visible region preedit gets collapsed to the old text. If I hold e.g. a to add an accent, then move my cursor within the text (e.g. with the mouse), I see the <ins>á</ins> revert to "a". Note that the character gets committed if I move focus away from that specific text area.

DJMcNab: However, this is treated as visible composition text. For example, using it in spotlight, changing it to æ changes the results.

parley/src/editing/editor.rs
```diff
@@ -1232,8 +1483,18 @@ where
...
+            if compose.range.is_empty() {
+                self.layout = builder.build(&self.buffer);
+                self.layout.break_all_lines(self.width);
+                self.layout
+                    .align(self.width, self.alignment, AlignmentOptions::default());
+                self.selection = self.selection.refresh(&self.layout);
+                self.layout_dirty = false;
+                self.generation.nudge();
+                return;
+            }
```
DJMcNab, line 1528: I'm finding it hard to work out how this is different to the "no compose" path. That should be commented.

parley/src/editing/split_string.rs
```diff
@@ -0,0 +1,161 @@
+// Copyright 2026 the Parley Authors
```
DJMcNab: I don't think this copyright year is right.

parley/src/editing/split_string.rs
```diff
@@ -0,0 +1,161 @@
...
+        return Some(0);
+    }
+    let target = usize::try_from(utf16_offset).ok()?;
+    let mut utf16_count = 0_usize;
+    let mut byte_offset = 0_usize;
+    for segment in text {
+        for ch in segment.chars() {
+            if utf16_count == target {
+                return Some(byte_offset);
+            }
```
DJMcNab, line 132: It isn't clear to me when this condition could ever be true...

parley/src/editing/split_string.rs
```diff
@@ -0,0 +1,161 @@
...
+
+/// Text offset encoding for converting host-provided positions into UTF-8 byte
+/// ranges over visible editor text.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
+pub enum TextIndexEncoding {
+    /// Offsets are already UTF-8 byte indices.
+    Utf8Bytes,
+    /// Offsets are UTF-16 code-unit indices.
+    Utf16CodeUnits,
+    /// Offsets are Unicode scalar-value counts.
```
DJMcNab: So Rust chars, right?

CHANGELOG.md
```diff
@@ -18,6 +18,20 @@ This release has an [MSRV] of 1.88.
 
 - `FallbackKey` now implements `From<Script>` for convenience. ([#594][] by [@xStrom][])
 
+#### Parley
+
+- `PlainEditorDriver` now exposes UTF-8 document-range editing helpers for host-side text input and composition handling, plus delete-to-edge helpers for semantic edit-command integration.
+- `PlainEditorDriver::commit_composition` to explicitly commit an active composition.
+- `SplitString::to_utf8_range` and `TextIndexEncoding` to convert encoded offsets over visible editor text without allocating.
```
DJMcNab: These should have the PR number and author attribution.

DJMcNab (commented):
This seems pretty reasonable to me. I've not dug into the weeds of the semantics of visible text areas, and I feel like we still need to be careful about editing when there is a "hidden preedit". I've not carefully reviewed the example or test changes.

The utf-16 and such helpers are really nice.

I also don't hugely like the term hidden for this. I understand who it's hidden from, but it doesn't feel intuitive for the app to be writing things in terms of "hidden from the app".

parley/src/editing/split_string.rs
waywardmonkeys: ah, because originally this file contained some other code that was all fresh / new and then the actual `SplitString` definition got moved in subsequently.

parley/src/editing/editor.rs
waywardmonkeys: I think a single platform or host could plausibly exercise both modes, because the distinction here is really about how the host describes the current composition state, not a permanent property of a platform. The common preedit-snapshot path maps to transient composition, while protocols that explicitly mark an existing document range as composing map to the visible-region case.

But this isn't yet an area where I understand the downstream needs well enough to predict what people would need (since we're in early days of building out some of this new pathway).

So I did consider exposing that distinction on `Composition`, but I held off because:

1. I don’t have a concrete caller need yet, and
2. I’m not sure “transient vs visible-region” is actually the most useful public question for embedders.

The API I wanted to preserve here was “what composing text is active, and where is it in document space?” If we later find that hosts genuinely need to branch on the composition kind, we can add that.

parley/src/editing/split_string.rs
waywardmonkeys: Yes.

parley/src/editing/editor.rs
waywardmonkeys: Yes, that's the motivation. Some hosts (AppKit via `doCommandBySelector`) surface newline/tab as semantic edit actions rather than text insertion, so the driver needs explicit entry points even though they are simple operations.

This is made a bit more complicated sometimes by the fact that enter/tab can mean inserting a character, field / focus traversal, or submit / search / send.

parley/src/editing/editor.rs
waywardmonkeys: This PR isn't claiming to model that AppKit behavior yet. The current consumers still mostly drive the preedit snapshot path and visible-region support was added so Parley can correctly represent hosts/protocols that explicitly mark an existing document range as composing. (Which right now is probably Android only once Android View is updated?)

parley/src/editing/editor.rs
waywardmonkeys: Fair point. I’ve tightened the docs here so they say explicitly that these delete the current selection if there is one, and otherwise delete to the relevant boundary. I also called out the AppKit semantic command correspondence for the line variants, since that’s the clearest existing source for these operations.

parley/src/editing/editor.rs
waywardmonkeys: Right, this is UTF-8 scalar-value boundaries, not a grapheme-cluster guarantee. I tightened the docs to say that explicitly.

parley/src/editing/editor.rs
waywardmonkeys: Good callout. The current behavior is to replace any existing composition, whether hidden preedit or a visible composing region, and I’ve documented that explicitly now.

parley/src/editing/editor.rs
waywardmonkeys: Yes. I tightened the docs on these methods to say false means the input was invalid and the editor is left unchanged.

parley/src/editing/editor.rs
waywardmonkeys: That’s fair. I expanded the docs here to spell out the replacement precedence: explicit replacement range first, otherwise current composition range if present, otherwise current selection. So this is the "apply the current composition snapshot" entry point, including the common "replace previous preedit with new preedit" case.

parley/src/editing/editor.rs
waywardmonkeys: Yes. I’ve updated the docs to say explicitly that a non-collapsed current selection is included in the deleted range.

parley/src/editing/split_string.rs
waywardmonkeys: I added a comment there. That branch is for the case where the requested UTF-16 boundary lands exactly at the seam between the two visible segments.

parley/src/editing/editor.rs
waywardmonkeys: Agreed. I added a comment there clarifying that a collapsed composition still exists as editor state, but there is no visible range to underline in the layout.

CHANGELOG.md
waywardmonkeys: Fixed.

parley/src/editing/editor.rs
```diff
@@ -343,88 +319,310 @@ where
...
+                .editor
+                .selection
+                .line_start(&self.editor.layout, true)
+                .text_range();
+            self.editor
+                .replace_range_with_selection(self.font_cx, self.layout_cx, range, "");
+        } else {
+            self.delete_selection();
+        }
+    }
```
nicoburns, line 340: I wonder if this function should just be the `if` case, and the user should be expected to do the if-else themselves?

parley/src/editing/editor.rs
```diff
@@ -343,88 +319,310 @@ where
...
+
+    /// Insert a newline at the current selection.
+    ///
+    /// This is useful for hosts that surface newline insertion as a semantic
+    /// editing action rather than ordinary text input.
+    pub fn insert_newline(&mut self) {
+        self.insert_or_replace_selection("\n");
+    }
```
nicoburns, line 393: Does this need to be a standalone function in this API? Couldn't hosts just call `self.insert_or_replace_selection("\n");` themselves? macOS exposes this as a separate action because normal insertions can trigger actions, but that isn't the case with `PlainEditor`'s API.

Same for tab.

parley/src/editing/editor.rs
```diff
@@ -884,14 +1094,15 @@ where
...
-    /// This is useful for suggesting an exclusion area to the platform for, e.g., IME candidate
-    /// box placement. This bounds the area of the preedit text if present, otherwise it bounds the
-    /// selection on the focused line.
-    pub fn ime_cursor_area(&self) -> BoundingBox {
-        let (area, focus) = if let Some(preedit_range) = &self.compose {
+    /// This is useful for suggesting an exclusion area to the platform text
+    /// input system for candidate box placement. This bounds the area of the
+    /// active composition if present, otherwise it bounds the selection on the
+    /// focused line.
+    pub fn text_input_area(&self) -> BoundingBox {
```
nicoburns, line 1101: Q: is this suitable for passing to Winit as the "ime_cursor_area"?

parley/src/editing/editor.rs
```diff
@@ -1130,48 +1342,117 @@ where
...
-            return;
-        }
-        let Some(compose) = &mut self.compose else {
-            return;
-        };
-        if compose.end <= old_range.start {
-            return;
+    fn visible_text_len(&self) -> usize {
+        self.buffer.len() - self.hidden_composition_range().map_or(0, Range::len)
+    }
```
nicoburns, line 1347: Could we document that this returns the length in *bytes*?

parley/src/editing/editor.rs
```diff
@@ -1130,48 +1342,117 @@ where
...
+
+    fn document_byte_to_raw_byte(&self, index: usize) -> Option<usize> {
```
nicoburns, line 1349: All of these methods could do with doc comments explaining what they do.

parley/src/editing/editor.rs
```diff
@@ -1284,3 +1577,31 @@ where
         node.add_action(accesskit::Action::SetTextSelection);
     }
 }
+
+fn validate_text_range(text: &str, range: Range<usize>) -> Option<Range<usize>> {
+    (range.start <= range.end && text.get(range.clone()).is_some()).then_some(range)
+}
+
+fn transformed_range_start(index: usize, replaced: &Range<usize>, inserted_len: usize) -> usize {
```
nicoburns, line 1585: Tranformed from what to what?

parley/src/editing/editor.rs
nicoburns: Perhaps we could have a generic `delete_selection_or(cb: impl FnOnce(&mut PlainEditorDriver)` method?

nicoburns (changes requested):

parley/src/editing/split_string.rs
DJMcNab: What I was trying to express is that:
1) If target is zero, that is caught at the very start of the method.
2) There is no case where code could run between the "upper" and "lower" iterations of this exact same block, so if this would happen, it would be caught by the "lower" check.

PR #604, opened 2026-04-13
Fix word deletion behavior with multiple spaces

Rigidity:
When using the delete word methods, if there are multiple spaces they are treated as a standalone word. Thus, when backdeleting with the cursor at the end of `word `, it deletes the whole thing, but with `word    `, it only deletes the four spaces.

I'm not sure of other implications, but I tested this fix locally and it seems to work fine with both single space gaps and multi space gaps, for forward and backward word deletion. It matches the native behavior of MacOS.

### DIFF
--- a/parley/src/editing/editor.rs
+++ b/parley/src/editing/editor.rs
@@ -267,7 +267,7 @@ where
         if self.editor.selection.is_collapsed() {
             let focus = self.editor.selection.focus();
             let start = focus.index();
-            let end = focus.next_logical_word(&self.editor.layout).index();
+            let end = focus.next_visual_word(&self.editor.layout).index();
             if self.editor.buffer.get(start..end).is_some() {
                 self.editor.buffer.replace_range(start..end, "");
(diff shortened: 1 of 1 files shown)


nicoburns:
Hmm... not sure about changing from logical to visual here. That's going to change the order in which words are deleted for RTL text.

Rigidity:
It looks like next_visual_word and previous_visual_word handle whitespace explicitly. So perhaps this logic can just be ported over to the two logical word methods

nicoburns:
That seems plausible to me, but I'm not confident in that.

xStrom:
I can confirm that the desired behavior on Windows is also to delete the word and the multiple spaces in one go. I think that fixing `next_logical_word` to ignore repeated whitespace is the way to go.

nicoburns:
Closing this, as this approach is not correct, but a PR to fix `next_logical_word` to ignore repeated whitespace would be welcome

PR #609, opened 2026-04-21
Fix going to start of text when moving up on single line

### LINKED ISSUE #608: `Selection::move_lines` does not go to start when there is only one line of editable text
Hello! We are using Parley to implement UI text inputs in Bevy for our 0.19 release very soon, and currently our text inputs are one line long. We currently target Parley v0.8.0 but the problem exists in `main` too.

We expected the 'up' key to move the cursor to the beginning of the text, but instead it doesn't move, even though pressing 'down' moves to the end of the text just fine. I tested this in Parley's `vello_editor` example too and see the same problem. If the editable text range is only one line long, then pressing 'up' will no longer to go the start of text.

When the user presses the 'up' key we call `PlainEditorDriver::move_up()` in parley/source/editing/editor.rs which internally calls `Selection::previous_line()` in selection.rs, then `Selection::move_lines()`. I've narrowed the problem down to `move_lines()`.

The problem is that the first branch of the `if` statement never passes in `move_lines()` to call `line_start()` because `line_limit > 0` is false.

jordanhalase:
We expect `move_lines()` to move the cursor to the start of text if attempting to move up when on the first line. This usually works, however, if there is only one line of editable text this stops working. This fixes the issue.

Closes #608

### DIFF
--- a/parley/src/editing/selection.rs
+++ b/parley/src/editing/selection.rs
@@ -274,7 +274,7 @@ impl Selection {
             .map(|(ix, _)| ix)
             .unwrap_or(line_limit);
         let new_line_index = line_index.saturating_add_signed(delta);
-        if delta < 0 && line_index.checked_add_signed(delta).is_none() && line_limit > 0 {
+        if delta < 0 && line_index.checked_add_signed(delta).is_none() {
             return self
                 .move_to_line(layout, 0, extend)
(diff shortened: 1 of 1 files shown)


nicoburns (approved):

nicoburns:
Thanks!

PR #610, opened 2026-04-21
Fontique supports FreeBSD

fundon:
`fontconfig-rs` supports FreeBSD OS, so it can be added.

### DIFF
--- a/fontique/Cargo.toml
+++ b/fontique/Cargo.toml
@@ -72,7 +72,7 @@ objc2-core-text = { version = "0.3.2", optional = true, default-features = false
     "CTFontCollection",
 ] }
 
-[target.'cfg(target_os = "linux")'.dependencies]
+[target.'cfg(any(target_os = "linux", target_os="freebsd"))'.dependencies]
 yeslogic-fontconfig-sys = { version = "6.0.0", optional = true }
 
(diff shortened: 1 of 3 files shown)


xStrom:
Have you confirmed it working on FreeBSD?

fundon:
Yes, it can work normally. 

Here is an article about font catalogs. https://docs.freebsd.org/en/books/handbook/x11/index.html#x-fonts

xStrom:
What I meant was, do you have a FreeBSD installation that you ran this on? I know it can work in theory, I'm just wondering if it works in practice.

fundon:
I have a FreeBSD-15.0 machine.  Parley  works well.
629c74deab5c92ac5e336d16a5609cc4

xStrom:
Nice, thank you!

I see you marked this PR as draft. Do you plan on adding more changes to this PR before it should be reviewed?

fundon:
I want to add CI for FreeBSD platform and you know there is no good way.

xStrom (approved):
Yes, CI for FreeBSD is tricky. We have never made the effort to get it working in CI for the projects where we've had FreeBSD support.

So continued support will depend on volunteers like you who actually use FreeBSD and report issues.

Thank you for this contribution!

PR #615, opened 2026-04-27
Implement Debug for PlainEditor and its dependent types

### LINKED ISSUE #546: `PlainEditor` does not implement `Debug`
This is generally nice for debugging state, but also prevents any wrapping structs from deriving `Debug`.

NandishwarSingh:
Adds #[derive(Debug)] to PlainEditor along with the three field types that previously lacked it: Generation, LayoutAccessibility, and Layout. The Brush trait already requires Debug, so deriving Debug on Layout<B> and PlainEditor<T> is mechanical; all other field types either already derive Debug or are well-known core types that do.

This allows downstream code to wrap PlainEditor in structs that derive Debug, and makes editor state easier to inspect in logs.

Closes #546.

### DIFF
--- a/parley/src/layout/layout.rs
+++ b/parley/src/layout/layout.rs
@@ -15,7 +15,7 @@ use crate::layout::{
 };
 
 /// Text layout.
-#[derive(Clone)]
+#[derive(Clone, Debug)]
 pub struct Layout<B: Brush> {
     pub(crate) data: LayoutData<B>,
 }
--- a/parley/src/layout/layout.rs
+++ b/parley/src/layout/layout.rs
@@ -7,6 +7,7 @@ use crate::layout::alignment::unjustify;
 use crate::layout::data::LayoutData;
 use crate::style::Brush;
 use core::cmp::Ordering;
+use core::fmt;
 
 use crate::IndentOptions;
 use crate::layout::{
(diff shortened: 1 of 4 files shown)


parley/src/layout/layout.rs
```diff
@@ -15,7 +15,7 @@ use crate::layout::{
 };
 
 /// Text layout.
-#[derive(Clone)]
+#[derive(Clone, Debug)]
```
DJMcNab: I'm worried that these debug results will be extremely long. cc @alice-i-cecile; do you want the full details here (by default), or something more compact.

A possible middle-ground would be for the default debug implementation to show `Layout {spans: 6, text: "the text", ...more_fields}`, then display the full content in the alternative debug implementation.

DJMcNab (commented):
This is reasonable to have here, although I'm worried about the practical usability of this.
Currently, this is going to be almost impossible to actually read.

parley/src/layout/layout.rs
NandishwarSingh: Good point — auto-derive on `Layout<B>` will dump the full `LayoutData` (runs, glyphs, line breaks, styles…) which is rarely what you want at a glance.

I'm happy to switch this to a manual impl that's compact by default and emits full detail under the alternate (`{:#?}`) form. Sketch:

```rust
impl<B: Brush> fmt::Debug for Layout<B> {
    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
        if f.alternate() {
            f.debug_struct("Layout").field("data", &self.data).finish()
        } else {
            f.debug_struct("Layout")
                .field("width", &self.data.width)
                .field("height", &self.data.height)
                .field("lines", &self.data.lines.len())
                .field("runs", &self.data.runs.len())
                .field("styles", &self.data.styles.len())
                .finish_non_exhaustive()
        }
    }
}
```

So `{:?}` prints e.g. `Layout { width: 240.0, height: 48.0, lines: 3, runs: 5, styles: 2, .. }`, while `{:#?}` keeps the full pretty-printed dump for when you actually need it.

Happy to push that as a follow-up commit. Want me to go ahead, or would you rather wait for @alice-i-cecile's input on the desired default first?

DJMcNab: I would be happier with that, yeah. The only other thing which I'm not certain about is that there isn't really a hint about which text this is for. I think the simplest option would be to also include the `text_len` field.

The only other thing is that this detail should be documented in text somewhere; probably in the doc comment on `Layout`, as well as briefly mentioned in the CHANGELOG.

I don't think we need to block on Alice here. The only thing which isn't clear to me is whether these data would be printed automatically by Bevy automatically or not (i.e. whether the alternative flag can be set).

alice-i-cecile: Compact would be ideal please!

We're not planning on printing this regularly: just one-off developer logging.

NandishwarSingh: Pushed the manual Debug impl with text_len + docs + CHANGELOG note. Let me know if anything else needs tweaking.

DJMcNab (approved):
Thank you!

For awareness, I've pushed a commit to streamline some of the docs before landing

NandishwarSingh:
> Thank you!
> 
> For awareness, I've pushed a commit to streamline some of the docs before landing

Thanks for the review and the doc polish, @DJMcNab!

PR #617, opened 2026-04-30
Fix color emoji matching when emoji with non printing variation selector

fundon:
On macOS, after loading the system font, specifying a color emoji font for a range did not take effect.

Some special emoji will be followed by an invisible variation selector [`\u{fe0f}`](https://codepoints.net/U+FE0F).
This causes the font query to not match correctly.

There are some emojis with non printing variation selector:

* ✌️
* ❤️
* ☎️


fundon:
CI failing: It may be related to the generated font.

Is there a command to generate fonts?

parley/src/shape/mod.rs
```diff
@@ -247,7 +247,14 @@ fn fill_cluster_in_place(
         is_emoji_or_pictograph |= info.is_emoji_or_pictograph();
         *code_unit_offset_in_string += ch.len_utf8();
 
-        let contributes_to_shaping = info.contributes_to_shaping();
+        // If the color emoji with an invisible variation selector, ignore it.
+        // e.g. `✌️` (U+270C + U+FE0F)
+        // <https://codepoints.net/U+FE0F>
+        let is_emoji_with_invisible_variation_selector =
+            is_emoji_or_pictograph && info.is_variation_selector();
+
```
conor-93, line 264: I'd be careful with terminology here, the variation selection isn't "invisible", rather it just doesn't have a glyph of its own. It still has a visible impact on presentation (at least, it can, depending on the platform and font), just not on shaping. I'd prefer `is_emoji_with_non_printing_variation_selector`. I think this needs to be reflected in the comment too.

Just U+270C: `✌` (basic or emoji presentation depends on platform/font, I can see in GitHub comments it is an emoji in edit mode, and a basic glyph in view mode)
U+270C + U+FE0E: `✌︎` (force basic presentation)
U+270C + U+FE0F: `✌️` (force emoji presentation

I think it'd be worth linking to https://www.unicode.org/reports/tr37/ as well.

conor-93 (changes requested):
Thanks for the fix! I agree these variation selectors shouldn't affect font queries. I've got one comment about how the change is articulated.

parley/src/shape/mod.rs
fundon: This is very clear. Thank you.

behdad:
By the way, there's also this project that @drott and I put together:

  https://github.com/google/emoji-segmenter

It's used in Chrome. The ragel code can produce Rust output as well, if that's of interest.

fundon:
> By the way, there's also this project that @drott and I put together:
> 
> https://github.com/google/emoji-segmenter
> 
> It's used in Chrome. The ragel code can produce Rust output as well, if that's of interest.

Interesting, thank you for sharing. 

I'll try.

parley/src/shape/mod.rs
```diff
@@ -247,7 +247,20 @@ fn fill_cluster_in_place(
         is_emoji_or_pictograph |= info.is_emoji_or_pictograph();
         *code_unit_offset_in_string += ch.len_utf8();
 
-        let contributes_to_shaping = info.contributes_to_shaping();
+        // If the color emoji with non printing variation selector, ignores the variation selector.
+        //
```
conor-93: nit: grammar
```suggestion
        // If the color emoji has a non-printing variation selector, ignore the variation selector.
        //
```

conor-93 (changes requested):
Thanks! I've added a few suggestions for the documentation, after these are included I'm happy to approve/merge.

parley/src/shape/mod.rs
```diff
@@ -247,7 +247,20 @@ fn fill_cluster_in_place(
         is_emoji_or_pictograph |= info.is_emoji_or_pictograph();
         *code_unit_offset_in_string += ch.len_utf8();
 
-        let contributes_to_shaping = info.contributes_to_shaping();
+        // If the color emoji with non printing variation selector, ignores the variation selector.
```
conor-93: These are lower priority (and skin tone modifiers might be already considered variation selectors by ICU4X (i.e. `is_variation_selector`)), but worth including, as I don't think we've exhaustively considered all codepoint types which shouldn't impact font selection queries for maximum correctness, so I'd extend the comment a bit.

Also, Behdad's segmenter could be used (at least as a reference) when we try an exhaustive solution, so I've included a this in the comment suggestion:

```suggestion
        // TODO: Explore ignoring other modifiers in determining `contributes_to_shaping`:
        //  regional indicators, subdivision flag tag sequences, skin tone modifiers
        //  See also: https://github.com/google/emoji-segmenter
        // If the color emoji with non printing variation selector, ignores the variation selector.
```

conor-93 (changes requested):

conor-93 (approved):

parley/src/shape/mod.rs
fundon: Very good help document.

parley_tests/tests/draw.rs
```diff
@@ -172,3 +172,34 @@ fn draw_colr_emoji() {
...
+    collection.load_system_fonts();
+
+    let text = "\u{270c}\u{fe0f}\u{2705}\u{270c}\u{fe0f}";
+
+    test_with_configs(&mut env, |env| {
+        let mut builder = env.ranged_builder(text);
+        builder.push_default(StyleProperty::FontSize(24.0));
+        builder.push_default(StyleProperty::FontFamily(FontFamily::named(
+            "Apple Color Emoji",
+        )));
```
DJMcNab, line 194: I haven't fully reasoned about the implications of a test which depends on the system font.
Will this test fail if Apple updates the design of their emoji?

fundon: ~~I'm improving this test case, removing system dependencies, and adding a new emoji font for testing.~~

fundon: The core problem is: the first query has matched the correct font, but it will continue to `contributes_to_shaping` and make the second query, and finally return the wrong results.

https://github.com/linebender/parley/blob/42c1ae5ae294f23d96fc0f9891be45d615a5ed0d/parley/src/shape/mod.rs#L581-L586

fundon: If the system font has already been loaded. It will enter the second query if `contributes_to_shaping = true`:

https://github.com/linebender/parley/blob/42c1ae5ae294f23d96fc0f9891be45d615a5ed0d/fontique/src/collection/mod.rs#L350-L362

fundon: I'll try to delete this hardcode.

conor-93: Sorry, I merged this PR too eagerly. I should have asked if you were ready! Feel free to request me as a reviewer on a follow-up.

PR #624, opened 2026-05-22
`attributed_text`:   Generalize attributed text storage and segment ranges

waywardmonkeys:
This updates `attributed_text` so attributed content can be backed by non-contiguous text storage while keeping range validation explicit and reusable.

Goals:

* Support chunk-readable text storage for rope-like or sparse storage backends.
* Store validated `TextRange` values in `AttributedText`.
* Have attribute segment iteration yield validated ranges instead of raw byte ranges.
* Provide a combined segment view for callers that want the range and active spans together.

Non-goals:

* This does not add a concrete rope implementation.
* This does not resolve attribute conflicts or add style cascade behavior.
* This does not change downstream Parley layout behavior.
 
 ## Concepts

* `TextStorage`: the storage abstraction for text bytes and UTF-8 boundary checks.
* `TextChunk`: a borrowed contiguous slice from storage, with its global `TextRange`.
* `TextRange`: a validated byte range whose endpoints are in bounds and on UTF-8 boundaries.
* `AttributeSegments`: the sweep-line segment iterator over applied attribute spans.
* `ActiveSpans`: the active attribute spans for the most recently yielded segment.

## Changes

* Add `TextChunk` and `TextStorage::chunks` for reading validated ranges from contiguous or chunked text storage.
* Add `TextStorage::as_str`, returning `Some(&str)` only for contiguous storage.
* Change `AttributedText::as_str` to return `Option<&str>`.
* Store attribute ranges internally as `TextRange`.
* Change attribute query and segment APIs to expose `TextRange` values.
* Add `AttributeSegments::next_segment` and `AttributeSegment` for callers that want range + active spans as one borrowed view.
* Ensure `ActiveSpansIter` reports an exact `size_hint`, matching its `ExactSizeIterator` implementation.

### DIFF
--- a/attributed_text/src/text_storage.rs
+++ b/attributed_text/src/text_storage.rs
@@ -3,6 +3,50 @@
 
 use alloc::string::String;
 use alloc::sync::Arc;
+use core::ops::Range;
+
+use crate::{Error, TextRange};
+
+/// A borrowed contiguous chunk of text from a [`TextStorage`].
+///
+/// The [`TextRange`] is expressed in the storage's global byte coordinate space, while
+/// [`Self::text`] returns the borrowed UTF-8 text for just this chunk.
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+pub struct TextChunk<'a> {
+    range: TextRange,
+    text: &'a str,
+}
+
+impl<'a> TextChunk<'a> {
+    /// Creates a text chunk for `range`.
+    ///
+    /// # Panics
+    ///
+    /// Panics if the byte length of `text` does not match `range`.
+    #[must_use]
+    pub fn new(range: TextRange, text: &'a str) -> Self {
+        assert_eq!(
+            range.len(),
+            text.len(),
+            "text chunk length must match its range"
+        );
+        Self { range, text }
+    }
+
+    /// Returns this chunk's range in the storage's global byte coordinate space.
+    #[must_use]
+    #[inline]
+    pub const fn range(self) -> TextRange {
+        self.range
+    }
+
+    /// Returns this chunk's borrowed UTF-8 text.
+    #[must_use]
+    #[inline]
+    pub const fn text(self) -> &'a str {
+        self.text
+    }
+}
 
 /// A block of text that will be wrapped by an [`AttributedText`].
 ///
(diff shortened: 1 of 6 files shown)


waywardmonkeys:
This was done with the assistance of Codex (GPT 5.5, xhigh).

waywardmonkeys:
Perhaps also GPT 5.4 as I've had this around for a while.

attributed_text/src/text_storage.rs
```diff
@@ -22,6 +66,28 @@ pub trait TextStorage {
...
+
+    /// Returns a validated [`TextRange`] for this storage.
+    ///
+    /// This is equivalent to calling [`TextRange::new`] with `self`.
+    #[inline]
+    fn validate_range(&self, range: Range<usize>) -> Result<TextRange, Error> {
+        TextRange::new(self, range)
+    }
+
+    /// Iterates over borrowed chunks covering `range`.
```
tomcur, line 85: Perhaps emphasize chunks are contiguous and non-empty.

```suggestion
    /// Iterates over contiguous, non-empty chunks of text covering `range`.
```

attributed_text/src/text_storage.rs
```diff
@@ -52,13 +152,113 @@ impl TextStorage for Arc<str> {
...
+                    let local_end = end - chunk_start;
+                    return Some(TextChunk::new(
+                        TextRange::new_unchecked(start, end),
+                        &chunk[local_start..local_end],
+                    ));
+                }
+            }
+
+            None
+        }
```
tomcur, line 260: The following perhaps reads a bit more clearly.

```suggestion
            let mut chunk_start = 0;
            self.chunks.iter().filter_map(move |&chunk| {
                let offset = chunk_start;
                let start = range.start().max(chunk_start);
                chunk_start += chunk.len();
                let end = range.end().min(chunk_start);
                (start < end).then(|| {
                    let text = &chunk[start - offset..end - offset];
                    TextChunk::new(TextRange::new_unchecked(start, end), text)
                })
            })
        }
    }
```

tomcur (approved):
The new `TextChunk` is nice for non-contiguously stored text! The docs for `TextStorage::chunks` and `TextStorage::as_str` should probably link to each other for discoverability. I've not gone through everything with a fine-toothed comb, but the main changes look good.

`TextStorage` is no longer dyn-compatible, but that probably does not matter. Orthogonal to this PR, it may be worth thinking about what behavior around invalid `TextRange`s should be throughout the crate (panic, clamp, something else?).

PR #625, opened 2026-05-22
ci: Fix some clippy lints, update stable Rust, typos

waywardmonkeys:


parley/src/resolve/mod.rs
```diff
@@ -281,7 +281,7 @@ impl ResolveContext {
         if self.tmp_variations.is_empty() {
             return Resolved::default();
         }
-        self.tmp_variations.sort_by(|a, b| a.tag.cmp(&b.tag));
+        self.tmp_variations.sort_by_key(|a| a.tag);
```
nicoburns, line 284: Unrelated to this PR, but we should probably use unstable sort here (and below).

nicoburns (approved):

PR #628, opened 2026-05-25
Make line commit unconditional and remove try_commit_line macro

nicoburns:
This refactors and simplifies the line-breaking code.

Changes made:

- Eliminate the "fake conditionality" from the `try_commit_line` function (it returned a `bool` as if it was conditional (which historically it used to be), but in reality it was always returned `true`). Rename it just `commit_line` to reflect this reality.
- Take advantage of this non-conditionality to make `start_new_line` call `commit_line` internally and remove the `try_commit_line` macro (the macro was to reduce boilerplate which no longer exists if commit_line is called from `start_new_line`).
- Breaking (low-level API): Make `append_cluster_to_line` increment `self.cluster_idx` (callers should no longer increment this themselves)
- Breaking (low-level API): Make `append_inline_box_to_line` increment `self.item_idx` (callers should no longer increment this themselves)

### DIFF
--- a/parley/src/layout/line_break.rs
+++ b/parley/src/layout/line_break.rs
@@ -1151,53 +1151,6 @@ impl<B: Brush> Drop for BreakLines<'_, B> {
     }
 }
 
-// fn cluster_range_is_valid(
-//     mut cluster_range: Range<usize>,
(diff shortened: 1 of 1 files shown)


taj-p (approved):
Image

PR #630, opened 2026-05-26
attributed_text: add reusable sweep segmentation for unchecked spans

waywardmonkeys:
The main public addition is AttributeSegmentsWorkspace::for_each_span_segment_uncheattributed_text: add reusable sweep segmentation for unchecked spans

This adds a reusable sweep-line path to `attributed_text` so callers can segment a trusted span slice without rebuilding a temporary `AttributedText`.

The main public addition is `AttributeSegmentsWorkspace::for_each_span_segment_unchecked`, which reports each resolved segment plus the active span indices in application order. The name is intentionally explicit: the method does not validate in release builds that the `TextRange` values belong to the text length supplied by the caller.

This also adds `AttributedText::from_attributes_unchecked` for constructing attributed text from an existing attribute vector while preserving application order. Callers should prefer `apply_attribute_bytes` or `TextRange::new` unless the spans have already been checked against the exact text storage.

This is needed by the styled-text builder path, which already has patch ranges in `TextRange` form and wants to resolve overlaps without allocating an intermediate attributed-text value. The API is callback-based so the sweep result can be exposed without committing to extra iterator structs.

### DIFF
--- a/attributed_text/src/attribute_segments.rs
+++ b/attributed_text/src/attribute_segments.rs
@@ -13,17 +13,19 @@ use crate::AttributedText;
 use crate::TextRange;
 use crate::TextStorage;
 
-fn build_segment_state<T: Debug + TextStorage, Attr: Debug>(
-    attributed: &AttributedText<T, Attr>,
+fn build_segment_state_from_ranges<I>(
+    len: usize,
+    attr_count: usize,
+    ranges: I,
     workspace: &mut AttributeSegmentsWorkspace,
-) {
-    let len = attributed.len();
+) where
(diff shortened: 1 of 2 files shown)


tomcur (approved):
This looks good.

My main concern is somewhat orthogonal and take-it-or-leave-it: the API uses the word "validated" for parts where the data is assumed to be correct, whereas often in Rust that relation is flipped, using `try_`, `checked_`, or some other term for APIs that validate (otherwise assuming the data passed in is correct).

waywardmonkeys:
This assumes the caller validated it. More of an unchecked.

PR #631, opened 2026-05-26
styled_text: add compact style spans

waywardmonkeys:
Add `styled_text` as a first-slice crate for compact styled text storage.

The core crate stores resolved complete styles as compact `StyleId` spans over `attributed_text`, with layout and paint payloads interned separately so paint-only changes can share layout identity. It includes a builder that resolves overlapping patch ranges and reusable segment workspaces for downstream layout or paint passes.

This intentionally leaves lowering to Parley, cascading, editing policy, and application-specific style vocabularies outside the core `styled_text` crate.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -11,7 +11,14 @@ env:
   RUST_MIN_VER: "1.88"
   # List of packages that will be checked with the minimum supported Rust version.
   # This should be limited to packages that are intended for publishing.
-  RUST_MIN_VER_PKGS: "-p parley -p parley_core -p parley_data -p parlance -p fontique -p attributed_text"
+  RUST_MIN_VER_PKGS: >-
+    -p parley
+    -p parley_core
+    -p parley_data
+    -p parlance
+    -p fontique
+    -p attributed_text
+    -p styled_text
   # List of features that depend on the standard library and will be excluded from no_std checks.
   FEATURES_DEPENDING_ON_STD: "std,default,png,system,vello_hybrid"
   # List of packages that can not target Wasm.
@@ -105,6 +112,10 @@ jobs:
 
       - name: Run cargo rdme (parlance)
         run: cargo rdme --check --heading-base-level=0 --workspace-project=parlance
+
+      - name: Run cargo rdme (styled_text)
+        run: cargo rdme --check --heading-base-level=0 --workspace-project=styled_text
+
 
   clippy-stable:
     name: cargo clippy
--- a/styled_text/src/lib.rs
+++ b/styled_text/src/lib.rs
@@ -1,0 +1,132 @@
+// Copyright 2026 the Parley Authors
+// SPDX-License-Identifier: Apache-2.0 OR MIT
+
+//! Styled Text stores text with compact full-style identifiers.
+//! It builds on [`attributed_text`] for range storage and segment resolution,
+//! then adds interned style payloads for callers that need compact, reusable
+//! style data.
+//!
+//! The core idea is that each resolved text segment points at a [`StyleId`].
+//! The [`StyleSet`] behind that id stores complete style records, with
+//! layout-affecting payloads and paint-only payloads interned separately.
+//! That means a paint-only change can share layout style identity with the
+//! surrounding text, which is the information a shaping or layout cache usually
+//! wants.
+//!
+//! This is deliberately not a document model.
+//! It does not own shaping, font resolution, inline boxes, cascading, or
+//! renderer-specific style semantics.
+//! Callers choose their own layout and paint payload types, then adapt the
+//! resolved segments to Parley or another layout system.
+//!
+//! Unlike an API that sets individual style bits on ranges and leaves a later
+//! stage to interpret those bits, `styled_text` resolves patches into complete
+//! style records before lowering.
+//! That keeps the core crate independent of any toolkit's style vocabulary
+//! while giving downstream code a simple stream of text ranges and full styles.
+//!
+//! ## Concepts
+//!
+//! - [`StyledText`] stores the text, the resolved [`StyleId`] spans, and the
+//!   shared [`StyleSet`].
+//! - [`StyleSet`] interns layout payloads, paint payloads, and the joined style
+//!   records that point at them.
+//! - [`StylePatch`] is the small trait callers implement to say how a partial
+//!   style change updates their own full style types.
+//! - [`StyledTextBuilder`] appends text and applies patches in the order they
+//!   were applied to the builder.
+//! - [`StyledSegmentsWorkspace`] is reusable scratch storage for iterating
+//!   resolved styled segments without reallocating every time.
+//!
+//! ## Building styled text
+//!
+//! ```no_run
+//! use styled_text::{StylePatch, StyledSegmentsWorkspace, StyledTextBuilder};
+//!
+//! #[derive(Clone, Debug, PartialEq, Default)]
+//! struct LayoutStyle {
+//!     font_size: f32,
+//! }
+//!
+//! #[derive(Clone, Debug, PartialEq, Default)]
+//! struct PaintStyle {
+//!     rgba: [u8; 4],
+//! }
+//!
+//! #[derive(Clone, Debug, Default)]
+//! struct TextStyleChange {
+//!     font_size: Option<f32>,
+//!     rgba: Option<[u8; 4]>,
+//! }
+//!
+//! impl StylePatch<LayoutStyle, PaintStyle> for TextStyleChange {
+//!     fn apply_to(&self, layout: &mut LayoutStyle, paint: &mut PaintStyle) {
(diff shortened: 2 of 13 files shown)


waywardmonkeys:
This was done with the assistance of Codex (GPT 5.3, 5.4, 5.5, medium, high, xhigh).

waywardmonkeys:
Some things I have in mind for follow up PRs:

1. Public builder polish
  - Make the builder API feel like the obvious way to construct styled text.
  - Reduce “find a range, apply patch” ergonomics where possible.
  - Maybe add a push-oriented builder shape for appending text with current style state.
2. Base style / patch convenience
  - Provide a small default patch type or helper for common style updates.
  - Keep the trait-based S`tylePatch` escape hatch, but avoid making every user  define boilerplate for basic cases.
3. Better Parley adapter ergonomics
  - Make `styled_text_parley` easier to call from real code.
  - Keep the reusable workspace story front and center.
  - Avoid multiplying APIs; one good path is better than a pile of helpers.
4. Benchmarks / memory measurements
  - Add focused measurements for style interning, paint-only changes, lowering into Parley, and workspace reuse.
  - Maybe add `malloc_size_of` or equivalent only if we decide the dependency is acceptable.

PoignardAzur (commented):
Quick note: I think this code would be much easier to review if you split it into roughly one file per type. Right now it's kind of hard to find where types are defined without doing a lot of Ctrl+F.

waywardmonkeys:
I have removed `styled_text_parley` from this PR leaving this one with just the base `styled_text` work ... I will be doing the `styled_text_parley` as a separate PR on top of this one.

styled_text/src/lib.rs
```diff
@@ -0,0 +1,132 @@
...
+//! Callers choose their own layout and paint payload types, then adapt the
+//! resolved segments to Parley or another layout system.
+//!
+//! Unlike an API that sets individual style bits on ranges and leaves a later
+//! stage to interpret those bits, `styled_text` resolves patches into complete
+//! style records before lowering.
+//! That keeps the core crate independent of any toolkit's style vocabulary
+//! while giving downstream code a simple stream of text ranges and full styles.
+//!
+//! ## Concepts
```
DJMcNab, line 28: Nit:
```suggestion
//! # Concepts
```
(Plus elsewhere in this file)

This isn't blocking, but it means that if I make the README changes before this lands then we get consistency.

.github/workflows/ci.yml
```diff
@@ -106,6 +113,10 @@ jobs:
       - name: Run cargo rdme (parlance)
         run: cargo rdme --check --heading-base-level=0 --workspace-project=parlance
 
+      - name: Run cargo rdme (styled_text)
+        run: cargo rdme --check --heading-base-level=0 --workspace-project=styled_text
```
DJMcNab, line 117: I know I need to go back to this repo and apply this change more widely, but we may as well start with the new best practise.

This previous mess was my fault, and I now know that the standard approach is to use h1 subheadings in Rust code.

```suggestion
        run: cargo rdme --check --workspace-project=styled_text
```

styled_text/src/style_set.rs
```diff
@@ -0,0 +1,426 @@
...
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        f.debug_tuple("StyleId").field(&self.0).finish()
+    }
+}
+
+/// Interned component identifiers for a full style.
+///
+/// This record is exposed for diagnostics, invalidation decisions, and adapters
+/// that need to know whether two full styles share the same layout or paint
+/// payload. Text spans should store [`StyleId`], not this component record.
```
DJMcNab, line 97: As a conceptual question, are the memory savings from interning a 2-id structure really worth much? Are we expecting to retain the same `StyleId` an average of more than twice for the majority of cases - or is this a play for handling huge documents (order of book length?)?

styled_text/src/style_set.rs
```diff
@@ -0,0 +1,426 @@
...
+    /// Use this for ids that may not belong to this set; see
+    /// [`segment_style`](Self::segment_style) for the panicking convenience.
+    #[must_use]
+    pub fn try_segment_style(&self, id: StyleId) -> Option<SegmentStyle<'_, L, P>> {
+        let record = self.get_style(id)?;
+        Some(SegmentStyle {
+            id,
+            record,
+            layout: self.get_layout(record.layout_id())?,
+            paint: self.get_paint(record.paint_id())?,
```
DJMcNab, line 209: Should these unwrap, or debug assert or something? It seems to me that this can only fail if something has gone wrong in our own bookkeeping.

styled_text/src/style_set.rs
```diff
@@ -0,0 +1,426 @@
...
+    }
+}
+
+/// Borrowed style payloads for a resolved segment.
+#[derive(Debug, PartialEq)]
+pub struct SegmentStyle<'a, L, P> {
+    id: StyleId,
+    record: StyleRecord,
+    layout: &'a L,
+    paint: &'a P,
```
DJMcNab, line 234: Is there a conceptual reason these aren't public fields (potentially `#[non_exhaustive]`)

styled_text/src/style_set.rs
```diff
@@ -0,0 +1,426 @@
...
+    #[inline]
+    pub const fn paint(self) -> &'a P {
+        self.paint
+    }
+}
+
+/// Builder for [`StyleSet`] values.
+///
+/// Interning is currently a linear search over retained vectors. That keeps the
+/// first slice dependency-free and compact; callers can reuse a builder during
```
DJMcNab, line 292: What is the "first slice" here?

styled_text/src/style_set.rs
```diff
@@ -0,0 +1,426 @@
...
+
+impl<L: PartialEq, P: PartialEq> StyleSetBuilder<L, P> {
+    /// Interns layout and paint payloads as a full style and returns its identifier.
+    pub fn intern_style(&mut self, layout: L, paint: P) -> StyleId {
+        let layout = self.intern_layout(layout);
+        let paint = self.intern_paint(paint);
+        self.intern_style_record(StyleRecord { layout, paint })
+    }
+}
+
```
DJMcNab, line 386: So to check my understanding, it's not possible to directly intern a style record which is a combination of two existing layouts and paints from ids?

styled_text/src/text.rs
```diff
@@ -0,0 +1,216 @@
...
+/// Style payloads live in a shared [`StyleSet`], while each applied span stores
+/// only compact identifiers.
+///
+/// The [`StyleSet`] is a frozen palette: it is shared (`Arc`) and only grown
+/// through the builder, so it cannot gain new styles once a `StyledText` holds
+/// it. Mutation here means reassigning ranges to styles that are *already*
+/// interned — for example clearing and reapplying syntax-highlight spans over a
+/// fixed theme with [`clear_styles`](Self::clear_styles) and
+/// [`apply_style`](Self::apply_style). To introduce new styles, rebuild through
+/// [`StyledTextBuilder`](crate::StyledTextBuilder) or a fresh [`StyleSet`].
```
DJMcNab, line 24: First-pass comment: I wonder if we can expose `Arc::make_mut` in a sensible way here.

styled_text/src/text_builder.rs
```diff
@@ -0,0 +1,393 @@
...
+where
+    L: Clone + PartialEq,
+    P: Clone + PartialEq,
+    Patch: StylePatch<L, P>,
+{
+    /// Finishes the builder and returns compact resolved styled text.
+    ///
+    /// Overlapping patches are applied in the order they were applied to the
+    /// builder over a clone of the base layout and paint styles for each
+    /// resolved non-base segment.
```
DJMcNab, line 172: This is (up to) $$\mathcal{O}(n^2)$$ in the number of patches, right?

styled_text/src/text_builder.rs
```diff
@@ -0,0 +1,393 @@
...
+
+        let mut style_builder = StyleSetBuilder::with_capacity(patches.len().saturating_add(1));
+        let base_style = style_builder.intern_style(base_layout.clone(), base_paint.clone());
+
+        let resolved_text = if patches.is_empty() {
+            AttributedText::new(text)
+        } else {
+            let mut resolved_spans = Vec::new();
+            let mut workspace = AttributeSegmentsWorkspace::new();
+            let mut pending: Option<(TextRange, StyleId)> = None;
```
DJMcNab, line 190: I think this needs a comment here. I'm having a hard time grasping what this does.

DJMcNab (approved):
This seems reasonable to me.

Especially as we're not going to depend on this in Parley, it's fairly low-risk for experimentation. And it does unlock some doors for things we will need (HTML richtext copy/paste, etc.).

styled_text/src/style_set.rs
waywardmonkeys: oh, a note to myself that should be removed. First slice of the implementation.

PR #637, opened 2026-06-03
Fix font selection for emoji followed by a variation selector

kane50613:
#617 stopped variation selectors from counting toward shaping, but the coverage check still treated one as a hit when the cluster had a single shaping char. 

So ❤️ (`U+2764 + U+FE0F`) looked fully covered by a font that doesn't actually have `U+2764`, and got picked over the emoji font.

### DIFF
--- a/parley/src/analysis/cluster.rs
+++ b/parley/src/analysis/cluster.rs
@@ -350,10 +350,11 @@ impl<'a> Mapper<'a> {
             return 1.;
         }
         let mut mapped = 0;
+        let has_contributing = self.chars.iter().any(|c| c.contributes_to_shaping);
         for (c, g) in self.chars.iter().zip(glyphs.iter_mut()) {
             if !c.contributes_to_shaping {
                 *g = f(c.ch);
-                if self.map_len == 1 {
+                if !has_contributing {
                     mapped += 1;
                 }
             } else {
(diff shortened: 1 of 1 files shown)


parley/src/analysis/cluster.rs
```diff
@@ -350,10 +350,11 @@ impl<'a> Mapper<'a> {
             return 1.;
         }
         let mut mapped = 0;
+        let has_contributing = self.chars.iter().any(|c| c.contributes_to_shaping);
```
conor-93: I think we should avoid doing a full extra iteration over the cluster's `chars` every time we try to select a font for a cluster. It's probably pretty cheap (`any` will be breaking the iteration for the first char, most of the time), but we have this information available in `fill_cluster_in_place` already, so let's add a new field to `CharCluster` and populate it there.

I have a test that this trips (emoji with VS16 selector) in my own application that I will validate your fix against tomorrow. Thanks for looking into this!

kane50613: I've tried moving `has_contributing` out of the hot loop

conor-93 (approved):
Thank you! I've tested in my application, and this has fixed the bug (an emoji glyph with a VS16 selector is now correctly falling back to a font face that does support it, instead of using a primary font which only has the VS16 codepoint in its cmap table).

kane50613:
@conor-93 sorry for bothering, could you release a 0.10.1 for this fix? otherwise I'll have to continue using v0.9 until v0.11 being released, thanks

DJMcNab:
Apologies, I don't think anyone noticed this request. As you're probably aware, v0.11.0 is now out, so it's a bit moot. For future reference, I'd recommend asking about this in Zulip ([#parley](https://xi.zulipchat.com/#narrow/channel/205635-parley))

PR #639, opened 2026-06-14
Implement `baseline` field on inline boxes

nicoburns:
If the `baseline` field is set, then the box vertically aligned to that baseline rather than to it's bottom edge. This will be particularly useful in conjunction with an implementation of `vertical-align` but is also usable standalone.

Code generated with GPT 5.5 and Opus 4.8 Medium and has been manually reviewed (PR description written by hand).

## Changes made

- Add `baseline: Option<f32>` field to `InlineBox` and `PositionedInlineBox`.
- **Update the line height computation to account for the height taken up by baseline-aligned inline boxes.** This has a tricky corner case in the case that leading is negative due to a small line-height being specified but inline boxes are large enough to expand the line-height: in this case the resultant line-height should be the size of line expanded by the inline boxes, but then shrunk by the amount of negative leading that *would* have been applied if the line's height had not been expanded.
- Add tests for inline boxes with baselines (+ teach the test harness to render inline box baselines)
- Add a couple of convenience helpers (`first_baseline` and `last_baseline`) to `Layout` to allow easy access to the first and last baselines of a `Layout` (largely unrelated to the rest of the PR in terms of implementation, but small an useful in conjunction with the other changes).

## Questions:

- This makes computing the running line height quite a bit more expensive. Do we need to account for this by computing it less often?

### Before:

Screenshot 2026-06-25 at 14 41 01

### With this PR:

Screenshot 2026-06-25 at 14 42 25


waywardmonkeys:
Offhand, this seems correct to me. It doesn't make future possible support for using the ORC any more difficult.

My main feeling about it is that it starts to feel like there should be some helper functions (and that would let us test more directly) for things like creating the positioned inline box since that's effectively just poking at data on the inline box and using a bit of line metrics data.

waywardmonkeys:
(Also, I'm happy to see things like this happening with improved support for baselines. This will be useful when embedding controls and stuff into text in other applications...)

nicoburns:
What is the "ORC"?

waywardmonkeys:
`U+FFFC` ￼ object replacement character

parley/src/layout/layout.rs
```diff
@@ -74,6 +74,24 @@ impl<B: Brush> Layout<B> {
         self.data.full_width
     }
 
+    /// Returns the y-offset of the baseline of the first line in the layout.
+    /// Returns `None` if the layout is empty.
+    pub fn first_baseline(&self) -> Option<f32> {
+        self.lines().next().map(|line| {
+            let metrics = line.metrics();
+            metrics.block_min_coord + metrics.baseline
+        })
```
tomcur: I believe this should be as follows, as `LineMetrics::baseline` is absolute within the layout.

```suggestion
        self.lines().next().map(|line| line.metrics().baseline)
```

Separately, are the first and last lines (and specifically their baselines) general enough to expose as a new API? I suppose they align with `baseline-source`, so they're useful, but the metrics can be accessed by users through `layout.lines().next().map(|l| l.metrics().baseline)`. Removing these removes the mental overhead of understanding what these two methods mean, and wondering whether they're perhaps different from the metrics on the lines.

parley/src/layout/layout.rs
```diff
@@ -74,6 +74,24 @@ impl<B: Brush> Layout<B> {
...
+    }
+
+    /// Returns the y-offset of the baseline of the last line in the layout.
+    /// Returns `None` if the layout is empty.
+    pub fn last_baseline(&self) -> Option<f32> {
+        self.lines().last().map(|line| {
+            let metrics = line.metrics();
+            metrics.block_min_coord + metrics.baseline
+        })
```
tomcur: Similarly (and changing `last` to `next_back`),

```suggestion
        self.lines().next_back().map(|line| line.metrics().baseline)
```

parley/src/layout/line_break.rs
```diff
@@ -929,7 +1062,9 @@ impl<'a, B: Brush> BreakLines<'a, B> {
                     if item.kind == InlineBoxKind::InFlow {
                         // Default vertical alignment is to align the bottom of boxes with the text baseline.
                         // This is equivalent to the entire height of the box being "ascent"
-                        line.metrics.ascent = line.metrics.ascent.max(item.height);
+                        let baseline = item.baseline.unwrap_or(item.height);
+                        line.metrics.ascent = line.metrics.ascent.max(baseline);
+                        line.metrics.descent = line.metrics.descent.max(item.height - baseline);
```
tomcur: I think `finish_line` is not accounting correctly for the "text strut" that the newly-introduced `line_height` does account for.

For example, the following snippet generates a layout of two lines with large line height. The second line has an inline box with ascent `2` and descent `12`. It results in these metrics:

```
line 0 (text):     line_height=40, baseline=25
line 1 (text+box): line_height=40, baseline=61
```

Note the distance between the lines' baselines is not equal to the line height (you'd expect the second line's baseline to land at `65`, which is what happens if you remove the inline box from the snippet).

```rust

let text = "One\nTwo";
let mut builder = env.ranged_builder(text);
builder.push_default(StyleProperty::LineHeight(LineHeight::FontSizeRelative(2.5)));
    id: 0,
    kind: InlineBoxKind::InFlow,
    index: text.find("Two").unwrap() + 1, // between 'T' and 'w'
    width: 15.0,
    height: 14.0,
    baseline: Some(2.0),
});
layout.break_all_lines(None);

let (line0, line1) = (layout.get(0).unwrap(), layout.get(1).unwrap());
let (m0, m1) = (line0.metrics(), line1.metrics());
println!("line 0 (text):     line_height={}, baseline={}", m0.line_height, m0.baseline);
println!("line 1 (text+box): line_height={}, baseline={}", m1.line_height, m1.baseline);
```

Perhaps `finish_line` and `line_height` can share some of their computation?

There's a visual example of the problem in https://github.com/linebender/parley/pull/639#pullrequestreview-4590963930.

parley/src/layout/line_break.rs
```diff
@@ -55,6 +76,83 @@ struct LineState {
...
+    ///   is its own extent shrunk by it.
+    #[inline(always)]
+    fn line_height(&self) -> f32 {
+        let leading = self.text_line_height - self.text_ascent - self.text_descent;
+        let leading_above = leading * 0.5;
+        // Use the exact complement so that `leading_above + leading_below == leading`.
+        let leading_below = leading - leading_above;
+        // A box only grows a side if it extends past the text plus its positive half-leading.
+        let threshold_above = self.text_ascent + leading_above.max(0.0);
+        let threshold_below = self.text_descent + leading_below.max(0.0);
```
tomcur, line 134: I think these `max`es that only let inline boxes interact with positive leading make it such that the line height calculation doesn't match the web. It seems browsers do use the text strut directly (including negative leading), such that an inline box can always grow the line height.

It's possible I'm missing something!

```suggestion
        // A box only grows a side if it extends past the text strut.
        let threshold_above = self.text_ascent + leading_above;
        let threshold_below = self.text_descent + leading_below;
```

See this example: https://developer.mozilla.org/en-US/play?id=WwRq21onl0nFU3sOmurhpP2iTYHKms7tGOiwTN2TrLznBaenaGt8xv5eHT6YmEnSSu5LFab33ool6RiX.

The inline box  there is much taller than the text strut, and the line height of the line it's on is bumped to match.

parley/src/inline_box.rs
```diff
@@ -16,6 +16,8 @@ pub struct InlineBox {
     pub width: f32,
     /// The height of the box in pixels
     pub height: f32,
+    /// The baseline of the box in pixels, relative to the top of the box
+    pub baseline: Option<f32>,
```
tomcur, line 22:
```suggestion
    /// The baseline of the box in pixels, relative to the top of the box.
    ///
    /// If `None`, the baseline is the box's bottom edge.
    pub baseline: Option<f32>,
```

I'd actually be partial to removing the `Option` here, requiring users to define the baseline explicitly. If they want the baseline to be the bottom, they just set it to the same value as `InlineBox::height`. The benefit is that it forces an understanding of what the baseline is relative to, and reduces the mental load of what `None` means.

tomcur (commented):
I like this! I believe some details are not quite right yet; the main issue seems to be handling of leading, in case line height is larger or smaller than `ascent+descent`. Especially `finish_line` doesn't seem to calculate baselines/line heights correctly (and it does not agree with the new `line_height`). The new `line_height` is better, but I believe it should also account for negative leading to align with web behavior. I placed comments on both of these.

The following LLM'd test shows what I think is wrong handling of inline boxes and line heights. The inline box has height 40, descending from the text baseline, with lines having a height of 60. It should fit comfortably within the second line, yet the distance between the first and second lines' baselines becomes 50, and the second and third lines' 85! Paste it e.g. in `parley_tests/src/basic.rs` and run `cargo test -p parley_tests -- inbox_below_baseline_keeps_grid`.

 Remove the asserts to get the snapshot out, which looks as follows.

Image

And without an inline box:

Image

```rust
#[test]
fn inbox_below_baseline_keeps_grid() {

    let text = "AAAA\nBBBB\nCCCC";
    let mut builder = env.ranged_builder(text);
    builder.push_default(StyleProperty::LineHeight(LineHeight::Absolute(60.0)));
        id: 0,
        kind: InlineBoxKind::InFlow,
        index: 7, // between the B's on the middle line
        width: 20.0,
        height: 40.0,
        baseline: Some(0.0),
    });
    layout.break_all_lines(None);
    layout.align(Alignment::Start, AlignmentOptions::default());

    // Every line has the same 60px line height, so consecutive baselines must be exactly 60px
    // apart. The inline box on the middle line fits within the line and must not disturb this.
    let baselines: Vec<f32> = (0..layout.len())
        .map(|i| layout.get(i).unwrap().metrics().baseline)
        .collect();
    assert_eq!(
        baselines[1] - baselines[0],
        60.0,
        "middle line baseline off-grid: {baselines:?}"
    );
    assert_eq!(
        baselines[2] - baselines[1],
        60.0,
        "bottom line baseline off-grid: {baselines:?}"
    );

}
```

parley/src/layout/layout.rs
tomcur: A lighter-weight alternative would be to expose `first_line` / `last_line`, but that also increases API surface somewhat unnecessarily.

parley/src/layout/line_break.rs
nicoburns: This was generated by the AI and also seemed weird to me, and I had it change it. But that led to failing tests, and when I tested it out interactively in the browser I found that this original version actually matched. Specifically I found that the behaviour was:

> If, ignoring inline boxes, the `line-height` / `font-size` would already cause there to be negative leading (roughly: line-height < font-size), then this negative leading was preserved even once inline boxes were added to the line. The inline boxes would still expand the size of the line, but the line would end up the same amount too small as it previously was (as if the negative leading was re-applied after the line was resized to accomodate the box).

I will take a look at your example.

parley/src/inline_box.rs
nicoburns: I would be open to that. I guess I figured `None` reduced the burden on users who don't care about baselines as they can just specify `None`. Whereas if it's plain `f32` then they need to know what behaviour they want.

(also, I think we might want to make the fallback behaviour more sophisticated in future, for example by using the center of the box if the `alignment-baseline` (not yet implemented in Parley) is specified as `center` or `middle` - users could apply that themselves, but it might be nice if they don't have to for things like images that don't have any actual text baselines).

parley/src/layout/line_break.rs
tomcur: This part:

> The inline boxes would still expand the size of the line, but the line would end up the same amount too small as it previously was [..]

matches my understanding, but I believe that also matches the `max`es-removed-code: the line is grown to fit the inline box, and subsequently the ascenders and descenders of the straddling lines overlap the inline box as much as they previously overlapped the inline-box-less text strut.

nicoburns:
> The inline box has height 40, descending from the text baseline, with lines having a height of 60. It should fit comfortably within the second line

Should it? If the baseline is halfway down the line-height (probably actually slightly further down, because normally `ascent` > `descent`) then I'd expect there to be < 40px of available space below the baseline with a line-height of 60px. And I wouldn't expect the inline box to push the line up into the above half-leading (I would expect to generate extra space in the descent / below half-leading...

No?

tomcur:
> Should it? If the baseline is halfway down the line-height (probably actually slightly further down, because normally ascent > descent) then I'd expect there to be < 40px of available space below the baseline with a line-height of 60px.

Ah, that's right (and I chose a bad height for the inline box, not accounting for the fact that leading was shared above and below the glyphs when thinking about the example). The example still shows the strange behavior of the distance between line 1 and line 2's baselines decreasing.

Keeping everything the same, except giving the inline box a height of `15` (having it descend by `15`), produces the following. This time the inline box *should* comfortably fit, but the distance between baselines are `55` for line 1 and line 2, and `65` for line 2 and 3. (The delta in baselines still exists at an inline box height of `6`, at which point the distances are `59` and `61` respectively, but disappears at `5` and below).

inbox_below_baseline_keeps_grid-0

nicoburns:
> Keeping everything the same, except giving the inline box a height of `15` (having it descend by `15`), produces the following. This time the inline box _should_ comfortably fit, but the distance between baselines are `55` for line 1 and line 2, and `65` for line 2 and 3. (The delta in baselines still exists at an inline box height of `6`, at which point the distances are `59` and `61` respectively, but disappears at `5` and below).

Hmm... I wonder if this comes back to the computation in `finish_line` not matching the new algorithm. I think I'll look into that first and then come to this if it's not fixed.

tomcur:
> Hmm... I wonder if this comes back to the computation in finish_line not matching the new algorithm. I think I'll look into that first and then come to this if it's not fixed.

I think that's right. I think the issue in `finish_line` is that it does not account the text strut separately, and so it loses the "inherent" line height information; `line_height` does track the strut separately. (If it's possible to share more of the calculations between the two functions, there'd be less risk of desync.)

tomcur (commented):
Sections 10.8 and 10.8.1 of the CSS 2.1 visual formatting model (https://www.w3.org/TR/CSS2/visudet.html#line-height) seem to agree that negative leading is taken into account for the line-height-with-inline-boxes calculations.

> Still for each glyph, determine the leading L to add, where L = ['line-height'](https://www.w3.org/TR/CSS2/visudet.html#propdef-line-height) - AD. Half the leading is added above A and the other half below D, giving the glyph and its leading a total height above the baseline of A' = A + L/2 and a total depth of D' = D + L/2.
>
> **Note.** L may be negative.
>
> The height of the inline box encloses all glyphs and their half-leading on each side and is thus exactly 'line-height'.

Additionally:

> The line box height is the distance between the uppermost box top and the lowermost box bottom. (This includes the [strut,](https://www.w3.org/TR/CSS2/visudet.html#strut) as explained under ['line-height'](https://www.w3.org/TR/CSS2/visudet.html#propdef-line-height) below.)

And then the following is interesting RE the discussion in Text Office Hours a few hours ago, and may provide a pointer as to the correct implementation.

> The minimum height consists of a minimum height above the baseline and a minimum depth below it, exactly as if each line box starts with a zero-width inline box with the element's font and line height properties. We call that imaginary box a "strut." (The name is inspired by TeX.).

I.e., the code could potentially be organized as to handle the line's text strut exactly as it handles general inline boxes, using font data.

parley/src/layout/layout.rs
nicoburns: I would personally like these methods as they align with concepts in the CSS specs, and save a bit of awkward boilerplate. And they're only small so I'm not too worried about maintenance. OTOH, as they're only small, it's not a huge deal to me if we don't include them.

nicoburns:
@tomcur I think I've resolved the primary issue with the previous state of this PR (that the line_height computation in `finish_line` didn't match the new box-baseline-aware algorithm). And the code itself is partially deduplicated. I've also applied the small fixes to just use `.baseline` and to use `next_back()` rather than `.last()`.

I don't fully have my head around how this ought to work with negative leading, but thus far every time I've attempted to not special case this it seems to be leading to worse output in the snapshots. Perhaps this is now good enough that we can merge out and work out improvements as a follow-up?

parley/src/layout/line_break.rs
```diff
@@ -1075,19 +1223,40 @@ impl<'a, B: Brush> BreakLines<'a, B> {
...
+        let above = if box_ascent > threshold_above {
+            box_ascent
+        } else {
+            ascent + leading_above
         };
 
         let y = self.state.line_y;
-        line.metrics.baseline =
-            ascent + leading_above + if quantize { y.round() as f32 } else { y as f32 };
+        line.metrics.baseline = above + if quantize { y.round() as f32 } else { y as f32 };
```
tomcur, line 1292: I believe `box_ascent` and `box_descent` also need to be `round`ed when quantizing. Currently the `above` here can end up fractional.

I *think* it's right to round those values directly (similar to ascent/descent and the leading pair), as my understanding is that rounding later could get "line rhythm issues".

parley_tests/tests/basic.rs
```diff
@@ -170,13 +178,161 @@ fn inbox_separated_by_whitespace() {
...
+    );
+    assert_eq!(
+        baselines[2] - baselines[1],
+        60.0,
+        "bottom line baseline off-grid: {baselines:?}"
+    );
+
+    env.check_layout_snapshot(&layout);
+}
+
```
tomcur, line 335:
```suggestion
```

This test also is duplicated here and in `line_break.rs`. Perhaps delete this one?

parley/src/layout/line_break.rs
```diff
@@ -494,19 +606,28 @@ impl<'a, B: Brush> BreakLines<'a, B> {
...
+                            // It is then the caller's responsibility to handle placement of the box.
+                            InlineBoxKind::CustomOutOfFlow => {
+                                return Some(YieldData::InlineBoxBreak(BoxBreakData {
+                                    inline_box_id: inline_box.id,
+                                    inline_box_index: item.index,
+                                    advance: self.state.line.x,
+                                }));
+                            }
+                        };
+                    let height_contribution = ascent_contribution + descent_contribution;
```
tomcur: A bit of a nit, but for `InFlow`, `height_contribution` is now reconstructing `inline_box.height` from two parts. It would be slightly more accurate (and clearer?) to encode the tuple as `(width_contribution, height_contribution, ascent_contribution, descent_contribution)`.

Definitely not blocking though.

parley/src/inline_box.rs
tomcur: > I think we might want to make the fallback behaviour more sophisticated in future

That's fair. I think also for e.g. vertical text, we may want to synthesize a different baseline (usually the middle I believe!).

I agree to go with the `Option`, but the docs need to be updated (with something like my suggestion at the start of this comment thread).

tomcur (commented):
This looks good. There are some lingering deviations from the web through how negative leading is handled, but I'm comfortable with us landing this.

I have a few further comments that I wrote inline.

`LineMetrics::{ascent,descent}` have now changed to not take inline boxes into account (which I think is correct), but `LineMetrics::leading` still has the old semantics; feel free to also change that field here, but otherwise I could do a quick follow-up after this lands.

parley_tests/tests/basic.rs
nicoburns: Oops. I've deleted the other one, because I think the other tests in this file are more similar.

parley/src/layout/line_break.rs
nicoburns: I've swapped this around so it's `(width_contribution, height_contribution, resolved_baseline)` and then the ascent/descent are computed below. This way there's no reconstruction: it just defers computation of values that weren't computed yet.

parley/src/layout/line_break.rs
nicoburns: This makes sense to me (although I don't think I fully understand and am more just pattern matching).

I have applied the change in the same place where the regular ascent and descent are rounded.

tomcur (approved):

PR #640, opened 2026-06-16
Line break opportunity override

taj-p:
## Context

Context can be found here: https://canva.link/soclxevd7n6cbcm. TLDR: Chromium and Safari apply custom line breaking opportunity behaviour.

For example, browsers can prevent breaking before/after the "/" in "I would like 1/2 dozen eggs". Breaking on the slash being better or worse is subject to preference. Although ICU dictates that breaking is allowed here, browsers have opted against adopting that behaviour.

## Intent

Allow consumers to override line break opportunity behaviour similar to how Chromium and Safari apply their own preferences when it comes to line breaking behaviour.

AI Disclaimer: Built with Opus 4.8. All documentation hand written. Went through several iterations.

<details>
<summary> AI generated example of matching Chromium's behaviour </summary>

```rs
use std::ops::RangeInclusive;
use parley::LineBreakOverrideFn;

/// Chromium's `ALL_CHAR`: the printable ASCII range `'!'..=0x7F`.
const ALL: RangeInclusive<char> = '!'..='\u{7f}';

/// A line-break opportunity table for printable ASCII pairs, mirroring
/// Chromium's `blink::LineBreakData::FillAscii`.
///
/// Indexed `[before][after]`. Only pairs where *both* characters are printable
/// ASCII (`0x21..=0x7F`) are authoritative; every other pair defers to ICU.
pub struct AsciiLineBreakTable {
    // Boxed to keep the 16 KiB table off the stack. A `[u128; 128]` bitset
    // works too if you want it smaller.
    allow: Box<[[bool; 128]; 128]>,
}

impl AsciiLineBreakTable {
    fn blank() -> Self {
        Self { allow: Box::new([[false; 128]; 128]) }
    }

    /// Equivalent of Chromium's
    /// `SetPairValue(before_lo, before_hi, after_lo, after_hi, value)`.
    fn set(&mut self, before: RangeInclusive<char>, after: RangeInclusive<char>, allow: bool) {
        for b in (*before.start() as u32)..=(*before.end() as u32) {
            for a in (*after.start() as u32)..=(*after.end() as u32) {
                self.allow[b as usize][a as usize] = allow;
            }
        }
    }

    /// `Some(true)`  -> force a break opportunity
    /// `Some(false)` -> suppress any break opportunity
    /// `None`        -> defer to ICU/UAX#14 (anything not printable ASCII)
    fn lookup(&self, before: char, after: char) -> Option<bool> {
        let (b, a) = (before as u32, after as u32);
        if (0x21..=0x7F).contains(&b) && (0x21..=0x7F).contains(&a) {
            Some(self.allow[b as usize][a as usize])
        } else {
            None
        }
    }

    /// Direct port of `FillAscii()`. Order matters: later `set` calls overwrite
    /// earlier ones, exactly as Chromium's `SetPairValue` calls do.
    pub fn chromium() -> Self {
        let mut t = Self::blank();

        t.set(ALL, ALL, false);
        t.set(ALL, '('..='(', true);
        t.set(ALL, '<'..='<', true);
        t.set(ALL, '['..='[', true);
        t.set(ALL, '{'..='{', true);
        t.set('-'..='-', ALL, true);
        t.set('?'..='?', ALL, true);
        t.set('-'..='-', '$'..='$', false);
        t.set(ALL, '!'..='!', false);
        t.set('?'..='?', '"'..='"', false);
        t.set('?'..='?', '\''..='\'', false);
        t.set(ALL, ')'..=')', false);
        t.set(ALL, ','..=',', false);
        t.set(ALL, '.'..='.', false);
        t.set(ALL, '/'..='/', false);
        // Note: Chromium hard-codes '-' followed by [0-9] in `ShouldBreakFast()`.
        t.set('-'..='-', '0'..='9', false);
        t.set(ALL, ':'..=':', false);
        t.set(ALL, ';'..=';', false);
        t.set(ALL, '?'..='?', false);
        t.set(ALL, ']'..=']', false);
        t.set(ALL, '}'..='}', false);
        t.set('$'..='$', ALL, false);
        t.set('\''..='\'', ALL, false);
        t.set('('..='(', ALL, false);
        t.set('/'..='/', ALL, false);
        t.set('0'..='9', ALL, false);
        t.set('<'..='<', ALL, false);
        t.set('@'..='@', ALL, false);
        t.set('A'..='Z', ALL, false);
        t.set('^'..='`', ALL, false);
        t.set('a'..='z', ALL, false);
        t.set('{'..='{', ALL, false);
        t.set('\u{7f}'..='\u{7f}', ALL, false);

        t
    }

    /// Convert into a closure installable on a Parley builder.
    pub fn into_override(self) -> Box<LineBreakOverrideFn> {
        Box::new(move |before, after| self.lookup(before, after))
    }
}
```

</details>


parley/src/analysis/mod.rs
```diff
@@ -227,7 +228,14 @@ pub(crate) enum Boundary {
...
 }
 
-pub(crate) fn analyze_text<B: Brush>(lcx: &mut LayoutContext<B>, mut text: &str) {
+pub(crate) fn analyze_text<B: Brush>(
+    lcx: &mut LayoutContext<B>,
+    mut text: &str,
+    line_break_override: Option<&LineBreakOverrideFn>,
+) {
+    #[cfg(not(feature = "line-break-overrides"))]
+    let _ = line_break_override;
```
DJMcNab: Just to check my understanding, this will always be `None` here, right? I don't *think* it's worth debug asserting that, but raising it as something which jumped out at me.

parley/src/analysis/mod.rs
```diff
@@ -424,20 +434,40 @@ pub(crate) fn analyze_text<B: Brush>(lcx: &mut LayoutContext<B>, mut text: &str)
...
+                is_line = true;
                 _ = lb_iter.next();
             }
         }
 
+        #[cfg(feature = "line-break-overrides")]
+        // This leaves word boundaries intact. Consumers can only impact line boundaries.
+        {
+            if let (Some(p), Some(o)) = (prev_char, lb_override) {
+                if let Some(forced) = o(p, ch) {
```
DJMcNab: Nit: Can we use e.g. `override` instead of `o` here?

parley/src/builder.rs
```diff
@@ -9,20 +9,39 @@ use super::style::{Brush, StyleProperty, TextStyle, WhiteSpaceCollapse};
...
+/// Returning:
+///    - `Some(true)`  : forces a line break opportunity between the pair
+///    - `Some(false)` : suppresses any opportunity
+///    - `None`        : defers to the default (ICU) behavior
+///
+/// Mandatory breaks are unaffected (e.g. `\n`).
+///
+/// This is typically used to force preferential line breaking decisions when it
+/// comes to ASCII punctuation like `/`, '-', etc. For example, to prevent break
+/// opportunities within "1/2".
```
DJMcNab: I'd probably also like this doc to explicitly mention matching the behaviour of existing systems such as browsers.

parley/src/builder.rs
```diff
@@ -50,6 +69,11 @@ impl<B: Brush> RangedBuilder<'_, B> {
         self.lcx.inline_boxes.push(inline_box);
     }
 
+    #[cfg(feature = "line-break-overrides")]
+    pub fn set_line_break_override(&mut self, overrides: Option<Box<LineBreakOverrideFn>>) {
```
DJMcNab: This needs docs; we really should go through and localise `expect(missing_docs)` on every item, so new items are properly forced to have them.

They only need to be simple, e.g.:
```
/// Set the callback which will be called as a first provider of line breaking decisions.
///
/// See [`LineBreakOverrideFn`] for more details.
```

parley/src/builder.rs
```diff
@@ -142,6 +168,11 @@ impl<B: Brush> StyleRunBuilder<'_, B> {
...
```
DJMcNab: Same here.

parley/src/builder.rs
```diff
@@ -224,6 +257,11 @@ impl<B: Brush> TreeBuilder<'_, B> {
             .set_white_space_mode(white_space_collapse);
     }
 
+    #[cfg(feature = "line-break-overrides")]
+    pub fn set_line_break_override(&mut self, overrides: Option<Box<LineBreakOverrideFn>>) {
```
DJMcNab: And here, of course :)

DJMcNab (approved):
I'm happy with this implementation of this feature. It is obviously unfortunate that we have extra complexity here, but there at least shouldn't be any costs to users who don't use it.

The one blocking concern from my side is the missing docs.

We should in theory also document the new feature in lib.rs (and therefore also README.md), however I'm happy to skip that because the already existing features aren't properly documented in that way.

I'm approving, but I'd like to get sign-off from Nico or Tom, in case I'm missing something else (e.g. it *could* be worth having a couple of extra tests, such as the interaction with `\n`?).

parley/src/analysis/mod.rs
taj-p: Yup! It should be `None` all the time because when the feature is disabled there is no setter

parley/src/builder.rs
taj-p: Good idea! Fixed in https://github.com/linebender/parley/pull/640/commits/a546ac4cf80ebfcdaef138061e728f8b1e5e7858

parley/src/analysis/mod.rs
taj-p: Yes - definitely. Fixed in https://github.com/linebender/parley/pull/640/commits/a546ac4cf80ebfcdaef138061e728f8b1e5e7858

parley/src/builder.rs
taj-p: Agree! Fixed in https://github.com/linebender/parley/pull/640/commits/a546ac4cf80ebfcdaef138061e728f8b1e5e7858

parley/src/builder.rs
taj-p: Fixed in https://github.com/linebender/parley/pull/640/commits/a546ac4cf80ebfcdaef138061e728f8b1e5e7858

parley/src/builder.rs
taj-p: Fixed in https://github.com/linebender/parley/pull/640/commits/a546ac4cf80ebfcdaef138061e728f8b1e5e7858

taj-p:
> it could be worth having a couple of extra tests, such as the interaction with \n

Added these tests in https://github.com/linebender/parley/pull/640/commits/e5c7e19b9f44d53484d1bc71e72608fb0e67b11f

nicoburns (commented):
This generally makes sense to me.

Some questions:

1. Why is this feature flagged? Seems like it could just be an always-on feature that is only activated when a `LineBreakOverrideFn` is passed in?
2. Have you benchmarked performance? (perhaps that's why it's feature flagged?)
3. Could we include the line-breaking function for Chrome in the library (or at least in an example?). I expect quite a lot of people would find that useful

taj-p:
>  Why is this feature flagged? Seems like it could just be an always-on feature that is only activated when a `LineBreakOverrideFn` is passed in?

Good call!! Thank you. Removed in https://github.com/linebender/parley/pull/640/changes/8749d01b7546ce0dc344123148b4ac9e1896715c.

---

>  Have you benchmarked performance? (perhaps that's why it's feature flagged?)

I mostly just saw noise:

image


---


>  Could we include the line-breaking function for Chrome in the library (or at least in an example?). I expect quite a lot of people would find that useful

Yes, definitely. At first I wasn't sure others would want this - @DJMcNab actually recommended this too. I've added it in https://github.com/linebender/parley/pull/640/commits/8749d01b7546ce0dc344123148b4ac9e1896715c as part of the library.

Daniel is looking at adding tests to Parley to ensure its behaviour is consistent with Chrome (which I think is useful for trying to emulate it, but also as a means to assert correctness against a mature reference).

parley/src/break_overrides.rs
```diff
@@ -0,0 +1,236 @@
...
+///
+/// #### Compared to Safari
+///
+/// Chromium (and this table) differs from Safari in 9 cases:
+///  - A "-" followed by one of "!|$|)|/|:|;|?|]|}" is suppressed in Chromium, but broken in Safari.
+///
+/// #### Compared to Firefox
+///
+/// Firefox always defers to the default ICU behavior.
+pub static CHROMIUM_LINE_BREAK_TABLE: AsciiLineBreakTable = AsciiLineBreakTable::chromium();
```
DJMcNab: This is about 4KiB of data, right? I wonder how compressible it is? I don't think this blocks; consumers who really care about binary size can use their own compression form; certainly for non-web, this is negligible.

parley/src/break_overrides.rs
```diff
@@ -0,0 +1,236 @@
...
+/// See: <https://source.chromium.org/chromium/chromium/src/+/main:third_party/blink/renderer/platform/text/character_property_data_generator.cc;l=449-495>
+///
+/// ### Differences from other browsers
+///
+/// #### Compared to Safari
+///
+/// Chromium (and this table) differs from Safari in 9 cases:
+///  - A "-" followed by one of "!|$|)|/|:|;|?|]|}" is suppressed in Chromium, but broken in Safari.
+///
+/// #### Compared to Firefox
```
DJMcNab: Nit: The standard style is to use a single `#` for a heading in a comment. You've probably picked up the "using multiple" `#`s from me - sorry!

```suggestion
/// # Differences from other browsers
///
/// ## Compared to Safari
///
/// Chromium (and this table) differs from Safari in 9 cases:
///  - A "-" followed by one of "!|$|)|/|:|;|?|]|}" is suppressed in Chromium, but broken in Safari.
///
/// ## Compared to Firefox
```

parley/src/break_overrides.rs
```diff
@@ -0,0 +1,236 @@
...
+///
+/// Firefox always defers to the default ICU behavior.
+pub static CHROMIUM_LINE_BREAK_TABLE: AsciiLineBreakTable = AsciiLineBreakTable::chromium();
+
+/// A line break override table for ASCII character pairs.
+///
+/// See [`LineBreakOverrideFn`] for more details.
+///
+/// The whole table is built at compile time so [`CHROMIUM_LINE_BREAK_TABLE`],
+/// for example, has no runtime cost.
```
DJMcNab: Nit: Maybe something like:
```suggestion
/// All table operations are `const`, which means that [`CHROMIUM_LINE_BREAK_TABLE`]
/// needs no explicit construction step.
```

Specifically, this doc suggests a truism about all such tables, when in fact it depends on the use case.

parley/src/break_overrides.rs
```diff
@@ -0,0 +1,236 @@
...
+#[derive(Clone)]
+pub struct AsciiLineBreakTable {
+    /// Bit `after` set in row `before` means the pair has an explicit override.
+    overridden: [u128; 128],
+    /// Bit `after` set in row `before` means a break is allowed for that pair.
+    allow: [u128; 128],
+}
+
+impl AsciiLineBreakTable {
+    /// A table that defers every pair to the default ICU behavior.
```
DJMcNab, line 111: This comment doesn't quite have the right focus, but the scenario where that causes issues is so remote that I don't think it's worth spending effort improving it - I can't write a better short comment for here quickly.

parley/src/break_overrides.rs
```diff
@@ -0,0 +1,236 @@
...
+    }
+
+    /// Override every pair in the cartesian product of the two inclusive ASCII
+    /// ranges.
+    pub const fn with_pairs(
+        mut self,
+        before_lo: char,
+        before_hi: char,
+        after_lo: char,
+        after_hi: char,
```
DJMcNab: Thoughts on `RangeInclusive<char>` here? I'd also argue that `u8` would be a defensible choice here, but either one is fine. This doesn't block, although obviously it has a future issue.

parley/src/break_overrides.rs
```diff
@@ -0,0 +1,236 @@
...
+        allow_break: bool,
+    ) -> Self {
+        let mut before = before_lo as u32;
+        while before <= before_hi as u32 {
+            let mut after = after_lo as u32;
+            while after <= after_hi as u32 {
+                if before < 128 && after < 128 {
```
DJMcNab: I'd probably expect the else branch here to panic. Is there a particular reason it doesn't? Obviously we'd need to specify that all the inputs must be within ASCII ranges if we did that, but we probably should anyway.

DJMcNab (approved):
With Nico's in-principle sign-off, I'm happy to land this. I do have a few nits here; the most blocking is probably the API for `with_pairs`; everything else is just doc nits, and unimportant nits at that. I'd be fine for this to land as-is, though.

parley/src/break_overrides.rs
taj-p: Brilliant catch. Thank you! Fixed in https://github.com/linebender/parley/pull/640/commits/7ef61237784685b5955daaa9e098583fbe465331

parley/src/break_overrides.rs
taj-p: Much better. Thank you! Fixed in https://github.com/linebender/parley/pull/640/commits/7ef61237784685b5955daaa9e098583fbe465331

parley/src/break_overrides.rs
```diff
@@ -4,6 +4,7 @@
 //! Overrides for line-break opportunities.
 
 use alloc::boxed::Box;
+use core::ops::RangeInclusive;
```
DJMcNab, line 6: We should have an MSRV reminder in Cargo.toml about using `core::range::Range*` instead of `core::ops::Range*`. However, I'm more than happy to land this PR without that.

DJMcNab (approved):
Thank you! Presuming that we've validated that this works as expected internally, I think this is good to land.

parley/src/break_overrides.rs
taj-p: Great question. Yes - it's more than I expected. I simply copied the approach in Chromium - but we can probably do better. I've opened https://github.com/linebender/parley/pull/641 for discussion

parley/src/break_overrides.rs
taj-p: Added in https://github.com/linebender/parley/pull/640/commits/fd446b5ae30770610bafc22d62121c2cb0781136

parley/src/builder.rs
```diff
@@ -50,6 +53,13 @@ impl<B: Brush> RangedBuilder<'_, B> {
         self.lcx.inline_boxes.push(inline_box);
     }
 
+    /// Set the callback which will be called as a first provider of line breaking decisions.
+    ///
+    /// See [`LineBreakOverrideFn`] for more details.
+    pub fn set_line_break_override(&mut self, overrides: Option<Box<LineBreakOverrideFn>>) {
```
taj-p: I realised consumers may not want to allocate for every call to builder 😅 . Going to make this borrow based prior to merge such that an API like the below works:

```rs
let lookup = |before, after| (before == '/' || after == '/').then_some(false);
let text = "1/2";
builder.set_line_break_override(Some(&lookup));
```

Where the builder looks like:


```rs
    pub fn set_line_break_override(&mut self, overrides: Option<&'b LineBreakOverrideFn>) {
        self.line_break_override = overrides;
    }
```

taj-p:
> Presuming that we've validated that this works as expected internally

Yup - confirmed this works as expected insofar that the case that erred before is now resolved.

parley/src/builder.rs
taj-p: Added in https://github.com/linebender/parley/pull/640/commits/a50f41a6b4fa1bbdf007b49192c1caa03a2316ce

taj-p:
cc @DJMcNab and @nicoburns 

Prior to merge, I ran Parley through our 100k corpus of tests and noticed that I forgot the special casing Chrome does for "-" in https://github.com/chromium/chromium/blob/c6bee15e8f336c8feabf539d8bbb540c134ec20a/third_party/blink/renderer/platform/text/text_break_iterator.cc#L224-L240.

Essentially, added these lines:

```rs
    // Before consulting 'before' / 'after' pair table, check for the special "-" case.
    //
    // Chromium doesn't allow breaking when it looks like the minus sign is for a negative
    // number like "Subtract -5 from X". But, Chromium does allow breaking when the minus
    // sign is part of a long URL like "AAAA-2222".
    //
    // See <https://github.com/chromium/chromium/blob/c6bee15e8f336c8feabf539d8bbb540c134ec20a/third_party/blink/renderer/platform/text/text_break_iterator.cc#L224-L240>
    if before == '-' && after.is_ascii_digit() {
        return Some(before_before.is_ascii_alphanumeric());
    }
```

This has been addressed in https://github.com/linebender/parley/pull/640/changes/a50f41a6b4fa1bbdf007b49192c1caa03a2316ce..b7f4cd0196133b11fd51bf1e45468ac77e3acfea.

parley/src/break_overrides.rs
```diff
@@ -0,0 +1,269 @@
...
+/// Mandatory breaks are unaffected (e.g. `\n`).
+///
+/// This is typically used to force preferential line breaking decisions when it
+/// comes to ASCII punctuation like `/`, `-`, etc. For example, to prevent break
+/// opportunities within "1/2".
+///
+/// A separate use case is to match the line breaking behavior of existing systems
+/// such as web browsers. See [`CHROMIUM_LINE_BREAK_OVERRIDE`] for a ready-made
+/// override function that mirrors Chromium's behavior.
+pub type LineBreakOverrideFn = dyn Fn(char, char, char) -> Option<bool> + Send + Sync;
```
DJMcNab: I'm not a huge fan of the `before_before` argument... I'd almost be tempted to say we could do something like: `&str, usize`, but I also don't really like that.

A more scalable approach would be to make this `FnMut`, if the lifetimes work out. We'd obviously have to guarantee that it was single pass in-order, but I don't know how likely we are to get in-layout parallelism.

Luckily, this is an extremely niche api (people are unlikely to write an implementation of this themselves). If my other suggestion can't work, then I think landing this as is would be *fine*, but it doesn't spark joy.

parley/src/analysis/mod.rs
```diff
@@ -424,20 +431,39 @@ pub(crate) fn analyze_text<B: Brush>(lcx: &mut LayoutContext<B>, mut text: &str)
...
             if *l == byte_pos {
-                boundary = Boundary::Line;
+                is_line = true;
                 _ = lb_iter.next();
             }
         }
 
+        // This leaves word boundaries intact. Consumers can only impact line boundaries.
+        if let (Some(prev), Some(lb_override)) = (prev_char, line_break_override) {
+            let before_before = prev_prev_char.unwrap_or('\0');
```
DJMcNab: Do we need to unwrap here? What does that win over just passing an Option argument?

This is minor polish thing - if this doesn't change, this is still fine to land.

DJMcNab (commented):
A few extra thoughts here. Given the nicheness of this API, I don't particularly care about getting this perfect, and so none of these block.

nicoburns (approved):

parley/src/analysis/mod.rs
taj-p: Moved to option in https://github.com/linebender/parley/pull/640/commits/f372423cf7dd12dbb0f076901f27677b2f3b5196

parley/src/break_overrides.rs
taj-p: I tried the `FnMut` approach but didn't like it. Elected to use `non_exhaustive` here in case we do want to end up passing the `&str` with a `usize` in addition to the `char`s in https://github.com/linebender/parley/pull/640/commits/f372423cf7dd12dbb0f076901f27677b2f3b5196

PR #641, opened 2026-06-17
Compressing Line Break Opportunity Tables

taj-p:
Stacked on top of https://github.com/linebender/parley/pull/640.

In https://github.com/linebender/parley/commit/e453b83a97e76d0d895392cd7c62434f2e5ee2ef, you can see how we can condense the Chromium table from 4KiB to 290 B while keeping lookup O(1).


nicoburns:
If the final table is only 290 bytes then could we build the table offline and just include the precompiled table in the final binary?

EDIT: ah but need the general machinery anyway for user-specified tables?

taj-p:
> If the final table is only 290 bytes then could we build the table offline and just include the precompiled table in the final binary?
> 
> EDIT: ah but need the general machinery anyway for user-specified tables?

We could do that! The table build process is all `const` time compatible so only ~290B lands in the binary via the current process so we won't benefit too much from offline preparation if we're worried about binary size. 

> EDIT: ah but need the general machinery anyway for user-specified tables?

I guess so!

nicoburns (approved):

PR #642, opened 2026-06-19
Testing: Edge-case linebreaking matches Chromium

DJMcNab:
This new test uses Chromium as a source for line breaking data, and checks that we get the same result in the edge cases.
It does not bring Chrome into our CI; these are instead collected manually and committed in the repo, to ensure stability.

This has requires a few tweaks to how we call Parley to more closely match Chrome, including working around a cache collision bug in Chrome (which will only be reasonably found in a harness like this, so I haven't reported it).

On my machine (which is admittedly quite fast), these tests pass in 3s. As such, I don't expect their performance to be a significant limiting factor here. A release mode run would make them ~instant (0.10s).

AI usage: Code developed with Claude Opus 4.8, and then significantly tweaked to make sane and more understandable. I also used Sonnet to review, which caught a couple of issues.


parley_tests/linebreaking_browser_recorder/data/Arimo.csv
conor-93, line 1: I think we need to copy the licenses in `parley_dev/assets/fonts/arimo_fonts` and `parley_dev/assets/fonts/roboto_fonts` into this folder if we're going to use these fonts in this way.

parley_tests/linebreaking_browser_recorder/Cargo.toml
conor-93, line 1: For crate names, could we go with these names, for more consistency `linebreaking_*`, and clarity (on the first name)?
- `linebreaking_browser_recorder`
- `linebreaking_tests`

Open to suggestions.

parley_tests/linebreaking_cases/src/lib.rs
```diff
@@ -0,0 +1,164 @@
...
+        // However, in real world conditions, we do want to make sure that we handle font sizes
+        // which don't neatly fall on that grid, so add an offset which still maintains the non-collision.
+        let offset = rng.random_range(0_f32..0.0095_f32);
+        let font_size = font_size + offset;
+        let em_factor = rng.random_range(MIN_EM_FACTOR..=MAX_EM_FACTOR);
+        let initial_width = font_size * em_factor;
+        let word_count = rng.random_range(MIN_WORDS..=MAX_WORDS);
+
+        let mut text = String::new();
+        // TODO: Ideally, this generation would be more free-form, to e.g. test line breaking with punctuation.
```
conor-93, line 125: non-blocking: Do we want to expand beyond just `VALID_LETTERS` to include numerals and punctuation in this PR? If nothing else, glyphs with narrower advance width (punctuation) might be more likely to trip boundary cases.

parley_tests/linebreaking_browser/index.html
```diff
@@ -0,0 +1,56 @@
...
+    }
+
+    th {
+      background: #f3f3f3;
+    }
+  </style>
+</head>
+
+<body>
+  <h1>Parley ↔ Chrome line-breaking data collection</h1>
```
conor-93: ditto

parley_tests/linebreaking_browser/src/main.rs
```diff
@@ -0,0 +1,340 @@
...
+//! given a random width. We then find and record the minimum width which also causes the same break.
+//! A future extension will be to evaluate with a more reasonable bound across multiple lines of text;
+//! this test is designed to catch the edge cases for that test.
+//!
+//! Build/run with [Trunk](https://trunkrs.dev):
+//!
+//! ```sh
+//! rustup target add wasm32-unknown-unknown
+//! cargo install --locked trunk
+//! trunk serve --open   # run from this crate's directory
```
conor-93: I haven't confirmed this, but Claude suggests `--open` just uses your default browser, not necessarily Chrome? I don't think there's a way around this, but it's worth documenting. Maybe erroring out after `let version = chrome_version(&window);` accordingly, but I'm not sure. WDYT?

parley_tests/linebreaking_browser_recorder/data/Arimo.csv
DJMcNab: I'm happy enough to oblige, but that's a surprising interpretation to me. Would you be able to expand on why that's needed?

parley_tests/linebreaking_browser/src/main.rs
DJMcNab: Two things:
1) I didn't realise we had `--open` there, although I'm not sure I would have changed it if I had seen it - thanks for the callout. That's my bad for not cleaning up that part of Claude's work.
2) I think erroring if we don't have a Chrome version would be reasonable, although it's worth noting that the way the metadata is provided is my protection against that. That is, a use of this with a non-chromium browser should be caught in review.

parley_tests/linebreaking_cases/src/lib.rs
DJMcNab: Punctuation is definitely out for now, but the thinking is that we'd expand to that soon after #640 lands. I'll try again with adding numbers though.

parley_tests/linebreaking_browser_recorder/data/Arimo.csv
DJMcNab: In particular, I'd like to understand why this same concern didn't apply to the snapshot tests which used these fonts.

conor-93: No, I think you're right, because there's no way to recreate the fonts from that CSV data (maybe advance width of glyphs at a stretch). I was initially concerned because the licensed font file is referenced (by name) but I don't think this actually matters.

parley_tests/linebreaking_browser_recorder/Cargo.toml
DJMcNab: I've renamed the recorder and the test. I think that the case generator is at a local maximum; concretely, the name `_tests` is worse because it doesn't actually contain any tests...

parley_tests/linebreaking_cases/src/lib.rs
DJMcNab: I've added numbers as well. No new failures.

nicoburns:
Interesting, am I understanding correctly that the recorder is intended to be served with Trunk, and then opened manually in Chrome? I wonder if there would be any value in aligning with Taffy's test generator (https://github.com/DioxusLabs/taffy/tree/main/scripts/gentest) which runs Chrome using WebDriver? Perhaps not worth the lift in the short term, but worth considering in future?

(Taffy's runner currently uses the `fantoccini` crate which has the advantage of making it compatible with other browsers (although we currently only run it against chrome) but the disadvantage of requiring one to install Chrome and a matching version of `chromedriver` manually. But if one was willing to fix on Chrome then there are other Rust crates that will download a test version of Chrome and run it automatically without any setup from the user at all)

DJMcNab:
Yeah, that's the current state. There would be *some* value in having it be automated, but my expectation is that these values change so rarely anyway that it isn't especially worthwhile.

Edit: I've done a little bit of looking into gentest, and it looks to be a lot more general here?

conor-93 (approved):

PR #645, opened 2026-06-21
Point `tango` to git rev to fix paired benchmarks

tomcur:
`tango` recently switched to running the paired comparison benchmarks in separate processes (https://github.com/bazhenov/tango/pull/78). This is fortunate, as it turns out to also fix an issue I encountered in https://github.com/linebender/parley/pull/634#discussion_r3449043356.

On current `main`, if I run `cargo export target/benchmarks -- bench --bench=main` followed by `cargo bench -q --bench=main -- compare target/benchmarks/main -p -t 4.`, I get the following!

```
Default Style - arabic 20 characters               [   8.9 us ...   8.9 us ]      +0.35%
Default Style - latin 20 characters                [   4.3 us ...   4.1 us ]      -3.50%*
Default Style - japanese 20 characters             [   8.2 us ...   8.1 us ]      -0.44%
Default Style - arabic 1 paragraph                 [  48.5 us ...  48.1 us ]      -0.66%
Default Style - latin 1 paragraph                  [  17.2 us ...  16.1 us ]      -6.52%*
Default Style - japanese 1 paragraph               [  71.3 us ...  69.6 us ]      -2.50%*
Default Style - arabic 4 paragraph                 [ 214.5 us ... 212.7 us ]      -0.86%
Default Style - latin 4 paragraph                  [  66.0 us ...  62.0 us ]      -5.96%*
Default Style - japanese 4 paragraph               [ 100.7 us ...  98.5 us ]      -2.18%*
Styled - arabic 20 characters                      [   9.9 us ...   9.9 us ]      +0.41%
Styled - latin 20 characters                       [   5.3 us ...   5.3 us ]      -1.01%*
Styled - japanese 20 characters                    [   8.8 us ...   8.7 us ]      -0.60%*
Styled - arabic 1 paragraph                        [  51.3 us ...  51.0 us ]      -0.60%
Styled - latin 1 paragraph                         [  21.4 us ...  20.9 us ]      -2.52%*
Styled - japanese 1 paragraph                      [  78.7 us ...  77.5 us ]      -1.52%*
Styled - arabic 4 paragraph                        [ 232.7 us ... 232.1 us ]      -0.27%
Styled - latin 4 paragraph                         [  83.2 us ...  80.6 us ]      -3.18%*
Styled - japanese 4 paragraph                      [ 109.7 us ... 108.3 us ]      -1.34%*
```

[...]

PR #646, opened 2026-06-23
Bump `tango` and remove `parley_bench` build script

tomcur:
Tango now works without options/build scripts, see https://github.com/bazhenov/tango/pull/80. Denis, maintainer of Tango, kindly notified us of this in https://github.com/linebender/parley/pull/645#discussion_r3458322271!

(The bump is not strictly necessary I believe, but ensures our config is synchronized with the README of the `tango` rev we point to.)


taj-p (approved):
🎉

PR #647, opened 2026-06-23
Move bidi algorithm into `parley_core`

tomcur:
Kicking off the sequence to end up at (roughly) the shape of https://github.com/linebender/parley/pull/634.

Instead of starting with `parley_data`, I figured to first move the bidi algorithm as it wires up the `parley` and `parley_core` crates plus handles some lints that `parley` suppressed and `parley_core` doesn't.

Two more PRs are ready: first, simply moving the line break opportunity overrides introduced in https://github.com/linebender/parley/pull/640 (similarly moving files and handling lints), second and more interestingly, migrating `parley`'s analysis without any real changes into `parley_core` and introducing  `parley_core::{Analysis, Analyzer}`.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2904,6 +2904,7 @@ dependencies = [
  "linebender_resource_handle",
  "oxipng",
  "parlance",
+ "parley_core",
  "parley_data",
  "parley_dev",
  "peniko",
@@ -2924,7 +2925,10 @@ dependencies = [
 
 [[package]]
 name = "parley_core"
-version = "0.0.0"
+version = "0.11.0"
+dependencies = [
+ "icu_properties",
+]
 
 [[package]]
 name = "parley_data"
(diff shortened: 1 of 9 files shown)


tomcur:
I've posed a question about how exactly we want to go about merging these PRs in https://github.com/linebender/parley/pull/634#issuecomment-4780940217.

DJMcNab (approved):
I think we should move the conversation about how we land this to Zulip, but this pr on its own is good to land (with awareness that we'll potentially need to reconsider what is `pub` later)

Reading this PR makes me really want to refactor/clean-up/better document the bidi.rs file. The current documentation of input/output types is really far from ideal.

tomcur:
> I think we should move the conversation about how we land this to Zulip

I agree! Wrote here: [#parley > parley_core PR sprint? @ 💬](https://xi.zulipchat.com/#narrow/channel/205635-parley/topic/parley_core.20PR.20sprint.3F/near/606191279).

PR #648, opened 2026-06-23
Move line break opportunity overrides into `parley_core`

tomcur:
On top of https://github.com/linebender/parley/pull/647.

Temporarily removes `#[non_exhaustive]` from `LineBreakContext` to allow `parley` to construct it (becomes `non_exhaustive` again once the next PR lands).

PR #649, opened 2026-06-23
Migrate analysis from `parley` to `parley_core`

tomcur:
On top of #648.

This introduces `Analyzer` and `Analysis` as scratch, moves over `analyze_text` from Parley into Parley Core, as well as some types like `CharInfo`, `Boundary`. Parley now calls into Parley Core for analysis.

Analysis is unchanged, except Parley Core is given the word break config spans by Parley, where Parley previously reached into the style runs during analysis.

`CharInfo` has different fields in https://github.com/linebender/parley/pull/634 (the big PR where all of shaping was migrated to `parley_core`), and we may move to something like that later. For now, this is probably fine.

<details>
<summary>Benches neutral as expected</summary>

```
$ cargo export target/benchmarks -- bench --bench=main
$ cargo bench -q --bench=main -- compare ../target/benchmarks/main -t 8.
Default Style - arabic 20 characters               [   8.7 us ...   8.7 us ]      -0.26%
Default Style - latin 20 characters                [   4.1 us ...   4.2 us ]      +1.69%*
Default Style - japanese 20 characters             [   8.1 us ...   8.2 us ]      +0.49%
Default Style - arabic 1 paragraph                 [  47.0 us ...  46.7 us ]      -0.66%
Default Style - latin 1 paragraph                  [  16.0 us ...  16.5 us ]      +2.78%*
Default Style - japanese 1 paragraph               [  69.1 us ...  68.7 us ]      -0.55%
Default Style - arabic 4 paragraph                 [ 203.1 us ... 196.7 us ]      -3.13%*
Default Style - latin 4 paragraph                  [  61.5 us ...  61.7 us ]      +0.39%
Default Style - japanese 4 paragraph               [  98.1 us ...  97.0 us ]      -1.14%*
Styled - arabic 20 characters                      [   9.7 us ...   9.7 us ]      +0.04%
Styled - latin 20 characters                       [   5.3 us ...   5.3 us ]      +1.55%*
Styled - japanese 20 characters                    [   8.6 us ...   8.6 us ]      -0.46%
Styled - arabic 1 paragraph                        [  49.3 us ...  49.0 us ]      -0.76%
Styled - latin 1 paragraph                         [  20.4 us ...  20.6 us ]      +0.93%
Styled - japanese 1 paragraph                      [  75.0 us ...  74.4 us ]      -0.90%
Styled - arabic 4 paragraph                        [ 220.1 us ... 215.6 us ]      -2.02%*
Styled - latin 4 paragraph                         [  79.5 us ...  80.0 us ]      +0.59%
Styled - japanese 4 paragraph                      [ 106.2 us ... 105.3 us ]      -0.85%
```
</details>

### DIFF
--- a/parley/Cargo.toml
+++ b/parley/Cargo.toml
@@ -34,13 +34,11 @@ fontique = { workspace = true }
 parlance = { workspace = true }
 parley_core = { workspace = true }
 core_maths = { version = "0.1.1", optional = true }
-parley_data = { workspace = true, features = ["baked"] }
 accesskit = { workspace = true, optional = true }
 hashbrown = { workspace = true }
 harfrust = { workspace = true }
 icu_normalizer = { workspace = true, features = ["compiled_data"] }
 icu_properties = { workspace = true, features = ["compiled_data"] }
-icu_segmenter = { workspace = true, features = ["compiled_data"] }
 
 [dev-dependencies]
 parley_dev = { workspace = true }
--- a/parley/Cargo.toml
+++ b/parley/Cargo.toml
@@ -18,14 +18,21 @@ workspace = true
 
 [features]
 default = ["system"]
-std = ["fontique/std", "harfrust/std", "peniko/std", "skrifa/std", "parlance/std"]
+std = [
+    "fontique/std",
+    "harfrust/std",
+    "parley_core/std",
+    "peniko/std",
+    "skrifa/std",
+    "parlance/std",
+]
 libm = ["fontique/libm", "harfrust/libm", "peniko/libm", "skrifa/libm", "dep:core_maths"]
 # Enables support for system font backends
 system = ["std", "fontique/system"]
 accesskit = ["dep:accesskit"]
 # Enables dictionary-based line and word breaking for complex scripts (CJK, Thai, Khmer, Lao, Myanmar).
 # When disabled, a lightweight segmenter is used that falls back to character-level breaks for those scripts.
-complex-scripts = []
+complex-scripts = ["parley_core/complex-scripts"]
 
 [dependencies]
 skrifa = { workspace = true }
--- a/parley/src/analysis.rs
+++ b/parley/src/analysis.rs
@@ -1,4 +1,4 @@
-// Copyright 2025 the Parley Authors
+// Copyright 2026 the Parley Authors
 // SPDX-License-Identifier: Apache-2.0 OR MIT
 
 pub(crate) mod cluster;
--- a/parley/src/analysis.rs
+++ b/parley/src/analysis.rs
@@ -10,6 +10,8 @@ use parley_core::break_overrides::LineBreakOverrideFn;
 
 use parley_core::{AnalysisDataSources, AnalysisOptions};
 
+use parlance::WordBreak;
+
 pub(crate) fn analyze_text<B: Brush>(
     lcx: &mut LayoutContext<B>,
     text: &str,
@@ -17,15 +19,14 @@ pub(crate) fn analyze_text<B: Brush>(
 ) {
     let text = if text.is_empty() { " " } else { text };
 
-    // Collect the style runs word breaks. Gaps are `WordBreak::Normal`, so only non-`Normal`s need
+    // Collect the style runs' word breaks. Gaps are `WordBreak::Normal`, so only non-`Normal`s need
     // an entry.
     lcx.word_break.clear();
-    lcx.word_break.extend(lcx.style_runs.iter().map(|sr| {
-        (
-            sr.range.clone(),
-            lcx.style_table[sr.style_index as usize].word_break,
-        )
-    }));
+    lcx.word_break
+        .extend(lcx.style_runs.iter().filter_map(|sr| {
+            let word_break = lcx.style_table[sr.style_index as usize].word_break;
+            (word_break != WordBreak::Normal).then(|| (sr.range.clone(), word_break))
+        }));
 
     let options = AnalysisOptions {
         word_break: &lcx.word_break,
--- a/parley/src/analysis.rs
+++ b/parley/src/analysis.rs
@@ -1,0 +1,57 @@
+// Copyright 2025 the Parley Authors
+// SPDX-License-Identifier: Apache-2.0 OR MIT
+
+pub(crate) mod cluster;
+
+use crate::{Brush, LayoutContext};
+
+use icu_properties::props::{GeneralCategory, Script};
+use parley_core::break_overrides::LineBreakOverrideFn;
+
+use parley_core::{AnalysisDataSources, AnalysisOptions};
+
+pub(crate) fn analyze_text<B: Brush>(
+    lcx: &mut LayoutContext<B>,
+    text: &str,
+    line_break_override: Option<&LineBreakOverrideFn>,
+) {
+    let text = if text.is_empty() { " " } else { text };
+
+    // Collect the style runs word breaks. Gaps are `WordBreak::Normal`, so only non-`Normal`s need
+    // an entry.
+    lcx.word_break.clear();
+    lcx.word_break.extend(lcx.style_runs.iter().map(|sr| {
+        (
+            sr.range.clone(),
+            lcx.style_table[sr.style_index as usize].word_break,
+        )
+    }));
+
+    let options = AnalysisOptions {
+        word_break: &lcx.word_break,
+        line_break_override,
+    };
+    lcx.analyzer.analyze(text, &options, &mut lcx.analysis);
+
+    // Pair the char infos with style indices (which we set later in the builder).
+    lcx.info.clear();
+    lcx.info
+        .extend(lcx.analysis.char_info().iter().map(|&ci| (ci, 0)));
+}
+
+/// All characters contribute to shaping except:
+/// - Control characters
+/// - Format characters, unless they use the "Inherited" script
+#[inline(always)]
+pub(crate) fn contributes_to_shaping(general_category: GeneralCategory, script: Script) -> bool {
+    if matches!(
+        general_category,
+        GeneralCategory::Control
+            | GeneralCategory::LineSeparator
+            | GeneralCategory::ParagraphSeparator
+    ) {
+        return false;
+    }
+
+    !(general_category == GeneralCategory::Format && script != Script::Inherited)
+}
--- a/parley/src/analysis.rs
+++ b/parley/src/analysis.rs
@@ -33,11 +33,6 @@ pub(crate) fn analyze_text<B: Brush>(
         line_break_override,
     };
     lcx.analyzer.analyze(text, &options, &mut lcx.analysis);
-
-    // Pair the char infos with style indices (which we set later in the builder).
-    lcx.info.clear();
-    lcx.info
-        .extend(lcx.analysis.char_info().iter().map(|&ci| (ci, 0)));
 }
 
 /// All characters contribute to shaping except:
--- a/parley/src/analysis.rs
+++ b/parley/src/analysis.rs
@@ -42,6 +42,8 @@ pub(crate) fn analyze_text<B: Brush>(
 /// All characters contribute to shaping except:
 /// - Control characters
 /// - Format characters, unless they use the "Inherited" script
+// TODO: this function is duplicated at the time of writing, as it also exists in `parley_core`.
+// Once more of `parley` is moved to `parley_core`, this function should be removable.
 #[inline(always)]
 pub(crate) fn contributes_to_shaping(general_category: GeneralCategory, script: Script) -> bool {
     if matches!(
--- a/parley_core/src/analysis.rs
+++ b/parley_core/src/analysis.rs
@@ -48,6 +48,7 @@ impl Analysis {
     /// Create a reusable [`Analysis`].
     ///
     /// Pass this to [`Analyzer::analyze`].
+    #[inline]
     pub fn new() -> Self {
         Self::default()
     }
@@ -60,6 +61,7 @@ impl Analysis {
     }
 
     /// The per-character info in source order.
+    #[inline(always)]
     pub fn char_info(&self) -> &[CharInfo] {
         &self.info
     }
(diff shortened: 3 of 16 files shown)


parley/src/analysis.rs
```diff
@@ -0,0 +1,57 @@
...
+    lcx.info.clear();
+    lcx.info
+        .extend(lcx.analysis.char_info().iter().map(|&ci| (ci, 0)));
+}
+
+/// All characters contribute to shaping except:
+/// - Control characters
+/// - Format characters, unless they use the "Inherited" script
+#[inline(always)]
+pub(crate) fn contributes_to_shaping(general_category: GeneralCategory, script: Script) -> bool {
```
tomcur, line 44: Note this function is copied over from the old `parley/src/analysis/mod.rs`. That original file is moved to `parley_core/src/analysis.rs` (with changes). The remaining `parley/src/analysis/cluster.rs` should also end up moving, dropping the `analysis` dir entirely.

parley/src/analysis.rs
```diff
@@ -0,0 +1,57 @@
+// Copyright 2025 the Parley Authors
```
taj-p:
```suggestion
// Copyright 2026 the Parley Authors
```

parley_core/src/analyzer.rs
```diff
@@ -0,0 +1,62 @@
...
+
+use crate::{bidi::BidiResolver, break_overrides::LineBreakOverrideFn};
+
+use crate::analysis::{Analysis, analyze_text};
+
+/// Reusable scratch for [`Analyzer::analyze`].
+#[derive(Default)]
+pub struct Analyzer {
+    pub(crate) bidi: BidiResolver,
+    pub(crate) bidi_props: Vec<(BidiClass, BidiMirroringGlyph)>,
```
taj-p: I think `bidi_props` is a leak. It's pushed once per char but never cleared

parley/Cargo.toml
taj-p: This needs to forward the flag to `parley_core` (otherwise, I'm not sure it'll be honoured in `parley_core`).

parley/src/analysis.rs
taj-p: Will we remove this duplicate function soon then? Should we leave a `TODO`?

parley_core/src/analysis.rs
```diff
@@ -22,25 +22,79 @@ use icu_segmenter::{
...
+    /// Bidi level for each character, parallel to `info`.
+    ///
+    /// Empty if the text is all LTR.
+    pub(crate) levels: Vec<u8>,
+
+    /// The base bidi level of the paragraph of text.
+    pub(crate) paragraph_level: u8,
+}
+
+impl Analysis {
```
taj-p, line 47: Should these pass through methods be `#[inline(always)]`?

parley_core/src/analysis.rs
```diff
@@ -515,37 +607,39 @@ pub(crate) fn analyze_text<B: Brush>(
...
-                lcx.info
+                analysis
+                    .info
                     .iter()
-                    .map(|info| (info.0.bidi_class, info.0.bracket)),
+                    .map(|info| (info.bidi_class, info.bracket)),
             ),
             None,
         );
+        // TODO: perhaps core::mem::swap?
```
taj-p: Yea - I think this makes sense. Either by making `bidi.levels` `pub(crate)` or adding a helper:

```rs
    pub fn swap_levels(&mut self, dst: &mut Vec<BidiLevel>) {
        core::mem::swap(&mut self.levels, dst);
    }
```

parley/src/analysis.rs
```diff
@@ -0,0 +1,57 @@
...
+
+pub(crate) fn analyze_text<B: Brush>(
+    lcx: &mut LayoutContext<B>,
+    text: &str,
+    line_break_override: Option<&LineBreakOverrideFn>,
+) {
+    let text = if text.is_empty() { " " } else { text };
+
+    // Collect the style runs word breaks. Gaps are `WordBreak::Normal`, so only non-`Normal`s need
+    // an entry.
```
taj-p: Per this comment, should we be `.filter(...)`ing out `WordBreak::Normal` to take advantage of core filling unknown ranges with `WordBreak::Normal`? Or should we update this comment to correspond better to the code?

parley/src/analysis.rs
```diff
@@ -0,0 +1,57 @@
...
+    let options = AnalysisOptions {
+        word_break: &lcx.word_break,
+        line_break_override,
+    };
+    lcx.analyzer.analyze(text, &options, &mut lcx.analysis);
+
+    // Pair the char infos with style indices (which we set later in the builder).
+    lcx.info.clear();
+    lcx.info
+        .extend(lcx.analysis.char_info().iter().map(|&ci| (ci, 0)));
```
taj-p: We can avoid this additional work by storing the `CharInfo` and style index in parallel vecs rather than copying back to a tuple vec

parley/Cargo.toml
taj-p: Do we also need to forward to `parley_core/std`?

parley_core/src/analysis.rs
```diff
@@ -515,37 +607,39 @@ pub(crate) fn analyze_text<B: Brush>(
...
-                    is_variation_selector,
-                    is_region_indicator,
-                    general_category == GeneralCategory::Control,
-                    is_emoji_or_pictograph,
-                    contributes_to_shaping(general_category, script),
-                    force_normalize,
-                ),
-                0, // Style index is populated later
+            let bracket = data_sources.brackets().get(ch);
+            analyzer.bidi_props.push((bidi_class, bracket));
```
taj-p: I think this might be the leak?

taj-p (commented):
Largely LGTM! Added some comments

parley_core/src/analysis.rs
```diff
@@ -326,29 +423,29 @@ pub(crate) fn analyze_text<B: Brush>(
...
-        .expect("analyze_text requires at least one style run");
+    let mut segments = DenseWordBreaks::new(options.word_break, text.len());
+    // `text` is non-empty (checked above), so there is always at least one segment.
+    let first_segment = segments.next().unwrap();
+    let contiguous_word_break_substrings = WordBreakSegmentIter::new(text, segments, first_segment);
 
-    let contiguous_word_break_substrings =
-        WordBreakSegmentIter::new(text, rest_runs.iter(), lcx, first_style_run);
     let mut global_offset = 0;
     let mut line_boundary_positions: Vec<usize> = Vec::new();
```
taj-p, line 443: We should move this scratch space to `Analyzer` at some point 😅

parley_core/src/analysis.rs
taj-p: At some point, we might want to pass `levels` into `bidi.resolve` to circumvent this entirely

parley_core/src/analysis.rs
```diff
@@ -22,25 +22,79 @@ use icu_segmenter::{
...
+
+/// The result of [`Analyzer::analyze`].
+#[derive(Debug, Default)]
+pub struct Analysis {
+    /// Info for each character.
+    pub(crate) info: Vec<CharInfo>,
+
+    /// Bidi level for each character, parallel to `info`.
+    ///
+    /// Empty if the text is all LTR.
```
DJMcNab, line 40: Is the theory that even if we do have RTL, you're likely to have some LTR numerals in there, so need bidi awareness?

parley_core/src/analyzer.rs
tomcur: Whoops, I accidentally copied that over from the big `parley_core` PR, where we do use it (and clear it :sweat_smile:): https://github.com/linebender/parley/pull/634/changes#diff-5bf51df3f98cf8aad5e4e8931ac4daa515f616d1ad0d5530ecad7124c360eccaR542.

I've removed the field now, because in this new PR we still carry bidi info on `CharInfo`.

parley_core/src/analysis.rs
tomcur: Yeah, good point. Pushed that.

parley_core/src/analysis.rs
tomcur: Are you asking whether the `Vec` could also be empty in case the text is all RTL? I suppose we could do that; `parley` currently only optimizes the all-LTR case and this PR doesn't touch that:

https://github.com/linebender/parley/blob/98885084c0997e9d3f25bff29973870724a829f8/parley/src/shape/mod.rs#L126

I suppose in principle that could fall back to the paragraph level (which can be RTL) instead of to `0`.

parley_core/src/analysis.rs
tomcur: Yes (for posterity, see https://github.com/linebender/parley/pull/649#discussion_r3492249566).

parley/src/analysis.rs
tomcur: Yep, should happen when font selection is moved to `parley_core`. Added a todo!

parley_core/src/analysis.rs
tomcur: I've made it `pub(crate)` for now. I agree with passing `levels` into `bidi.resolve`.

parley/src/analysis.rs
tomcur: Filtering is right I think: it means the common case of all `WordBreak::Normal` is an empty vec.

parley/src/analysis.rs
tomcur: That makes sense to me. It'll be a bit of a bigger diff, as the existing `LayoutContext::info` is read in many places, but happy to do it as part of this PR (tomorrow)! https://github.com/linebender/parley/blob/98885084c0997e9d3f25bff29973870724a829f8/parley/src/context.rs#L36

taj-p: Whether it happens in this PR or as a separate PR, I don't mind - will leave this with you 🙏

parley_core/src/analysis.rs
```diff
@@ -230,54 +288,99 @@ pub(crate) enum Boundary {
...
+            Self {
+                word_break,
+                next: 0,
+                cursor: 0,
+                text_len,
+            }
+        }
+    }
+
+    impl Iterator for DenseWordBreaks<'_> {
```
taj-p, line 321: I considered whether we could just pass an iterator of `lcx.word_break` to avoid a collect, but since we iterate over the entire range, I think it's probably better to "collect all at once" and pass it in (as is currently proposed).

parley_core/src/analysis.rs
```diff
@@ -326,29 +427,29 @@ pub(crate) fn analyze_text<B: Brush>(
...
+    // `text` is non-empty (checked above), so there is always at least one segment.
+    let first_segment = segments.next().unwrap();
+    let contiguous_word_break_substrings = WordBreakSegmentIter::new(text, segments, first_segment);
 
-    let contiguous_word_break_substrings =
-        WordBreakSegmentIter::new(text, rest_runs.iter(), lcx, first_style_run);
     let mut global_offset = 0;
     let mut line_boundary_positions: Vec<usize> = Vec::new();
+
+    let data_sources = AnalysisDataSources::new();
```
taj-p, line 445: Thinking ahead: There might be a breaking API change in the future to pass this in because consumers may need to mutate `AnalysisDataSources` to pass in data at runtime. We can probably do that later - as needed - because I don't think stabilising the API ASAP is a high priority?

taj-p (approved):
LGTM 🎉

parley_core/src/analysis.rs
tomcur: I considered this too, it could work. It's mainly trading off simplicity and a collect versus the code either needing to be monomorphized or getting vtable indirection (and as we probably expect the collections to remain quite sparse, it's not necessarily a win).

parley/src/analysis.rs
tomcur: Done :)

<details>
<summary>Seems to be a slight win</summary>

```
$ cargo bench -q --bench=main -- compare ../target/benchmarks/main -t 8.
Default Style - arabic 20 characters               [   8.8 us ...   8.7 us ]      -1.33%*
Default Style - latin 20 characters                [   4.2 us ...   4.1 us ]      -1.38%*
Default Style - japanese 20 characters             [   8.0 us ...   8.0 us ]      -0.63%
Default Style - arabic 1 paragraph                 [  47.5 us ...  47.1 us ]      -0.82%
Default Style - latin 1 paragraph                  [  16.1 us ...  16.2 us ]      +0.28%
Default Style - japanese 1 paragraph               [  68.6 us ...  68.6 us ]      -0.06%
Default Style - arabic 4 paragraph                 [ 204.1 us ... 201.3 us ]      -1.41%*
Default Style - latin 4 paragraph                  [  61.2 us ...  60.9 us ]      -0.51%
Default Style - japanese 4 paragraph               [  97.4 us ...  97.3 us ]      -0.12%
Styled - arabic 20 characters                      [   9.8 us ...   9.7 us ]      -1.23%*
Styled - latin 20 characters                       [   5.3 us ...   5.3 us ]      -1.51%*
Styled - japanese 20 characters                    [   8.6 us ...   8.7 us ]      +0.76%
Styled - arabic 1 paragraph                        [  50.1 us ...  50.0 us ]      -0.23%
Styled - latin 1 paragraph                         [  20.7 us ...  20.5 us ]      -1.06%*
Styled - japanese 1 paragraph                      [  74.6 us ...  75.1 us ]      +0.73%
Styled - arabic 4 paragraph                        [ 224.5 us ... 222.6 us ]      -0.84%
Styled - latin 4 paragraph                         [  80.2 us ...  80.0 us ]      -0.27%
Styled - japanese 4 paragraph                      [ 106.0 us ... 106.1 us ]      +0.17%
```
</details>

parley_core/src/analysis.rs
tomcur: > We can probably do that later - as needed - because I don't think stabilising the API ASAP is a high priority?

That matches my understanding!

PR #652, opened 2026-06-24
Release Parley v0.11.0

nicoburns:
- Update changelog
- Update package version numbers
- Group local crates together in Cargo.toml so it's easier to find them when updating the version numbers

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,11 +12,29 @@ Subheadings to categorize changes are `added, changed, deprecated, removed, fixe
 
 This release has an [MSRV] of 1.88.
 
+## [0.11.0] - 2026-06-24
+
+This release has an [MSRV] of 1.88.
+
 ### Added
 
 #### Parley
 
+- Line break opportunity override (#640 by [@taj-p][]).
+  The builders now have a `set_line_break_override` method which can be used to customise where the parley detects line-breaking opportunities. It can be called with `Some(parley::CHROMIUM_LINE_BREAK_OVERRIDE)` to make line-breaking match Chromium, or with a custom instance of `AsciiLineBreakTable` created with `AsciiLineBreakTableBuilder`. If you do not call the function then you will get the existing standard ICU-defined behavior.
 - `From` conversions into `StyleProperty` for `FontWidth`, `FontStyle`, `FontWeight`, `WordBreak`, `OverflowWrap`, and `TextWrapMode`, matching the existing conversions for font family, variations, features, and line height. (#643 by [@waywardmonkeys][])
+
+### Changed
+
+#### Parley
+
+- Upgrade to `read-fonts` 0.40, `skrifa` 0.43, `harfrust` 0.10 (#650 by [@nicoburns][])
+
+### Fixed
+
+#### Parley
+
+- Fix font selection for emoji followed by a variation selector (#637 by [@kane50613][])
 
 ## [0.10.0] - 2026-06-01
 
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -12,12 +12,30 @@ Subheadings to categorize changes are `added, changed, deprecated, removed, fixe
...
+## [0.11.0] - 2026-06-24
+
+This release has an [MSRV] of 1.88.
+
 ### Added
 
 #### Parley
 
+- Line break opportunity override (#640 by [@taj-p][]).
+  The builders now have a `set_line_break_override` method which can be used to customise where the parley detects line-breaking opportunities. It can be called with `Some(parley::CHROMIUM_LINE_BREAK_OVERRIDE)` to make line-breaking match Chromium, or with a custom instance of `AsciiLineBreakTable` created with `AsciiLineBreakTableBuilder`. If you do not call the function then you will get the existing standard ICU-defined behavior.
```
tomcur:
```suggestion
  The builders now have a `set_line_break_override` method which can be used to customise where Parley detects line-breaking opportunities. It can be called with `Some(parley::CHROMIUM_LINE_BREAK_OVERRIDE)` to make line-breaking match Chromium, or with a custom instance of `AsciiLineBreakTable` created with `AsciiLineBreakTableBuilder`. If you do not call the function then you will get the existing standard ICU-defined behavior.
```

tomcur (approved):
Thank you! I like the new organization of `Cargo.toml` :)

One typo fix.

DJMcNab:
What was breaking here?

nicoburns:
I thought https://github.com/linebender/parley/pull/640 was, but perhaps it actually wasn't. `semver-checks` isn't picking anything up. I've been trying to avoid bumping fontations in patch releases though, as it makes it hard for consumers to ensure that there are no duplicate fontations dependencies.

DJMcNab:
Fwiw, we did special work to ensure that fontations wasn't a public dependency so that this didn't happen.

I don't think that the answer to "some consumers want to lock in a specific fontations version" is to force all consumers to not have the latest fixes and features. Consumers who care about this can make that tradeoff themselves by pinning their Parley dependency.

For this release, I'd rather that we released it as 0.10.1, but it wouldn't be the end of the world were we to release it as 0.11.0. That being said, I think we should resolve this question outside of your unilaterally deciding it.

One proposal I wouldn't rule out is a comprise of supporting two fontations versions. That is, we would support 0.12.0's fontations version in 0.12.1, and 0.12.1's (but not 0.12.0's) in 0.12.2.

That would mean that we could only update fontations just before a release, but that's the current state anyway. If something breaks when testing, we can just only have one supported version for that release. Concretely, I'd also want us to revert to a single supported version on main after each release was actually cut.

Thanks for doing the work here! It's really great to unblock Parley Core.

nicoburns:
We have precedent in `swash` of supporting a range of Fontations versions. We currently support `>= 0.31.1, <= 0.43` and nobody has reported issues. This will default to latest versions unless you pin it somehow (exact dependency or using `Cargo.lock`) but does at least allow people do that if they want (which I imagine both Blitz and Canva would want)

I feel like we're overindexing on having tested/verifed that a version, that we merely allow the user to use if they so choose, works. As a user, I don't really care about that. It will probably work, and if it doesn't then I have an easy migration path (probably just cargo update). Specifying our versions so that users *cannot* select versions that likely do actually work just because we don't want to guarantee it feels very user-hostile.

PR #655, opened 2026-06-25
Handle punctuation in the Linebreaking harness test

DJMcNab:
See #642, which implemented this test originally, and #640, which added the punctuation handling to match Chrome.

This PR combines three things:
- We use a few strategies for test case generation; the existing alphanumeric, a few templates (such as simple URLs and file paths), a few artificial cases to test specific ops and then a catchall ASCII "soup". My plan is to extend this to have a long-running in-browser fuzzer, for which I expect the soup to find any remaining discrepancies in the ASCII subset.
- Making the linebreaking harness's Parley side use the Chrome line breaking.
- It also fixes a bug found in #640 for how Chrome handles spaces, which are always a soft wrap opportunity, whereas UAX-14 specifies that they shouldn't be treated as such. 

This last item is a change in Parley itself.

Code originally generated using Claude Code - Opus 4.8 (1M) and Sonnet 4.6 (1M), then manually edited to be what I want.

PR #656, opened 2026-06-26
Add a basic fuzzer for Linebreaking

DJMcNab:
> [!WARNING]
> 
> Do not review! This is fully Claude generated, and has not been reviewed by me.

This implements a fuzzer which runs the same testing strategy as #642 and #655, but over "all the seeds". The outcome from this fuzzer is that running it for 3 million seeds, there weren't any detected discrepancies for ASCII text. Note that the fuzzer is not clean of cases where Parley breaks differently to chrome currently. However, all of those cases are places where Parley doesn't break for a case which is narrower than the width generated using Chrome, which isn't an issue for me, so I don't want to track it down.
The harness has also detected some instability in the way we perform our measurements of Chrome (possibly due to the rapid changing of fonts?). I've not dug into this further for the same reason.

If/when we extend the testing to more languages, we'll want to revive this fuzzer, but for now I'd suggest not putting the effort into getting this working.


nicoburns:
> However, all of those cases are places where Parley doesn't break for a case which is narrower than the width generated using Chrome, which isn't an issue for me, so I don't want to track it down.

Does that mean that there is a discrepancy, but it's in the computation of the "content widths"?

DJMcNab:
There were several of those, which is an issue. But there were many more cases where given the same input width, Parley and Chrome changed line breaking decision in lockstep (that is, despite Chrome having one extra subpixel of epsilon which we didn't apply to Parley). As Parley's width is always wider than Chrome, a naïve analysis suggests that shouldn't be possible.

nicoburns:
Hmmm.... subpixels shouldn't make any difference for layout? (because we're laying in `f32` and not quantizing for the purposes of line breaking). May well be that Chrome uses something with more precision than `f32` though.

DJMcNab:
I'm not going to put any more development effort into this. I didn't find any new failures based on this, which is good.

PR #657, opened 2026-06-28
Implement password masking

nicoburns:
Implements password masking in the Parley.

Generated with Opus 4.8 High. Reviewed and Rewritten by Fable 5.1 Low. Reviewed manually. PR description hand written. Builds on design work and discussion from https://github.com/linebender/parley/pull/394

### Changes made:

- Add `grapheme_replacement: Option<char>` style (using `char` makes this not fully general, but it works for the common case and keeps things simple)
- Add `ShapeOptions::grapheme_replacement` to `parley_engine`. When set, the `Shaper` segments input text by grapheme cluster and feeds one replacement character per grapheme cluster into HarfRust (using the grapheme's first character as the cluster id) instead of the real text, so the shaped output has exactly one glyph per grapheme while every cluster still maps back to its source characters.
- Add a `password: bool` field to `PlainEditor` (+getter/setter methods) that uses this functionality with the bullet character (U+2022).

### Notes

- This deliberately doesn't cache the replacement glyph, as it is assumed that where this is used in practice there will only be one text run (so it will only be looked up once anyway).
- ~~This doesn't send the replacement character through a full shaping pipeline, as I figured that isn't necessary for the bullet char (which is probably the only case anyone cares about?). But let me know if this is actually required for some reason~~ The latest version *does* do this as it leads to a much neater implementation.
- Shaping items are split when `grapheme_replacement` changes, so masking a sub-range works (not necessarily useful, but it wasn't any extra work to support it).

### Tested with Blitz

https://github.com/user-attachments/assets/13dea77e-30f0-4339-9293-248d6243cd7d

### DIFF
--- a/parley/src/editing/editor.rs
+++ b/parley/src/editing/editor.rs
@@ -108,6 +108,8 @@ where
     font_size: f32,
     scale: f32,
     quantize: bool,
+    /// Whether to mask the text with a bullet character, as for a password field.
+    password: bool,
     // Simple tracking of when the layout needs to be updated
     // before it can be used for `Selection` calculations or
     // for drawing.
@@ -141,6 +143,7 @@ where
             font_size,
             scale: 1.0,
             quantize: true,
+            password: false,
             layout_dirty: true,
             alignment: Alignment::Start,
             // We don't use the `default` value to start with, as our consumers
@@ -1029,6 +1032,21 @@ where
         self.layout_dirty = true;
     }
 
+    /// Set whether the text should be masked as a password.
+    ///
+    /// When enabled, every grapheme cluster is rendered with the bullet character (`•`) instead
+    /// of its actual glyphs. The underlying text is unchanged, so editing, cursor movement,
+    /// selection and hit-testing continue to operate on the real text.
+    pub fn set_password(&mut self, password: bool) {
+        self.password = password;
+        self.layout_dirty = true;
+    }
+
+    /// Get whether the text is masked as a password.
+    pub fn is_password(&self) -> bool {
+        self.password
+    }
+
     /// Modify the styles provided for this editor.
     pub fn edit_styles(&mut self) -> &mut StyleSet<T> {
         self.layout_dirty = true;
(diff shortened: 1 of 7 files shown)


parley/src/editing/editor.rs
```diff
@@ -1232,6 +1250,11 @@ where
         for prop in self.default_style.inner().values() {
             builder.push_default(prop.to_owned());
         }
+        if self.password {
+            // Mask the entire text with the bullet character. Pushed last so it overrides any
+            // grapheme replacement that may have been set on the default styles.
+            builder.push_default(StyleProperty::GraphemeReplacement(Some('\u{2022}')));
```
tronical, line 1256: We do something similar in Slint and what we've observed is that sometimes that character isn't present on folks systems (yes, reality is sad :). It might make sense to just build the API around password replacement characters and then have a fallback to `*` inside parley, let the `GraphemeReplacement` bit an array of options, or ignore the problem :)

(thanks for bringing this into parley proper, btw!)

nicoburns: Hmm... I wonder if a better option would be to allow the user to override the font. In Blitz we're bundling `moz-bullet-font.otf` that we got from Mozilla/Firefox which we also use for list item bullets. It's ~5kb (not entirely sure why it's so big - maybe just required metadata?).

nicoburns:
I think the replacement may need to go slightly earlier in the pipeline so that it occurs before font-fallback (but still after shaping) and we can fallback with the replaced glyph instead of the original glyph. Otherwise you can get issues (invisible glyphs) in cases such as emoji where the font that matches for the original glyph doesn't have a glyph for the replacement glyph.

parley/src/editing/editor.rs
nicoburns: The latest version in this PR sends the replacement character through the usual shaping and font-fallback pipeline, so it should work if the user has *any* font on their system with the required character.

nicoburns:
I've done a second pass on this (with some assistance from Fable). The new version does replacement prior to shaping and seems significantly simpler (and also more correct - things like emoji are now being handled properly) than the old version. @tomcur What do you think?

PR #658, opened 2026-06-30
Only lowercase once in `GenericFamily::parse`

nicoburns:
Use a fixed-size buffer to only lowercase once in `GenericFamily::parse`. Not 100% sure that this will be a win, but I'd guess this is faster?

### DIFF
--- a/parlance/src/generic_family.rs
+++ b/parlance/src/generic_family.rs
@@ -64,34 +64,35 @@ impl GenericFamily {
     /// ```
     pub fn parse(s: &str) -> Option<Self> {
         let s = s.trim();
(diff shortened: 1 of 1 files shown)


tomcur (approved):
LGTM!

This more or less matches what we do in `color`.

E.g., https://github.com/linebender/color/blob/ed0438355faaed213cff35ab8e57c64d65329a36/color/src/parse.rs#L693-L706 and https://github.com/linebender/color/blob/ed0438355faaed213cff35ab8e57c64d65329a36/color/src/parse.rs#L469-L480.

DJMcNab:
Backlinking to https://github.com/linebender/parley/pull/654

PR #661, opened 2026-06-30
Remove `Glyph::style_index`, allow access to indices on `Cluster` and `GlyphRun`

tomcur:
On top of https://github.com/linebender/parley/pull/660.

A glyph's style index is equal to its owning `Cluster`'s style index, and style indices are constant by construction in `GlyphRun`. As those are the two containers through which a user can get at a `Glyph`, the `Glyph`s themselves don't need to carry the style index.

This is motivated by `parley::layout::Glyph` moving to `parley_core` soon, and so far `parley_core` does not encode style indices at all.

Added a changelog entry, but if `Glyph` is migrated to `parley_core` (without a shim in `parley`) before the next release, this entry will likely get absorbed into a bigger entry.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,12 @@ Subheadings to categorize changes are `added, changed, deprecated, removed, fixe
 
 This release has an [MSRV] of 1.88.
 
+### Changed
+
+#### Parley
+
+- Breaking change: the `Glyph::style_index` field was removed. Use `Clusters::{style, style_index}` or `GlyphRun::{style, style_index}` instead. (#661 by [@tomcur][])
+
 ## [0.11.0] - 2026-06-24
 
 This release has an [MSRV] of 1.88.
@@ -681,6 +687,7 @@ This release has an [MSRV][] of 1.70.
 [#640]: https://github.com/linebender/parley/pull/640
 [#643]: https://github.com/linebender/parley/pull/643
 [#650]: https://github.com/linebender/parley/pull/650
+[#661]: https://github.com/linebender/parley/pull/661
 
 [Unreleased]: https://github.com/linebender/parley/compare/v0.11.0...HEAD
 [0.11.0]: https://github.com/linebender/parley/compare/v0.10.0...v0.11.0
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,7 +16,7 @@ This release has an [MSRV] of 1.88.
 
 #### Parley
 
-- Breaking change: the `Glyph::style_index` field was removed. Use `Clusters::{style, style_index}` or `GlyphRun::{style, style_index}` instead. (#661 by [@tomcur][])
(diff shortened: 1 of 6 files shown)


nicoburns:
Does (/will) `parley_core` not also deal with clusters and mapping their style indexes?

tomcur:
Maybe; in the prototype big PR, `parley_core` does not hold style indices. There, users (like `parley`) instead feed their relevant shaping boundaries into `parley_core` when they call itemization (such boundaries are things like font size, but not things like decoration or color). Users that style text will probably hold on to something like a parallel vector of styles. All of this can be litigated of course.

While this change is motivated by `parley_core` not encoding it (as I had to shim some of this while working on the migration), I think this change does make sense by itself: currently, `Glyph` is redundantly encoding the style index, which (if nothing else) makes it confusing for users as to where the style actually comes from, and whether it can ever be different from the owning `GlyphRun` or `Cluster`.

nicoburns (approved):
I'm a bit skeptical of the broader change of `parley_core` not tracking styles at all (that seems like it might be inconvenient!), but this change on it's own makes sense to me.

I'm partially trusting you that this is indeed redundant. I have noticed that this *seemed* to be the case before, but never checked thoroughly.

DJMcNab:
I just want to sanity check how this interacts with https://faultlore.com/blah/text-hates-you/#style-can-change-mid-ligature? Obviously we don't handle this case 'correctly' currently, but I would in theory like for a consumer who really cares to be able to handle this.

(I do also like the aside of "Don’t ask about the code which line-breaks partial ligatures though.")

To be clear, this concern definitely doesn't block this PR...

Edit: oh, this was discussed in a different pr a bit; #660

tomcur:
> I just want to sanity check how this interacts with https://faultlore.com/blah/text-hates-you/#style-can-change-mid-ligature? Obviously we don't handle this case 'correctly' currently, but I would in theory like for a consumer who really cares to be able to handle this.
> [..]
> Edit: oh, this was discussed in a different pr a bit; https://github.com/linebender/parley/pull/660

`parley` has a few cluster types, a glyph run type, and this `Glyph` type.

The `Cluster` (which #660 touches) corresponds to a single character, always has exactly one style, and has 0 or more `Glyph`s. In the case of a ligature, each character of the ligature has a corresponding `Cluster`, but I believe only the one with a `LIGATURE_START` flag gets glyphs. We use the remaining ones with `LIGATURE_COMPONENT` for selection (the ligature's glyph's advance is spread over the owning `Cluster`s' advances). Firefox's behavior for styling is in principle recoverable from `Cluster`, I believe.

`GlyphRun`, on the other hand, flatmaps the `Cluster`s' glyphs, ignoring clusters without glyphs. For rendering, I believe that matches Chrome.

nicoburns:
> Obviously we don't handle this case 'correctly' currently, but I would in theory like for a consumer who really cares to be able to handle this.

I do agree, but want to note that I don't think there's consensus that Firefox's handling here is correct! (it's kinda one of those cases where nothing is really good).

PR #662, opened 2026-06-30
Remove `glifo`, `vello_hybrid` and `vello` dependencies from the workspace

nicoburns:
A big diff, but mostly removing code, and can be reviewed commit-by-commit. I can break this up if necessary.

- The snapshot updates are mostly from the `vello_cpu` version bump (and are mostly non-visible changes to me)
- Some are from the `glifo` -> `vello_cpu` change.
- The scaling for emojis with skew breaks when switching from `glifo` -> `vello_cpu`, but is fixed by the `vello_cpu` upgrade


nicoburns:
Hmm... tests are passing locally (on macOS!) but failing in CI (on macOS). This seems very strange to me given that we're rendering with vello_cpu...

DJMcNab:
I'd prefer to review this as three PRs, as they can each land independently.
- Removing the drawing benchmarks and vello_hybrid_render (which I propose doing in https://github.com/linebender/parley/pull/665)
- Removing the use of Glifo directly elsewhere.
- Porting the editor example to be Vello CPU only.

In particular, I think the first two will be significantly easier to review than the last, so don't want to mix in any minor review comments from them into the review of this last item.

waywardmonkeys:
Agree on splitting this up.

nicoburns:
I've extracted:
- https://github.com/linebender/parley/pull/668
- https://github.com/linebender/parley/pull/669

as the first steps here, in addition to Daniel's:
- https://github.com/linebender/parley/pull/665

These should be simpler to review than the other changes as they don't affect the tests.

The remaining changes after those are:

- Port tests to `vello_cpu` glyph rendering (remove `glifo`)
- Upgrade `vello_cpu` to latest (`0.0.9`)

We may want to do that as one step as:

- We can't upgrade `vello_cpu` without first removing `glifo` as newer `glifo` won't work
- Newer `vello_cpu` contains fixes to glyph rendering

DJMcNab:
We have landed those three PRs, but still do want the rest of this.

PR #663, opened 2026-06-30
Include all coverage variants for a family in a query

### LINKED ISSUE #322: Support resolving multiple font files in the same file covering different unicode ranges
Consider this `@font-face` configuration (taken from https://servo.org):

```css
/* latin */
@font-face {
  font-family: 'Space Grotesk';
  font-style: normal;
  font-weight: 400;
  font-stretch: 100%;
  src: url(https://fonts.bunny.net/space-grotesk/files/space-grotesk-latin-400-normal.woff2) format('woff2'), url(https://fonts.bunny.net/space-grotesk/files/space-grotesk-latin-400-normal.woff) format('woff'); 
  unicode-range: U+0000-00FF,U+0131,U+0152-0153,U+02BB-02BC,U+02C6,U+02DA,U+02DC,U+0304,U+0308,U+0329,U+2000-206F,U+20AC,U+2122,U+2191,U+2193,U+2212,U+2215,U+FEFF,U+FFFD;
}

/* latin-ext */
@font-face {
  font-family: 'Space Grotesk';
  font-style: normal;
  font-weight: 400;
  font-stretch: 100%;
  src: url(https://fonts.bunny.net/space-grotesk/files/space-grotesk-latin-ext-400-normal.woff2) format('woff2'), url(https://fonts.bunny.net/space-grotesk/files/space-grotesk-latin-ext-400-normal.woff) format('woff'); 
  unicode-range: U+0100-02BA,U+02BD-02C5,U+02C7-02CC,U+02CE-02D7,U+02DD-02FF,U+0304,U+0308,U+0329,U+1D00-1DBF,U+1E00-1E9F,U+1EF2-1EFF,U+2020,U+20A0-20AB,U+20AD-20C0,U+2113,U+2C60-2C7F,U+A720-A7FF;
}

/* vietnamese */
@font-face {
  font-family: 'Space Grotesk';
  font-style: normal;
  font-weight: 400;
  font-stretch: 100%;
  src: url(https://fonts.bunny.net/space-grotesk/files/space-grotesk-vietnamese-400-normal.woff2) format('woff2'), url(https://fonts.bunny.net/space-grotesk/files/space-grotesk-vietnamese-400-normal.woff) format('woff'); 
  unicode-range: U+0102-0103,U+0110-0111,U+0128-0129,U+0168-0169,U+01A0-01A1,U+01AF-01B0,U+0300-0301,U+0303-0304,U+0308-0309,U+0323,U+0329,U+1EA0-1EF9,U+20AB;
}
```

Source:  https://fonts.bunny.net/css?family=space-grotesk:400

Here we have 3 fonts with the same family name and attributes differing only in unicode range coverage. My reading of our current font fallback code is that only one of these will ever be tried because it is assumed that there will only be one font in each family with each uniqe set of attributes.

We should ensure that our font querying supports this case.

kane50613:
Fixes #322

_Generated with Claude and reviewed by me_

Add `matching::match_fonts` to collect every font sharing the best match's attributes, and have the query include the whole set. The caller already walks candidates per cluster against each font's charmap, so it now lands on the slice that covers the cluster.

### DIFF
--- a/fontique/src/collection/query.rs
+++ b/fontique/src/collection/query.rs
@@ -5,10 +5,11 @@
 
 use crate::{Charmap, CharmapIndex};
 
-use super::super::{Collection, SourceCache};
+use super::super::{Collection, SourceCache, matching::match_fonts};
 
 use alloc::vec::Vec;
 use parlance::Script;
+use smallvec::SmallVec;
 
 use super::{
     super::{Attributes, Blob, FallbackKey, FamilyId, FamilyInfo, GenericFamily, Synthesis},
@@ -140,22 +141,22 @@ impl<'a> Query<'a> {
             let Entry::Ok(family_info) = &family.family else {
                 continue;
             };
-            let mut best_index = None;
-            if let Some(font) = load_font(
+            let bucket = load_bucket(
                 family_info,
                 self.attributes,
                 &mut family.best,
-                false,
                 self.source_cache,
-            ) {
-                best_index = Some(font.family.1);
+            );
+            let mut yielded_default = false;
+            for font in bucket {
+                if font.family.1 == family_info.default_font_index() {
+                    yielded_default = true;
+                }
                 if f(font) == QueryStatus::Stop {
                     return;
                 }
             }
-            // Don't invoke for the default font if it's the same as the
-            // best match.
-            if best_index == Some(family_info.default_font_index()) {
+            if yielded_default {
                 continue;
             }
             if let Some(font) = load_font(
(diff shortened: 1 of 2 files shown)


fontique/src/collection/query.rs
DJMcNab: It isn't clear to me why `load_font` still has the `is_default` parameter. It doesn't seem to ever be set to false?

fontique/src/collection/query.rs
```diff
@@ -244,6 +245,58 @@ impl QueryFont {
...
+                });
+            }
+            if fonts.is_empty() {
+                return &[];
+            }
+            *status = Entry::Ok(fonts);
+            if let Entry::Ok(fonts) = status {
+                fonts
+            } else {
+                &[]
```
DJMcNab: How is this ever reachable? If it isn't, it should use `unreachable!`

fontique/src/matching.rs
```diff
@@ -9,6 +9,24 @@ use smallvec::SmallVec;
 
 const DEFAULT_OBLIQUE_ANGLE: f32 = 14.0;
 
+pub fn match_fonts(
```
DJMcNab:
```suggestion
pub(crate) fn match_fonts(
```
(You couldn't be expected to know this - essentially, the neighbouring code has bad use of `pub`, which we intend to fix at some point)

fontique/src/matching.rs
```diff
@@ -9,6 +9,24 @@ use smallvec::SmallVec;
...
+    weight: FontWeight,
+    synthesize_style: bool,
+) -> impl Iterator<Item = usize> + '_ {
+    let best = match_font(set, width, style, weight, synthesize_style).map(|index| {
+        let font = &set[index];
+        (font.width(), font.style(), font.weight())
+    });
+    set.iter().enumerate().filter_map(move |(index, font)| {
+        let (best_width, best_style, best_weight) = best?;
+        (font.width() == best_width && font.style() == best_style && font.weight() == best_weight)
```
DJMcNab, line 26: I'd like this to have a comment:

```suggestion
        // We need to return each font with the best attributes, as there might be multiple subsets.
        (font.width() == best_width && font.style() == best_style && font.weight() == best_weight)
```
As currently this reads quite confusingly.

DJMcNab (approved):
I'm happy enough for this to land as-is; it does address a real case fairly surgically.

However, I do have a couple of follow-up questions:
1) Is this definitely the right behaviour; we gate the 'later' fonts on being 'exactly as good' as the best match?
2) Presumably, a better long-term fix would be to actually implement unicode range support. Would we still need this in that case?

@kane50613, let me know when you're happy and I'll merge it.

kane50613:
@DJMcNab Thanks for reviewing and I've guided Claude to research on how Blink does this

1. Blink does the same: [FontFaceCache::GetOrCreate](https://source.chromium.org/chromium/chromium/src/+/main:third_party/blink/renderer/core/css/font_face_cache.cc;l=227;bpv=0;bpt=1) groups faces by attributes and font selection picks a single best bucket. 
Worse-attribute faces never fill coverage gaps, those fall through to the next family. Mixing weights/styles within one run would be worse than falling back.

3. Yes, we'd still need this. Blink builds unicode-range on the same grouping: equal-attribute faces form one [CSSSegmentedFontFace](https://source.chromium.org/chromium/chromium/src/+/main:third_party/blink/renderer/core/css/css_segmented_font_face.cc), with ranges filtering within the bucket. 
It would also fix a downside of this PR: `load_bucket` eagerly loads every variant's blob, since without range metadata the only way to check coverage is the charmap. With ranges, subsets outside the needed range are never loaded.

PR #665, opened 2026-06-30
Remove Parley's drawing benchmarks and Hybrid example

DJMcNab:
We discussed this in [#office hours > Text 2026-06-30](https://xi.zulipchat.com/#narrow/channel/359642-office-hours/topic/Text.202026-06-30/with/607486082). Essentially, these benchmarks are entirely benchmarking Vello CPU; it doesn't make any sense for this repository to own them. Their performance is unrelated to any choices we would make here...

cc @grebmeg FYI, you added some of these benchmarks for the LRU cache. There were also benchmarks of ink-skipping added by @valadaptive. You might want to migrate those into the Vello repository, if you think they'd be valuable to you.


DJMcNab:
Oh, oops. This was already done in https://github.com/linebender/parley/pull/662.

Edit: I still think we should land this PR

nicoburns (approved):

DJMcNab:
Wow, how on earth did this get a conflict *today*...

Edit: oh, it was Rust 2024 😆

Issue #555, opened 2026-02-17
Refine line breaking around graphemes and grapheme clusters


conor-93:
CRLF is treated as two line breaks, [though it should be treated as one](https://www.unicode.org/reports/tr13/tr13-9.html#Interpreting_characters_in_text:~:text=on%20input%20and%20in%20interpretation%2C%20treat%20CR%2C%20LF%2C%20CRLF%2C%20and%20NEL%20the%20same). In this case, we'd want to break around the whole grapheme cluster `\r\n`.

Using `examples/vello_cpu_render/src/main.rs`:

```rust
let text = String::from(
        "Start\nSome text here. Let's make it a bit longer so that line wrapping kicks in 😊. And also some اللغة العربية arabic text.\r\nThis is underlining pq and strikethrough t",
    );
```
<img width="236" height="257" alt="Image" src="https://github.com/user-attachments/assets/e940b539-9e2d-43d6-8407-e0d6436cdad3" />

```rust
let text = String::from(
        "Start\r\nSome text here. Let's make it a bit longer so that line wrapping kicks in 😊. And also some اللغة العربية arabic text.\r\nThis is underlining pq and strikethrough t",
    );
```
<img width="236" height="278" alt="Image" src="https://github.com/user-attachments/assets/43726e33-3e84-4cf0-a067-00540a864054" />

However, for ligatures, which form a single grapheme cluster, we still need to wrap on characters, e.g if layout width is 1px:
Current behaviour:
<img width="74" height="69" alt="Image" src="https://github.com/user-attachments/assets/9d09454e-72f9-403d-8914-c8177a111169" />
Expected:
<img width="31" height="240" alt="Image" src="https://github.com/user-attachments/assets/2eaad233-6f27-4005-87e8-07b4f33950df" />

mvanhorn:
Implemented in https://github.com/linebender/parley/pull/667 - coalesces CR+LF into a single hard line break at the line-breaking layer so CRLF no longer produces a spurious empty line.

PR #667, opened 2026-07-01
Treat CRLF as a single hard line break

mvanhorn:
## What

A `\r\n` (CRLF) sequence is currently treated as two hard line breaks instead of one, so a document authored with CRLF line endings renders with a spurious empty line between every line compared with the same text using `\n`.

## Why

`to_whitespace` in `layout/data.rs` classifies both `'\r'` and `'\n'` as `Whitespace::Newline`, so a CRLF surfaces as two consecutive newline clusters. In `break_next`, the `is_newline` branch emits `BreakReason::Explicit` for each of them, producing two lines.

Per UAX#14 (CR × LF, do not break between) and the grapheme-cluster framing raised in the issue thread, CR, LF, and CRLF should each be a single hard break (CR+LF is one grapheme cluster).

## How

The fix stays at the line-breaking layer and does not touch `to_whitespace` / `ClusterInfo`, which are used broadly by cursor and selection code.

In `break_next`, when a newline cluster is a `'\r'` immediately followed by a `'\n'` newline, the `'\r'` is appended to the current line but the break is suppressed there, letting the following `'\n'` emit the single hard break so the CR and LF share one line. The lookahead reads the global cluster list, so it also coalesces a CRLF that shaping split across two runs (for example a style or font boundary starting at the `\n`). It requires the LF to be item-adjacent to the CR, so an inline box positioned between the two still keeps the CR's break. Lone `\r`, lone `\n`, LS (U+2028), and PS (U+2029) behaviour is unchanged.

## Scope

This covers the CRLF-as-one-hard-break part of the issue. The separate case mentioned in the thread, where an over-long ligature that forms a single grapheme cluster still needs character-level emergency wrapping, lives in a different code path and is left as follow-up, so this references rather than closes the issue.

## Tests

Added two tests in `tests/test_builders.rs`:

- `builders_crlf_counts_as_single_line_break`: `"a\r\nb"` yields the same line count as `"a\nb"` and `"a\rb"` (2 lines, no blank line between); doubled `"a\r\n\r\nb"` matches `"a\n\nb"` (3 lines, one blank line) rather than producing extra blanks; a trailing `"a\r"` matches `"a\n"`.
- `builders_crlf_across_run_boundary_counts_as_single_line_break`: a style change starting at the `\n` forces shaping to split the CRLF across two runs, and the pair still coalesces into a single break.

Happy to add a `CHANGELOG.md` entry under `[Unreleased]` once this has a PR number, if that is preferred.

Refs #555

### DIFF
--- a/parley/src/layout/line_break.rs
+++ b/parley/src/layout/line_break.rs
@@ -584,8 +584,41 @@ impl<'a, B: Brush> BreakLines<'a, B> {
                             if max_height_exceeded {
                                 return self.max_height_break_data(line_height);
                             }
+                            // A CRLF sequence is a single grapheme cluster and must produce
+                            // exactly one hard line break (UAX#14: CR × LF, do not break
+                            // between). When this newline is a CR immediately followed by an
+                            // LF, append the CR to the current line but suppress the break
+                            // here and let the LF emit the single break, so CR and LF share
+                            // one line. The lookahead reads the global cluster list so a CRLF
+                            // that shaping split across two runs (e.g. a style boundary at the
+                            // LF) is still coalesced. The LF must be item-adjacent to the CR:
+                            // if it lands in a later run it only coalesces when the next item
+                            // is that run (not an inline box sitting between the two), so an
+                            // inline box at the LF offset keeps the CR's break. Lone CR, lone
+                            // LF, LS, and PS are unaffected.
+                            let lf_is_item_adjacent = self.state.cluster_idx + 1 < cluster_end
+                                || self
+                                    .layout
+                                    .data
+                                    .items
+                                    .get(self.state.item_idx + 1)
+                                    .is_some_and(|item| item.kind == LayoutItemKind::TextRun);
+                            let is_cr_before_lf = cluster.info().source_char() == '\r'
+                                && lf_is_item_adjacent
+                                && self
+                                    .layout
+                                    .data
+                                    .clusters
+                                    .get(self.state.cluster_idx + 1)
+                                    .is_some_and(|next| {
+                                        next.info.whitespace() == Whitespace::Newline
+                                            && next.info.source_char() == '\n'
+                                    });
                             self.state
                                 .append_cluster_to_line(self.state.line.x, line_height);
+                            if is_cr_before_lf {
+                                continue;
+                            }
                             return self.start_new_line(
                                 BreakReason::Explicit,
                                 max_advance,
--- a/parley/src/tests/test_builders.rs
+++ b/parley/src/tests/test_builders.rs
@@ -544,3 +544,87 @@ fn builders_mixed_styles() {
         false,
     );
 }
+
+#[test]
+fn builders_crlf_counts_as_single_line_break() {
+    let mut fcx = create_font_context();
+    let mut line_count = |text: &str| -> usize {
+        let mut lcx: LayoutContext<ColorBrush> = LayoutContext::new();
+        let ropts = RangedOptions {
+            scale: 1.0,
+            quantize: false,
+            max_advance: None,
+            text,
+        };
+        let layout = build_layout_with_ranged(&mut fcx, &mut lcx, &ropts, |rb| {
+            set_root_style(rb);
+        });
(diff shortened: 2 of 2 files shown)


DJMcNab:
So that we're all on the same page, can you confirm whether you've read https://linebender.org/wiki/llm-policy/? We know that we do need to link to this more prominently, so it's reasonable if you haven't.

I'm keen to see this fixed, although I'm worried that this will conflict with our ongoing work to move the code into Parley core (see https://github.com/linebender/parley/pull/634 and the PRs resulting from it).

nicoburns:
> I'm keen to see this fixed, although I'm worried that this will conflict with our ongoing work to move the code into Parley core (see https://github.com/linebender/parley/pull/634 and the PRs resulting from it).

From a "git" point of view, I suspect changes to `line_break.rs` are unlikely to conflict with parley core work. There could of course be semantic changes. Intuitively, it seems to me that the correct fix for this is to combine CLRF into one cluster rather than handling it during line breaking, but I'm not confident in that (@tomcur do you have thoughts?).

nicoburns (changes requested):
I think this could be handled more simply:

Just below `let cluster = run.get(self.state.cluster_idx - cluster_start).unwrap();`:

```rust
// Handle CRLF: if the current cluster represents an CR and the next char is an LF
// then we skip the CR and just process the LF. This allows us to treat the combined CRLF
// as a single line break rather than as two line breaks.
let cluster = if cluster.info().source_char() == '\r'
    && self.state.cluster_idx + 1 < cluster_end
    && let Some(next_cluster) =
        run.get(self.state.cluster_idx + 1 - cluster_start)
{
    self.state.cluster_idx += 1;
    next_cluster
} else {
    cluster
};
```

tomcur (commented):
> Intuitively, it seems to me that the correct fix for this is to combine CLRF into one cluster rather than handling it during line breaking, but I'm not confident in that (@tomcur do you have thoughts?).

There are a few cluster types. The one here is a `Cluster`, which intentionally maps one-to-one with a character. E.g., the text "ffi" has three clusters, it's three graphemes, and it may shape to a single ligature glyph. (And if it does form a ligature, we currently don't allow breaking between those clusters, as we don't reshape.)

A bit more similar to `\r\n` is the text "é" (as two codepoints U+0065 and U+0301 (e + ◌́)); it's also two clusters, but similarly it's a single grapheme. We don't really represent graphemes at the moment. For the "é"-case, `harfrust` merges it into a single shaping cluster and we then treat it as if it's a ligature. `\r\n` does not shape to a single shaping cluster, though. I have some ideas, as `parley_core` should/will expose graphemes, and `parley` should ideally handle them, but that's a bit more long-term.

For now, I think a quick fix like this in `line_break.rs` is fine.

nicoburns (approved):
Looking at this more closely, I think most of this complexity is required.

mvanhorn:
Sorry for the slow reply, and thanks for pointing me to it - I've read the LLM policy now. You're right that I should have disclosed upfront: this fix was done with AI assistance, and per your policy that belongs in the PR description from the start, not surfaced after the fact. I'll disclose clearly on anything I open here going forward. On @nicoburns' point about combining CRLF into a single cluster rather than handling it at line-break time - that does sound like the cleaner fix and a better fit if the code is moving into Parley core; happy to follow up in that direction if it'd help.

mvanhorn:
Thanks @nicoburns for taking the CRLF hard-break fix.

mvanhorn:
Thanks for the merge, @nicoburns! CRLF as a single hard line break should stop the double-spacing on Windows text.

DJMcNab:
@mvanhorn for awareness, I suspect your agent has gone out of control. Also, please don't send automated comments to our repositories.

mvanhorn:
Hi @DJMcNab — apologies, and thank you for flagging this directly. You're right. I've now read the LLM policy (I should have earlier), and I've turned off the automated commenting that reached your repos. Any future involvement here from me will be manual and minimal. Sorry for the noise, and thanks for merging the CRLF fix regardless.

PR #668, opened 2026-07-01
Remove glifo from `vello_cpu_render` example

nicoburns:
Extracted from https://github.com/linebender/parley/pull/662.

Converts the `vello_cpu_render` example to use `vello_cpu`s built-in text rendering rather than `glifo`

### DIFF
--- a/examples/vello_cpu_render/src/main.rs
+++ b/examples/vello_cpu_render/src/main.rs
@@ -1,25 +1,19 @@
 // Copyright 2024 the Parley Authors
 // SPDX-License-Identifier: Apache-2.0 OR MIT
 
-//! A simple example that lays out some text using Parley, extracts outlines using Skrifa and
-//! then paints those outlines using Vello CPU through Parley Draw.
+//! A simple example that lays out some text using Parley and paints the glyphs using
+//! Vello CPU's built-in glyph rendering.
 
 use std::path::Path;
-use std::sync::Arc;
 use std::time::Instant;
 
-use glifo::renderers::vello_renderer::replay_atlas_commands;
-use glifo::{
-    AtlasConfig, CpuGlyphCaches, GlyphCache, GlyphCacheConfig, GlyphRunBuilder, ImageCache,
-    PendingClearRect,
-};
 use parley::{GlyphRun, Layout, PositionedLayoutItem};
 use parley_examples_common::{
     ColorBrush, ExampleConfig, FrameKind, FrameStats, frame_sequence, output_dir, prepare_layouts,
 };
 use peniko::Color;
 use vello_cpu::{
-    Pixmap, RenderContext,
+    Glyph, Pixmap, RenderContext,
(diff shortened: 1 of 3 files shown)


examples/vello_cpu_render/src/main.rs
```diff
@@ -131,19 +108,7 @@ fn render_frame(
...
-
-                        run_renderer.render_decoration(
-                            x..=x1,
-                            baseline,
-                            offset,
-                            size,
-                            1.0, // buffer around exclusions
-                            renderer,
-                        );
+                        render_decoration(renderer, &decoration.brush, &glyph_run, offset, size);
```
DJMcNab, line 111: Am I right in thinking that this removes ink-skipping underline from the example? I presume that's an accident, as you would have mentioned it in the PR description otherwise.

I don't think it's necessarily a bad change, although I'm surprised it would require enough additional code that we couldn't just keep it.

DJMcNab (commented):
Other than the ink-skipping thing, this looks good to me (although someone probably does need to run this to validate it; I haven't done so yet).

examples/vello_cpu_render/src/main.rs
nicoburns: My understanding is that `vello_cpu` just doesn't expose ink skipping in a sensible way atm. We would have to pull in `glifo`, but the dependency there has been reversed in newer versions of `glifo` (`vello_cpu` now depends on `glifo` rather than the other way around) so we'd have to write our own glue code for rendering `glifo` into `vello_cpu`.

My thought it basically that getting this working (in this repo) is not worth atm while `glifo` is so immature. And we should just consider ink skipping "not ready" until such time as either:

- `vello_cpu` exposes it as an option in it's glyph run API
- `glifo` is in a state where it is ready for public use

DJMcNab: Perhaps I'm misunderstanding, but isnt that https://docs.rs/vello_cpu/latest/vello_cpu/type.GlyphRunBuilder.html#method.render_decoration?

nicoburns: Ohhh, probably. I was not aware of that method!

nicoburns: @DJMcNab I looked into adding the ink-skipping. However, the method was only added to `vello_cpu` in a newer version than we are currently depending on. So I think we have two options:

1. Remove the ink skipping and re-add it once we upgrade `vello_cpu` (this will require changing the tests to also not use `glifo` which I have been leaving until last because I'm expecting lots of snapshot changes).
2. Bundling all of the changes together.

IMO 1. might preferable?

DJMcNab: Sure, that seems fine. So otherwise I'll just run this locally and then it's probably ready to go?

DJMcNab (approved):

PR #669, opened 2026-07-01
Port `vello_editor` example to `vello_cpu` + rename to just `editor`

nicoburns:
Extracted from https://github.com/linebender/parley/pull/662.

Ports the `vello_editor` example to use `vello_cpu` instead of `vello`. This actually makes things quite a bit simpler as we avoid having to deal with all of the WGPU setup code.


examples/editor/src/text.rs
```diff
@@ -426,46 +427,32 @@ impl Editor {
...
+                    .hint(true)
+                    .normalized_coords(run.normalized_coords());
+                if let Some(glyph_xform) = glyph_xform {
+                    builder = builder.glyph_transform(glyph_xform);
+                }
+                builder.fill_glyphs(glyph_run.positioned_glyphs().map(|glyph| Glyph {
+                    id: glyph.id,
+                    x: glyph.x,
+                    y: glyph.y,
+                }));
```
nicoburns, line 460: Switching from `glyph_run.glyphs()` (+manually computing positions) to `glyph_run.positioned_glyphs()` was a drive-by change from the LLM, but a good one so I've kept it.

examples/editor/src/main.rs
```diff
@@ -284,12 +282,18 @@ impl ApplicationHandler<accesskit_winit::Event> for SimpleVelloApp<'_> {
...
             // Resize the surface when the window is resized
             WindowEvent::Resized(size) => {
-                self.context
-                    .resize_surface(&mut render_state.surface, size.width, size.height);
+                if let (Some(width), Some(height)) =
+                    (NonZeroU32::new(size.width), NonZeroU32::new(size.height))
+                {
+                    render_state.surface.resize(width, height).unwrap();
+                }
+                self.handle_resize(to_nonzero_u16(size.width), to_nonzero_u16(size.height));
```
DJMcNab, line 290: Are you sure that this still works when the window is minimised (so the surface doesn't get resized but the render context does...)

examples/editor/src/text.rs
```diff
@@ -46,6 +46,16 @@ fn convert_rect(rect: &parley::BoundingBox) -> peniko::kurbo::Rect {
...
+/// Set the current paint from a [`Brush`].
+///
+/// `vello_cpu`'s `PaintType` uses its own image/gradient types, so we can't pass a
+/// `peniko::Brush` directly. This example only ever uses solid-color brushes.
+fn set_brush(renderer: &mut RenderContext, brush: &Brush) {
+    if let Brush::Solid(color) = brush {
+        renderer.set_paint(*color);
+    }
+}
+
```
DJMcNab: This sparks anti-joy.

examples/editor/src/text.rs
```diff
@@ -426,46 +427,32 @@ impl Editor {
...
-                    scene.stroke(
-                        &Stroke::new(width.into()),
-                        transform,
-                        underline_brush,
-                        None,
-                        &line,
-                    );
+                    renderer.set_stroke(Stroke::new(width.into()));
+                    set_brush(renderer, &style.brush);
+                    renderer.stroke_path(&line.to_path(0.0));
```
DJMcNab, line 436: There are arguments for using ink skipping underline here, but that's non-blocking.

DJMcNab (commented):
I've not ran this yet. On a first pass, the `set_brush` function feels like a massive code smell; it isn't obvious to me why the 'just always pass a color around' and/or pass a Vello CPU paint around were so unviable that this was the best option.

nicoburns:
>  it isn't obvious to me why the 'just always pass a color around' and/or pass a Vello CPU paint around were so unviable that this was the best option.

I had exactly the same thought. And the answer is that `parley::Brush` requires `Default` (which `Color` doesn't implement) and `PartialEq` (which `PaintType` doesn't implement). So it was either this or a newtype until that's fixed.

examples/editor/src/main.rs
nicoburns: Hmm... we probably just need to make sure it doesn't render in that case.

DJMcNab:
I think using a newtype is probably the best option, as it's a similar amount of code, but more accurately reflects the state of the feature.

examples/editor/src/text.rs
nicoburns: We need to wait for a `vello_cpu` for this, then it will be trivial :)

examples/editor/src/main.rs
nicoburns: Testing locally on macOS, minimising seems to work fine.

nicoburns:
Updated to a newtype around `peniko::Color` for the Brush

examples/editor/src/main.rs
DJMcNab: That's not a huge surprise; the thing which needs to be checked is how it handles on Windows. I *believe* that minimising on macOS doesn't change the window size.

It's possible that this has changed on recent winits?
I no longer have a windows machine, so can't test this easily.

DJMcNab (approved):
I guess it's okay for this to land without testing minimising on Windows. Please do just add a comment that we've not tested it and it might not work before landing.

nicoburns:
Video evidence that this does not crash when minimised on Windows is attached. This was generated automatically by an AI running in a VM:

https://github.com/user-attachments/assets/d9e1dd20-10a5-4ee2-9c8c-708f32bc5e61

PR #670, opened 2026-07-02
Add `parley_core` itemizer, refactor `parley` to use it

tomcur:
This adds `Analysis::itemize`, returning an iterator of items to be shaped individually.

The itemizer takes care of splitting text on changing bidi levels and script (and more in the future, like glyph orientation). I've ported the heuristics `parley` currently uses to fill in `script` for characters without any particular script (i.e., `Common`, `Inherit`, or `Unknown`). We may at some point want to perform bracket matching (see e.g. [UAX 24 § 5.2](https://www.unicode.org/reports/tr24/#Common)), which will require some scratch space. At that point, an API like `Itemizer::itemize(&mut self, analysis: &Analysis, split_after: ...) -> Items<...>` might make sense instead.

<details>
<summary>This benches neutral on my machine</summary>

```
$ cargo bench -q --bench=main -- compare ../target/benchmarks/main -t 8.
Default Style - arabic 20 characters               [   8.7 us ...   8.7 us ]      -0.36%
Default Style - latin 20 characters                [   4.2 us ...   4.2 us ]      +0.31%
Default Style - japanese 20 characters             [   8.1 us ...   8.1 us ]      -0.31%
Default Style - arabic 1 paragraph                 [  47.4 us ...  47.5 us ]      +0.23%
Default Style - latin 1 paragraph                  [  16.2 us ...  16.1 us ]      -0.14%
Default Style - japanese 1 paragraph               [  69.8 us ...  69.1 us ]      -1.03%*
Default Style - arabic 4 paragraph                 [ 199.4 us ... 200.8 us ]      +0.72%
Default Style - latin 4 paragraph                  [  61.0 us ...  61.3 us ]      +0.56%
Default Style - japanese 4 paragraph               [  98.5 us ...  97.7 us ]      -0.90%
Styled - arabic 20 characters                      [   9.7 us ...   9.7 us ]      -0.24%
Styled - latin 20 characters                       [   5.4 us ...   5.4 us ]      +0.08%
Styled - japanese 20 characters                    [   8.7 us ...   8.5 us ]      -1.76%*
Styled - arabic 1 paragraph                        [  49.9 us ...  50.0 us ]      +0.23%
Styled - latin 1 paragraph                         [  20.7 us ...  20.7 us ]      -0.00%
Styled - japanese 1 paragraph                      [  75.7 us ...  75.0 us ]      -0.87%
Styled - arabic 4 paragraph                        [ 217.4 us ... 218.8 us ]      +0.64%
Styled - latin 4 paragraph                         [  80.0 us ...  80.8 us ]      +0.94%
Styled - japanese 4 paragraph                      [ 106.9 us ... 105.9 us ]      -0.93%
```
</details>


parley_core/src/itemize.rs
```diff
@@ -0,0 +1,295 @@
...
+        let start_byte_offset = self.char_indices.offset();
+        let mut item_char_len = 0;
+        loop {
+            let byte_offset = self.char_indices.offset();
+
+            let bidi_level = if self.bidi_levels.is_empty() {
+                self.paragraph_bidi_level
+            } else {
+                self.bidi_levels[0]
+            };
```
tomcur, line 148: Possibly we can start thinking about https://github.com/linebender/parley/pull/649#discussion_r3488789219.

parley/src/shape/mod.rs
```diff
@@ -90,140 +88,89 @@ pub(crate) fn shape_text<'a, B: Brush>(
...
+    let split_after = |item_range: parley_core::itemize::TextRange| {
+        // Split at inlines boxes, so each box falls on a shaping boundary.
+        {
+            let mut split = false;
+            // We loop because there may be multiple boxes at this index.
+            while let Some(inline_box) = inline_box_iter.peek() {
+                if inline_box.index < item_range.byte_range.end {
+                    // Inline boxes *before* this indexed are popped (this occurs if the itemizer
+                    // split a run and we were not called). We push all boxes to the layout when we
+                    // loop over the produced items.
```
DJMcNab: What do you think of expanding the parenthetical, e.g. to `this occurs if the itemiser split a run and we were not called, such as at a bidi boundary`

The second sentence of this comment also *probably* could just be removed, but idk.

parley_core/src/itemize.rs
```diff
@@ -0,0 +1,295 @@
...
+    /// Leading characters without a particular script inherit the script of the first character
+    /// *with* a particular script. These heuristics may change in the future. For more, see UAX 24
+    /// (<https://www.unicode.org/reports/tr24/>) Section 5.
+    pub script: Script,
+}
+
+/// The text itemizer, produced by [`Analysis::itemize`].
+pub struct Itemizer<'a, F> {
+    /// The characters and character indices of the input text.
+    char_indices: CharIndices<'a>,
```
DJMcNab: Mega nit: Perhaps something like?
```suggestion
    /// Our underlying iterator over the input text.
    char_indices: CharIndices<'a>,
```

parley_core/src/itemize.rs
DJMcNab: To clarify, this review comment means that this 'consuming' side implements that, but not the creation side?

parley_core/src/itemize.rs
```diff
@@ -0,0 +1,295 @@
...
+    /// Characters in the source text that do not have a particular script (i.e., they are one of
+    /// [`Script::COMMON`], [`Script::UNKNOWN`] or [`Script::INHERITED`]) get their script from
+    /// surrounding context. Currently, these just inherit the script of the preceding characters.
+    /// Leading characters without a particular script inherit the script of the first character
+    /// *with* a particular script. These heuristics may change in the future. For more, see UAX 24
+    /// (<https://www.unicode.org/reports/tr24/>) Section 5.
+    pub script: Script,
+}
+
+/// The text itemizer, produced by [`Analysis::itemize`].
```
DJMcNab: Probably we should say that this is an iterator in this doc comment.

parley_core/src/itemize.rs
```diff
@@ -0,0 +1,295 @@
...
+pub struct Itemizer<'a, F> {
+    /// The characters and character indices of the input text.
+    char_indices: CharIndices<'a>,
+    /// The per-char info, parallel to [`Self::char_indices`].
+    char_info: &'a [CharInfo],
+    /// The per-char bidi level, parallel to [`Self::char_indices`].
+    bidi_levels: &'a [BidiLevel],
```
DJMcNab, line 50: Is there a particular reason for these not to use [`core::slice::Iter`](https://doc.rust-lang.org/stable/std/slice/struct.Iter.html)? I guess you need to peek into both, so it doesn't actually win anything?

parley_core/src/itemize.rs
```diff
@@ -0,0 +1,295 @@
...
+                script = self.current_script;
+            }
+
+            // First iteration of the loop, initialize item properties.
+            if item_char_len == 0 {
+                item_bidi_level = bidi_level;
+                self.current_script = script;
+            }
+
+            if bidi_level != item_bidi_level || real_script(script) && script != self.current_script
```
DJMcNab: Can this `real_script(script)` here ever be false?

parley_core/src/itemize.rs
```diff
@@ -0,0 +1,295 @@
...
+        }
+    }
+}
+
+impl<F: FnMut(TextRange) -> bool> Iterator for Itemizer<'_, F> {
+    type Item = Item;
+
+    fn next(&mut self) -> Option<Self::Item> {
+        if self.char_info.is_empty() {
+            // We're already finished.
```
DJMcNab, line 128: Should we be validating that both bidi_levels and the text iterator are empty here, for added robustness?
Maybe that should just be a debug assertion?

parley_core/src/itemize.rs
```diff
@@ -0,0 +1,295 @@
...
+    use alloc::vec::Vec;
+    use parlance::Script;
+
+    use crate::{Analysis, AnalysisOptions, Analyzer};
+
+    use super::Item;
+
+    const LATN: Script = Script::from_str_unchecked("Latn");
+    const GREK: Script = Script::from_str_unchecked("Grek");
+    const ARAB: Script = Script::from_str_unchecked("Arab");
```
DJMcNab: Have we checked whether:

```
    const LATN: Script = Script::from_bytes(*b"Latn");
```
works? In that scenario, I might be in favour of removing `from_str_unchecked` entirely (obviously not in this PR)

parley/src/shape/mod.rs
```diff
@@ -90,140 +88,89 @@ pub(crate) fn shape_text<'a, B: Brush>(
...
-                break_run = true;
+            if split {
+                return true;
             }
         }
 
-        if level != item.level || script != item.script {
-            break_run = true;
+        let item_style_index = char_style_indices[item_range.char_range.start];
+        let style_index = char_style_indices[item_range.char_range.end];
```
DJMcNab, line 116: I feel like this should be a `char_range.end + 1`, but the fact that none of our tests indicate that suggest I'm mistaken. The name `split_after` certainly indicates that it semantically means "do we need to split after this range"

DJMcNab (approved):
This looks like a good split to me. None of the concerns are blocking, although I think a little bit of clarify improvement might help things here.

parley/src/shape/mod.rs
tomcur: Done, and removed the second line, too. I initially wrote it, as "popped" reads a bit as if they're ignored, but I suppose the code is clear enough that at this point it's just doing splitting (as opposed to pushing to layout).

parley_core/src/itemize.rs
tomcur: That's right. I've not verified whether there are any other places still consuming it, but probably it should be relatively simple to also optimize/symmetrize the RTL case now.

parley_core/src/itemize.rs
tomcur: Yeah, I considered making them iterators, but they always need peeking, so it doesn't really matter which way you go (under the assumption `Peekable<slice::Iter>` is as efficient - I haven't measured!). Instead I wrote the code to reslice to the tail every time.

parley_core/src/itemize.rs
tomcur: Whoops, good catch (went through a few iterations of script handling!). I believe dropping this additional check is also valid if we ever default the text's initial script differently (i.e., if there's no "first real script", we currently set it to `Latin` somewhat arbitrarily, following what `parley` already did, and this still works if we set it to `Common` instead).

parley_core/src/itemize.rs
DJMcNab: Also Peekable of slice is meaningfully less efficient!

parley_core/src/itemize.rs
tomcur: Perfect!

tomcur: (removing `from_str_unchecked` doesn't sound half bad to me!)

parley/src/shape/mod.rs
tomcur: The range given is already decided to be an item. I started writing an explanation here, but wrote them as the following docs on `itemize` instead.

```
/// The predicate is given a range encoding the current item and considers whether to split
/// after that item based on the next character. Iff the predicate returns `true`, the text is
/// split after that item; i.e., given a range of `start..end`, the predicate controls whether
/// that item is now finished, or whether it is extended to include `end`, at which point it'll
/// be `start..end+1` (though note [`TextRange`] encodes both byte and character offsets).
```

tomcur: (I wouldn't be opposed to renaming the predicate to something wordier like `extend_item_to_next_character` and flipping the returned `bool`'s semantics, but merging this for now.)

DJMcNab: Gotcha! It's an exclusive range is the key thing I was missing.
Which of course makes sense, as every other range is exclusive!

Issue #671, opened 2026-07-02
Fontique - Compile error for 32bit


sandreas:
Hey

I'm trying to compile a Slint App for a bit of an exotic architecture (`mipsel-unknown-linux-gnu` - 32 bit LE) with `cross` and I'm getting a compile error for `fontique`. Although Slint depends on `v0.10.0`, this should still apply for `v.0.11.0`. It seems that `self.shared_version` fails to convert and needs specific handling.

From https://docs.rs/atomic-counter/latest/atomic_counter/

> Both implementations are lock-free. Both are a very thin layer over [AtomicUsize](https://doc.rust-lang.org/std/sync/atomic/struct.AtomicUsize.html) which is more powerful but might be harder to use correctly.

For reference:
https://web.mit.edu/rust-lang_v1.25/arch/amd64_ubuntu1404/share/doc/rust/html/std/primitive.usize.html

Let me know if you need more information.

**Build command:**

```sh
cross +nightly build \
    -Z build-std=core,alloc,std,proc_macro,panic_abort \
    --target mipsel-unknown-linux-gnu \
    --release
```

**Error**

```sh
error[E0308]: mismatched types
    --> /cargo/registry/src/index.crates.io-1949cf8c6b5b557f/fontique-0.10.0/src/collection/mod.rs:268:45
     |
 268 |                 version: AtomicCounter::new(self.shared_version),
     |                          ------------------ ^^^^^^^^^^^^^^^^^^^ expected `u32`, found `u64`
     |                          |
     |                          arguments to this function are incorrect
     |
note: associated function defined here
    --> /rust/lib/rustlib/src/rust/library/core/src/sync/atomic.rs:2623:26
     |
2623 |               pub const fn new(v: $int_type) -> Self {
     |                            ^^^
...
3707 | / atomic_int! {
3708 | |     cfg(target_has_atomic = "32"),
3709 | |     cfg(target_has_atomic_primitive_alignment = "32"),
3710 | |     stable(feature = "integer_atomics_stable", since = "1.34.0"),
...    |
3722 | |     u32 AtomicU32
3723 | | }
     | |_- in this macro invocation
     = note: this error originates in the macro `atomic_int` (in Nightly builds, run with -Z macro-backtrace for more info)
help: you can convert a `u64` to a `u32` and panic if the converted value doesn't fit
     |
 268 |                 version: AtomicCounter::new(self.shared_version.try_into().unwrap()),
     |                                                                ++++++++++++++++++++

[...]

PR #672, opened 2026-07-02
Fix Fontique build on 32bit platforms + add CI job to check this

nicoburns:
Fixes https://github.com/linebender/parley/issues/671

Changes:

- Adds a 64/32 bit conditional compile for non-atomic integers, as these are converted into atomic ones in some places.
- Uses this for the `shared_version` field of the collection's `Inner` type.
- Add CI jobs for `powerpc-unknown-linux-gnu`

Generated by Opus 4.8

PR #673, opened 2026-07-02
Fontique: Fixed compile error for 32bit systems (#671)

sandreas:
This PR fixes fontique compile error for 32bit systems - see #671


nicoburns:
Thanks for the PR, but this has already been fixed by https://github.com/linebender/parley/pull/672

PR #674, opened 2026-07-03
Update PackTab to 1.9.0

taj-p:
## Intent

Upgrade to the new PackTab 1.9.0.

---

I previously tested the PR (https://github.com/behdad/packtab.rs/pull/2) in https://github.com/linebender/parley/pull/674/commits/22bbd1dc2855513d3fe226258a6cad45cdf5140a prior to updating this PR to use the new 1.9.0 version. That investigation is below:

Tested against cases our existing `--compression=5`, but also `--compression=0` and `--compression=10`. Our regressions tests pass for all cases - so, it's correct as far as Parley is concerned. 

At compression=10, I noticed a 7 byte reduction relative to current PackTab - I believe that's expected because the proposed PR drops dead padding bytes

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2850,8 +2850,7 @@ dependencies = [
 [[package]]
 name = "packtab"
 version = "1.8.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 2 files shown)


behdad:
Thank you. I'll merge & release packtab.rs.

behdad:
Thanks. packtab.rs 1.9.0 is released.

tomcur (approved):

Issue #619, opened 2026-05-03
Full support for the CSS `white-space` property


nicoburns:
The CSS `white-space` property is defined in sections 3 and 4 of the css-text-4 specification: https://drafts.csswg.org/css-text-4/#white-space-property. It is a very important property for accurately reproducing how browsers lay out text. And can also be used to account for differences between desired behaviour between editable and non-editable text around how whitespace is "hung" at the end of lines.

(the emojis in this issue were not generated by AI)

It is considered a "shorthand" property that decomposes into the following longhands:

#### ✅ `text-wrap-mode`

This property has two variants: `wrap` and `nowrap`. `nowrap` disables wrapping entirely.

```rust
enum TextWrapMode {
     Wrap,
     NoWrap,
}
```

Status: fully implemented.

#### ⚠️  `white-space-collapse`

This property controls:
- ⚠️ How whitespace is collapsed
- ❌ How whitespace at the end of the line is "hung" (or not hung)

```rust
enum WhiteSpaceCollapse {
      Collapse,
      Discard,
      Preserve,
      PreserveBreaks,
      PreserveSpaces,
      BreakSpaces,
}
```
See spec (https://drafts.csswg.org/css-text-4/#white-space-collapsing) for the meanings of each variant.

Status: we currently [implement](https://docs.rs/parley/latest/parley/style/enum.WhiteSpaceCollapse.html) only the `Collapse` and `Preserve` variants in a somewhat hacky way [in the `TreeBuilder`](https://docs.rs/parley/latest/parley/struct.TreeBuilder.html#method.set_white_space_mode), and we also only implement the "white space collapsing" behaviour, and do not implement the different "whitespace hanging" behaviours (implementing the latter will require us to pass `WhiteSpaceCollapse` into layout as a style which we do not currently do).

See also: discussion on https://github.com/linebender/parley/pull/485 which attempted fixes for whitespace hanging behaviour.

#### ⚠️ `white-space-trim`

Controls "whitespace trimming" at the beggining and end of spans.

```rust
enum WhiteSpaceCollapse {
      DiscardBefore,
      DiscardAfter,
      DiscardInner,
}
```

Status: some basic trimming is implemented hackily in the `TreeBuilder`, but there are bugs in the current implementation. And it also doesn't allow full control over the trimming using a style.

conor-93:
[...]

PR #676, opened 2026-07-06
Fix handling of line break overrides before newlines

DJMcNab:
The existing code had a bug where this would have made a soft break happen immediately before a mandatory break, which means that the mandatory break would have been pushed down a line, which is wrong.

Note that the a "visual" reproduction relies on Canva's forked, unconditional implementation of #619 (see https://github.com/conor-93/parley/pull/3), so I haven't added a test here.


nicoburns (approved):
Makes sense. And I'm also inclined to trust your internal testing of this until such time as we have a better setup here.

PR #677, opened 2026-07-06
Deprecate `parlance::Script::from_str_unchecked`

DJMcNab:
As it happens, we can use `from_bytes` exceedingly ergonomically, with more checks from the compiler, so this method has no reason to actually exist!


parlance/src/script.rs
```diff
@@ -33,9 +39,14 @@ impl Script {
...
     ///
-    /// This is intended for `const` construction and does not validate the input.
+    /// This was intended for `const` construction and does not validate the input.
+    /// This method has been deprecated in favour of [`Self::from_bytes`].
     #[must_use]
     #[inline(always)]
+    #[deprecated(
+        since = "0.1.1",
+        note = r#"You should use `from_bytes(*b"....")` instead."#
+    )]
```
nicoburns, line 49: Seeing as the next release will be breaking, is there any reason to deprecate this rather than just removing it?

nicoburns (approved):
Approved. I would consider just removing rather than deprecating, but will leave up to you.

parlance/src/script.rs
DJMcNab: My thinking is that if we do a 0.1.1 for a different reason, this should be deprecated then.

That is, this isn't a good enough reason to release a 0.2.0 by itself.

I didn't realise we already had a breaking change queued; it wasn't in the changelog...

tomcur (approved):
I wonder if we should also just remove `Script::parse`, as it just calls `Script::from_str` (the trait) and therefore we also already have `str::parse`.

DJMcNab:
Taking out of the queue until I understand what is already breaking and has landed.

DJMcNab:
> I wonder if we should also just remove `Script::parse`, as we already have `Script::from_str` (the trait) and therefore also `str::parse`.

Different pr though, right?

tomcur:
> Different pr though, right?

Yes, but this one made me look at the code :)

parlance/src/script.rs
nicoburns: Ohhh. I had not considered that this is in `parlance`. Yeah, best to keep non-breaking. `parlance` being stable is really convenient.

PR #681, opened 2026-07-10
Collapse whitespace in-place

nicoburns:
This is a straightforward local refactor (no intended logic changes) that optimises whitespace collapsing to:

1. Collapse whitespace in-place rather than allocating a new buffer
2. Reuse the `self.uncomitted_text` buffer for every span rather than creating a new one each time


nicoburns:
(on second thoughts, maybe I should not prioritise this to land first)

PR #689, opened 2026-07-14
Fontique: 2-tier font fallback

### LINKED ISSUE #644: Font Fallback not working for at least `汉` on Windows
The font family returned by fontique for the `Hani` script is `Yu Gothic UI` on Windows, which does not have the character `汉`. The character is in `Microsoft YaHei UI`, which is not returned at all.

Reproduction:

```rust
use fontique::{Collection, FallbackKey, Script, SourceCache};

let mut collection = Collection::default();
let mut source_cache = SourceCache::default();
let family_ids: Vec<_> = collection
    .fallback_families(FallbackKey::new(Script::from_bytes(*b"Hani"), None))
    .collect();
let family_names: Vec<_> = family_ids
    .iter()
    .map(|&id| collection.family_name(id).unwrap().to_owned())
    .collect();

println!("fallback families: {family_names:?}");

let family_id = family_ids[0];
let family_name = collection.family_name(family_id).unwrap().to_owned();
let family = collection.family(family_id).unwrap();
let font = family.default_font().unwrap();
let data = font.load(Some(&mut source_cache)).unwrap();
let charmap = font.charmap_index().charmap(data.as_ref()).unwrap();

println!("{family_name}: 汉 -> {:?}", charmap.map('汉'));
```

The problematic line seems to be here, which doesn't contain enough characters for more coverage:
https://github.com/linebender/parley/blob/8294d3e50d48e14dbd8be7982d8d4548996ef561/fontique/src/script.rs#L112

though apparently for DirectWrite only a single font is queried anyway, so it wouldn't be the only thing that needs to be fixed.
https://github.com/linebender/parley/blob/8294d3e50d48e14dbd8be7982d8d4548996ef561/fontique/src/backend/dwrite.rs#L200

### LINKED ISSUE #678: Fontique fails to find font for ❯ char (U+276F) on macOS
This character not rendering has been a long-standing issue in Blitz's TodoMVC example, and font-fallback failure in Fontique was determined to be the root cause by Fable which had this to say:

> The ❯ (U+276F, Dingbats block, script Common) isn't in the Helvetica Neue/Helvetica/Arial stack, and fontique's fallback is script-based (one CoreText-picked family per script sample) with no per-character last-resort search, so no font covering U+276Fis ever tried → .notdef. This is an upstream parley/fontique limitation.

I had GLM-5.2 write a test for it, and it had this to add:

1. The character has Common Unicode script, so parley's itemizer assigns it the Latin script when surrounded by Latin text (Common/Inherited scripts inherit from surrounding context per UAX #24).
2. The default system fallback for Latin (e.g. Helvetica on macOS) doesn't cover U+276F
3. One font present on my system (macOS) that does cover this glyph is "Zapf Dingbats".

The following test reproduces the issue on macOS (not tested on other OSs). It must be run with the `system` feature enabled:

```rust
#[cfg(all(target_os = "macos", feature = "system"))]
#[test]
fn font_fallback_heavy_angle_quote() {

    // Load system fonts so that fontique can find a fallback for U+276F.
    env.font_context().collection.load_system_fonts();

    let text = "\u{276F}";
    let builder = env.ranged_builder(text);
    layout.break_all_lines(None);
    layout.align(Alignment::Start, AlignmentOptions::default());

    let glyph_id = layout
        .lines()
        .flat_map(|line| line.items())
        .find_map(|item| match item {
            PositionedLayoutItem::GlyphRun(glyph_run) => glyph_run.glyphs().next(),
            PositionedLayoutItem::InlineBox(_) => None,
        })
        .map(|glyph| glyph.id)
        .expect("expected a glyph for the ❯ character");

    // The fallback font should have a real glyph for ❯, not a `.notdef`.
    assert_ne!(
        glyph_id, 0,
        "the ❯ character (U+276F) should not resolve to a .notdef glyph when a fallback font is available"
    );
```
}

nicoburns:
- Fixes https://github.com/linebender/parley/issues/678
- Fixes https://github.com/linebender/parley/issues/644

Generated with Fable High.

## Motivation

Font fallback completely fails to find a matching font for some glyphs.

This is because font fallback was keyed on (script, locale), and resolved by asking the OS for a single font family which matches a per-script sample string. If that one font lacked the actual character (for example, CJK extensions, symbols, punctuation shared between scripts, scripts without samples) the Fontique gave up, even if there was actually an installed font that had the required glyph.

## Changes made

Implement ~~3-tier~~ 2-tier fallback:

### Tier A (all backends) - Ordered list of fonts per script (script, locale)

This is the existing fallback mechanism, but modified to work with a list of fonts rather a single font per (script, locale).

* CoreText uses `CTFontCopyDefaultCascadeListForLanguages`
* fontconfig uses `font_sort` with coverage trimming instead of `font_match`
* Android uses all matching locale/script chain entries instead of the first

### Tier B (skipped on Android) - per-character system fallback

* Added `Query::set_fallback_chars()` where Parley passes the specific  characters we are falling back for
*  When script/locale lists (Tier A) lack coverage, the backend is asked for a font covering the actual text (`CTFontCreateForString` / `IDWriteFontFallback::MapCharacters` / `FC_CHARSET` match)
* This is cached per (character-set, locale) in the collection.

### ~~Tier C (all backends) - exhaustive coverage scan~~

~~As a last resort, all families are checked for real cmap coverage (guaranteeing a glyph is found if any font has one)~~

* ~~Fonts are sorted by name, with hidden/.-fonts last, and LastResort excluded)~~
* ~~Positive and negative results are cached, so the scan happens once per unique character set.~~
* ~~This scan is expensive, so there is an opt-out via a new  `exhaustive_fallback: bool` runtime setting on `Collection`~~

(Tier C removed per Chad's feedback)

This PR also:

* Removes the query.rs "Hani fallback hack" (#597), which is now covered by the general per-character fallback mechanism (PingFang/Heiti hvgl hack remains)
* Fixes a latent parley bug where empty layouts shaped their cursor-metrics run with a zero font size.

## Verification Performed

The AI has run temporary integration tests (not included in this PR) on macOS and Ubuntu.

Tested in Blitz on macOS, and it definitely seems to work and find glyphs where `main` doesn't.

### DIFF
--- a/fontique/src/collection/mod.rs
+++ b/fontique/src/collection/mod.rs
@@ -25,6 +25,7 @@ use crate::{AtomicCounter, CounterInt};
 use alloc::{string::String, sync::Arc, vec::Vec};
 use hashbrown::HashMap;
 use read_fonts::types::NameId;
+use smallvec::SmallVec;
 #[cfg(feature = "std")]
 use std::path::Path;
 #[cfg(feature = "std")]
@@ -50,6 +51,19 @@ pub struct CollectionOptions {
     ///
     /// The default value is `true`.
     pub system_fonts: bool,
+
+    /// If true, when a query provides fallback characters (see
+    /// [`Query::set_fallback_chars`]) and neither the requested families nor
+    /// the script fallback families cover them, the query will scan all
+    /// available fonts for one that does, guaranteeing that a glyph is found
+    /// if any font provides one.
+    ///
+    /// The scan requires loading font data to inspect character coverage.
+    /// Results (including negative results) are cached, so this cost is only
+    /// paid once per unique set of fallback characters.
+    ///
+    /// The default value is `true`.
+    pub exhaustive_fallback: bool,
 }
 
 impl Default for CollectionOptions {
@@ -57,6 +71,7 @@ impl Default for CollectionOptions {
         Self {
             shared: false,
             system_fonts: true,
+            exhaustive_fallback: true,
         }
     }
 }
@@ -244,6 +259,8 @@ struct Inner {
     #[allow(unused)]
     shared_version: CounterInt,
     fallback_cache: FallbackCache,
+    text_fallback_cache: TextFallbackCache,
+    exhaustive_fallback: bool,
 }
 
 impl Inner {
@@ -257,6 +274,8 @@ impl Inner {
             shared,
             shared_version: 0,
             fallback_cache: FallbackCache::default(),
+            text_fallback_cache: TextFallbackCache::default(),
+            exhaustive_fallback: options.exhaustive_fallback,
         }
     }
 
@@ -273,6 +292,8 @@ impl Inner {
     /// Load system fonts. If system fonts are already loaded then they will be reloaded.
     pub fn load_system_fonts(&mut self) {
         self.system = Some(System::new());
+        self.fallback_cache.reset();
+        self.text_fallback_cache.clear();
     }
 
     /// Returns an iterator over all available family names in the collection.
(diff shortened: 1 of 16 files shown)


fontique/src/collection/mod.rs
```diff
@@ -643,6 +770,61 @@ impl FallbackCache {
...
+                chars: chars.into(),
+                locale,
+            })
+            .or_default()
+    }
+}
+
+#[derive(Clone, PartialEq, Eq, Hash)]
+struct TextFallbackKey {
+    chars: SmallVec<[char; 8]>,
```
nicoburns, line 815: Is 8 chars a sensible size here?

fontique/src/collection/query.rs
```diff
@@ -101,18 +110,30 @@ impl<'a> Query<'a> {
...
+    /// When set, if none of the families or fallback families provide glyph
+    /// coverage for these characters, the query is extended with fonts
+    /// suggested by the platform for the specific characters and, if
+    /// [`exhaustive_fallback`] is enabled, with any font in the collection
+    /// that covers them.
+    ///
+    /// [`exhaustive_fallback`]: crate::CollectionOptions::exhaustive_fallback
+    pub fn set_fallback_chars(&mut self, chars: impl IntoIterator<Item = char>) {
+        let chars: SmallVec<[char; 8]> = chars.into_iter().collect();
+        if self.state.fallback_chars != chars {
```
nicoburns, line 130: I think we can eliminate this allocation.

dfrg:
I don’t have time for a deep dive on this right now but, at a glance, I’m generally supportive of this with a couple caveats:

1. The char lookup and cache is iffy to me. I’m not opposed in principle but I think we might be able to get by with appending a list of known large coverage fonts to the fallback chain for every query.  
2. I’d like to avoid adding an exhaustive search, even as an option. This is a perf killer and pretty much always produces the wrong result. 

Big +1 to returning multiple fallback fonts from the system. We should probably also extend the script sample array to support multiple samples per script. This would help, for example, with selecting the -ExtB CJK fonts on Windows.

PR #690, opened 2026-07-15
Correctly handle "doesn't compose" in `CharCluster`

tomcur:
We are treating "doesn't compose" as a zero-length composition.

This bug has no observable effect outside the module, as we later do a `map_len.max(1)` and end up treating this as a coverage of `0 / 1`, but that feels like accidental correctness (I hit the bug when I was changing some other code in this module).

### DIFF
--- a/parley_core/src/shape/cluster.rs
+++ b/parley_core/src/shape/cluster.rs
@@ -181,21 +181,14 @@ impl CharCluster {
                 }
 
                 let composer = analysis_data_sources.composing_normalizer();
-                let comp = composer.compose(self.chars[0].ch, self.chars[1].ch);
-                match comp {
-                    None => {
-                        // The characters don't compose.
-                        return None;
-                    }
-                    Some(ch) => {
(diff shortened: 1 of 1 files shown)


parley_core/src/shape/cluster.rs
```diff
@@ -183,7 +183,10 @@ impl CharCluster {
                 let composer = analysis_data_sources.composing_normalizer();
                 let comp = composer.compose(self.chars[0].ch, self.chars[1].ch);
                 match comp {
-                    None => {}
+                    None => {
+                        // The characters don't compose.
+                        return None;
+                    }
```
nicoburns: This could `also` be the `?` operator on `composer.compose`. Would that be better?

tomcur: Sure! Done.

nicoburns (approved):
I don't really understand the context here. But it seems like a small enough fix that if the tests are passing then it's fine to approve.

PR #691, opened 2026-07-15
Let `CharCluster`'s coverage check return `Coverage`s users can compare

tomcur:
The check now returns the fractional coverage, instead of the `Status::{Keep, Discard, Complete}` variants. I.e., the "best ratio" bookkeeping is moved from `parley_core` to `parley`: what exactly to do depending on coverage is more of a user concern.

The old `Status::{Keep, Discard}` are now encoded as a coverage being better or worse (`Ordering::{Greater, Less}`), and `Status::Complete` can be checked by `Coverage::is_complete`.

This returns the integer counts instead of a float, following the suggestion in https://github.com/linebender/parley/pull/688#discussion_r3574749976.

<details>
<summary>This seems to be a small performance win</summary>

```
 % cargo bench --bench main -- compare ../target/benchmarks/main -t 8.
Default Style - arabic 20 characters               [   8.8 us ...   8.7 us ]      -1.20%*
Default Style - latin 20 characters                [   4.2 us ...   4.2 us ]      -1.43%*
Default Style - japanese 20 characters             [   8.1 us ...   8.1 us ]      +0.32%
Default Style - arabic 1 paragraph                 [  47.6 us ...  46.9 us ]      -1.31%*
Default Style - latin 1 paragraph                  [  16.4 us ...  16.1 us ]      -2.09%*
Default Style - japanese 1 paragraph               [  69.4 us ...  69.1 us ]      -0.49%
Default Style - arabic 4 paragraph                 [ 199.5 us ... 197.9 us ]      -0.83%
Default Style - latin 4 paragraph                  [  61.6 us ...  60.4 us ]      -1.89%*
Default Style - japanese 4 paragraph               [  98.3 us ...  97.6 us ]      -0.77%
Styled - arabic 20 characters                      [   9.9 us ...   9.7 us ]      -2.04%*
Styled - latin 20 characters                       [   5.4 us ...   5.3 us ]      -0.72%
Styled - japanese 20 characters                    [   8.7 us ...   8.7 us ]      +0.33%
Styled - arabic 1 paragraph                        [  50.2 us ...  49.4 us ]      -1.64%*
Styled - latin 1 paragraph                         [  20.9 us ...  20.6 us ]      -1.31%*
Styled - japanese 1 paragraph                      [  75.6 us ...  75.3 us ]      -0.38%
Styled - arabic 4 paragraph                        [ 218.2 us ... 216.2 us ]      -0.93%
Styled - latin 4 paragraph                         [  81.3 us ...  80.2 us ]      -1.28%*
Styled - japanese 4 paragraph                      [ 106.7 us ... 106.7 us ]      +0.05%
```
</details>


parley_engine/src/shape/cluster.rs
```diff
@@ -410,27 +459,28 @@ impl Form {
...
-                    mapped += 1;
-                }
-            } else {
-                if covers(c.ch) {
-                    mapped += 1;
-                }
+            if c.contributes_to_shaping && covers(c.ch) {
+                mapped = mapped.saturating_add(1);
             }
         }
```
tomcur, line 475: A bit of a drive-by, but `self.has_contributing` is no longer read here (and I've dropped the field).

In the case of `!self.has_contributing`, i.e., where there are no `contributes_to_shaping` characters, the old code would suddenly start counting all characters, but still did not count them in `map_len` (which was 0, but then clamped to 1), meaning the ratio could become `>= 1.`. That doesn't seem useful.

parley_engine/src/shape/cluster.rs
```diff
@@ -86,15 +87,78 @@ impl Whitespace {
...
-    Complete,
+/// A font's coverage of a character cluster.
+///
+/// This is the fraction `covered / total` of characters covered by a font. Coverages are ordered by
+/// the value of the fraction (with `0 / 0` considered to be completely covered). Use, e.g.,
+/// [`Self::cmp`] to compare coverages.
+#[derive(Copy, Clone, Eq, Debug)]
+pub struct Coverage {
+    /// The number of characters covered by the font.
+    pub covered: u8,
```
DJMcNab, line 101: It's entirely conceivable that this is blatantly obvious in context of the rest of the file, so this is entirely non-blocking:


```suggestion
    /// The number of characters covered by the font.
    // This is only calculated in the context of a single cluster, which is limited to 32 characters.
    pub covered: u8,
```

(As an explanation for why u8 is appropriate here - make actually correct as appropriate)

parley_engine/src/shape/cluster.rs
```diff
@@ -86,15 +87,78 @@ impl Whitespace {
...
+    pub const NONE: Self = Self {
+        covered: 0,
+        total: 1,
+    };
+
+    /// Complete coverage, meaning all characters contributing to shaping are covered.
+    ///
+    /// A coverage compares equal to `COMPLETE` exactly when it is complete, see also
+    /// [`Self::is_complete`].
+    pub const COMPLETE: Self = Self {
```
DJMcNab, line 121: If there's a clever trick incentizing this being 0/0, we should have a comment here. Otherwise, I think I'd slightly prefer this to be e.g. `0xff/0xff` or just 1/1. 0/0 "meaning" 100% is an... idiosyncratic mathematical construction.

parley_engine/src/shape/cluster.rs
```diff
@@ -86,15 +87,78 @@ impl Whitespace {
...
+        // fractions as `Ordering::Equal`.
+        match (self.is_complete(), other.is_complete()) {
+            (true, true) => Ordering::Equal,
+            (true, false) => Ordering::Greater,
+            (false, true) => Ordering::Less,
+            (false, false) => (u16::from(self.covered) * u16::from(other.total))
+                .cmp(&(u16::from(other.covered) * u16::from(self.total))),
+        }
+    }
 }
```
DJMcNab, line 165: I'd like to see an extremely focused unit test which tests this, essentially entirely as a reviewers aid. I trust the maths and the code here (plus everything else still working)

parley_engine/src/shape/cluster.rs
```diff
@@ -86,15 +87,78 @@ impl Whitespace {
...
+    fn partial_cmp(&self, other: &Self) -> Option<Ordering> {
+        Some(self.cmp(other))
+    }
+}
+
+impl Ord for Coverage {
+    /// Coverages are ordered by the value of the fraction `covered / total`.
+    #[inline]
+    fn cmp(&self, other: &Self) -> Ordering {
+        // We check `is_complete` explicitly, such that fractions of `x / 0`, which are considered
```
DJMcNab, line 152: Oh, okay. It's confusing to me that we can end up with `x / 0` in the first place!

parley_engine/src/shape/cluster.rs
```diff
@@ -198,68 +261,60 @@ impl CharCluster {
         }
     }
 
-    /// Decide whether a candidate font is complete, should be kept, or should be discarded.
+    /// Calculate a candidate font's coverage of this character cluster.
```
DJMcNab, line 267: This is a much better doc comment - tells you why the function exists, not just what it "does"

parley_engine/src/shape/cluster.rs
```diff
@@ -198,68 +261,60 @@ impl CharCluster {
...
+    /// characters, or the characters of its normalized forms (see
+    /// <https://www.unicode.org/reports/tr15/>).
+    ///
+    /// Coverages are ordered by how much of the cluster they cover, see [`Coverage`]. Note
+    /// [`Coverage::total`] may be different from the count of [`Self::chars`], and may also differ
+    /// between subsequent calls due to returning the best coverage of either the original
+    /// characters or one of the normalized forms.
+    pub fn calculate_coverage(
         &mut self,
         covers: impl Fn(char) -> bool,
```
DJMcNab, line 280: Obviously this isn't a concern which needs addressing in this pr, but this signature doesn't make me feel reassured that we handle e.g. gsub correctly in font selection.

But this is coming from a place of minimal actual font research, beyond what I've picked up "on-the-job"

parley_engine/src/shape/cluster.rs
DJMcNab: So to close the explanation of why this is equivalent - we now just treat 0/0 as complete.

DJMcNab (commented):
I'm reasonably convinced this gives the same result for all inputs, and I think this refactor is well-justified.

Thanks!

parley_engine/src/shape/cluster.rs
tomcur: Yes, and e.g. `\n` would have a coverage of `0/0` (it has no characters that "contribute to shaping").

parley_engine/src/shape/cluster.rs
tomcur: These are not limited to 32 characters anymore (they were when this still did glyph mapping). Instead we now just saturate the counts. (A grapheme with that many characters probably is some serious Zalgo.)

parley_engine/src/shape/cluster.rs
tomcur: Well, my reasoning is that `0/0` is complete without pretending there's a covering character, i.e., it encodes the "vacuously true" case. I'm not *too* attached to this. `1/1` would be a reasonable alternative, or we could change both `NONE` and `COMPLETE` to `0/255` and `255/255` respectively (to still have some signal they're sentinels).

parley_engine/src/shape/cluster.rs
tomcur: Only because the `Coverage` fields are `pub` (or if we're buggy), but we need the same sort of branching in any case to support `0/0`, as otherwise everything compares equal to `0/0`.

parley_engine/src/shape/cluster.rs
tomcur: Yeah. Relatedly, there's an argument for removing this coverage-based check entirely, and just trying successive fonts until one succeeds to actually shape (retrying to shape the parts that failed to shape). That may also be faster in the likely case the primary font covers everything.

I will not claim complete understanding of this yet, though.

parley_engine/src/shape/cluster.rs
DJMcNab: I'm not too bothered.

DJMcNab (approved):
(I misclicked and meant to approve last time!)

The changes look good.

parley_engine/src/shape/cluster.rs
DJMcNab: Incidentally, trying to shape would also fix #703.

parley_engine/src/shape/cluster.rs
tomcur: I've decided to keep it at your suggestion of `255/255` (and matching `0/255`) for the lower risk of double-takes.

PR #696, opened 2026-07-17
Consolidate line height calculations

tomcur:
This merges the calculations from `finish_line` and `line_height`, moving most of the sizing of lines to their inline content out of the line handling code. Inlining should be able to take care of dead code elimination for the `line_height` method (the benchmarks don't show much movement at least, but I've not looked at the assembly).

The motivation for this is to consolidate line height calculations into one place, mainly as a cleaner starting point for my next PR (#697), which overhauls the calculations themselves.

<details>
<summary>This benches pretty much neutral</summary>

```
$ cargo bench --bench main -- compare ../target/benchmarks/main -t 8

Default Style - arabic 20 characters               [   8.7 us ...   8.8 us ]      +0.45%
Default Style - latin 20 characters                [   4.2 us ...   4.3 us ]      +1.57%*
Default Style - japanese 20 characters             [   8.1 us ...   8.1 us ]      +0.74%
Default Style - arabic 1 paragraph                 [  47.4 us ...  47.4 us ]      -0.03%
Default Style - latin 1 paragraph                  [  16.3 us ...  16.4 us ]      +0.61%
Default Style - japanese 1 paragraph               [  69.3 us ...  69.2 us ]      -0.09%
Default Style - arabic 4 paragraph                 [ 199.6 us ... 202.1 us ]      +1.26%*
Default Style - latin 4 paragraph                  [  61.5 us ...  62.0 us ]      +0.82%
Default Style - japanese 4 paragraph               [  98.4 us ...  97.7 us ]      -0.64%
Styled - arabic 20 characters                      [   9.8 us ...   9.8 us ]      -0.08%
Styled - latin 20 characters                       [   5.4 us ...   5.5 us ]      +1.25%*
Styled - japanese 20 characters                    [   8.7 us ...   8.6 us ]      -0.89%
Styled - arabic 1 paragraph                        [  50.0 us ...  50.1 us ]      +0.15%
Styled - latin 1 paragraph                         [  20.8 us ...  20.9 us ]      +0.56%
Styled - japanese 1 paragraph                      [  75.3 us ...  75.0 us ]      -0.32%
Styled - arabic 4 paragraph                        [ 217.6 us ... 221.7 us ]      +1.88%*
Styled - latin 4 paragraph                         [  81.5 us ...  81.6 us ]      +0.13%
Styled - japanese 4 paragraph                      [ 106.4 us ... 106.2 us ]      -0.22%
```
</details>

### DIFF
--- a/parley/src/layout/line_break.rs
+++ b/parley/src/layout/line_break.rs
@@ -33,12 +33,39 @@ impl LineLayout {
     }
 }
 
-#[derive(Clone)]
+#[derive(Clone, Default)]
 struct LineState {
     x: f32,
     items: Range<usize>,
     clusters: Range<usize>,
     num_spaces: usize,
+    box_metrics: LineBoxMetrics,
+    /// This is set to true if we encounter something on the line (either a glyph or an inline box)
+    /// that is taller than the `line_max_height`. When in this state `break_next` should yield control
+    /// flow to the caller to handle the constraint violation.
+    ///
+    /// This never happens when calling `break_all_lines` as it never sets `line_max_height`, and it defaults to `f32::MAX`.
+    max_height_exceeded: bool,
+
+    /// We lag the text-wrap-mode by one cluster due to line-breaking boundaries only
+    /// being triggered on the cluster after the linebreak.
+    text_wrap_mode: TextWrapMode,
+}
+
+impl LineState {
+    /// Reset the per-line running state in preparation for building a new line.
+    fn reset(&mut self) {
+        self.x = 0.0;
+        self.box_metrics = LineBoxMetrics::default();
+    }
+}
+
+/// The metrics of a line box.
+///
+/// Note a line box is distinct from an *inline* box. A line box is a single line within the layout,
+/// and can hold multiple inline boxes.
+#[derive(Clone, Copy, Debug)]
+struct LineBoxMetrics {
     /// Of the text on the line currently being built, the maximum ascent (distance above the
     /// baseline) of any run seen so far. This is the raw font ascent and does *not* include any
     /// leading.
@@ -63,25 +90,22 @@ struct LineState {
     /// Of the inline boxes on the line currently being built, the maximum descent (distance below
     /// the baseline) of any box seen so far. See [`Self::box_ascent`].
     box_descent: f32,
-    /// This is set to true if we encounter something on the line (either a glyph or an inline box)
-    /// that is taller than the `line_max_height`. When in this state `break_next` should yield control
-    /// flow to the caller to handle the constraint violation.
-    ///
-    /// This never happens when calling `break_all_lines` as it never sets `line_max_height`, and it defaults to `f32::MAX`.
-    max_height_exceeded: bool,
-
-    /// We lag the text-wrap-mode by one cluster due to line-breaking boundaries only
-    /// being triggered on the cluster after the linebreak.
-    text_wrap_mode: TextWrapMode,
 }
 
-impl Default for LineState {
+struct ResolvedLineBoxMetrics {
+    /// Height of the line box.
+    line_height: f32,
+    /// Distance from the top of the line box to the baseline.
+    above: f32,
+    /// Distance the line box content extends above the baseline (with negative leading clamped to 0, i.e., this can be larger than [`Self::above`]).
+    block_above: f32,
+    /// Distance the line box content extends below the baseline (with negative leading clamped to 0).
+    block_below: f32,
+}
+
+impl Default for LineBoxMetrics {
     fn default() -> Self {
         Self {
-            x: 0.0,
-            items: 0..0,
-            clusters: 0..0,
-            num_spaces: 0,
             // Text extents default to zero, so a line with no text contributes no ascent, descent
             // or leading.
             text_ascent: 0.0,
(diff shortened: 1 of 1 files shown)


parley/src/layout/line_break.rs
```diff
@@ -33,12 +33,39 @@ impl LineLayout {
...
+    fn reset(&mut self) {
+        self.x = 0.0;
+        self.box_metrics = LineBoxMetrics::default();
+    }
+}
+
+/// The metrics of a line box.
+///
+/// Note a line box is distinct from an *inline* box. A line box is a single line within the layout,
+/// and can hold multiple inline boxes.
```
DJMcNab: (and text runs?)

parley/src/layout/line_break.rs
```diff
@@ -63,25 +90,22 @@ struct LineState {
...
-    text_wrap_mode: TextWrapMode,
+struct ResolvedLineBoxMetrics {
+    /// Height of the line box.
+    line_height: f32,
+    /// Distance from the top of the line box to the baseline.
+    above: f32,
+    /// Distance the line box content extends above the baseline (with negative leading clamped to 0, i.e., this can be larger than [`Self::above`]).
+    block_above: f32,
+    /// Distance the line box content extends below the baseline (with negative leading clamped to 0).
+    block_below: f32,
```
DJMcNab, line 103: To clarify my understanding, is `above + block_below` equal to line height, or is the idea that you can calculate non-block `below` as `line_height - above`?

parley/src/layout/line_break.rs
DJMcNab, line 137: Hmm, this comment doesn't match my mental model... Guess it's back to the reference material on Monday!

DJMcNab, line 81: I'm presuming this is set only based on the `LineHeight` style?

parley/src/layout/line_break.rs
```diff
@@ -122,34 +142,101 @@ impl LineState {
...
+                self.box_descent,
+            )
+        };
+
+        let leading = self.text_line_height - (ascent + descent);
+
+        // The leading is distributed around the baseline using the text's *intrinsic* line height,
+        // never the (possibly larger) line height caused by inline boxes. This is what keeps
+        // consecutive baselines on a consistent grid: an inline box that fits within the line does
+        // not perturb the text strut, and therefore does not move the baseline.
```
DJMcNab, line 170: We keep using the word strut here. I checked the CSS docs (before Monday(!)), and still think that strut has a very specific definition:

>  it is considered to contain a “strut” (an invisible glyph of zero width) 

Which is just a concept that we don't have in this code. It seems a pretty terrible idea for us to claim this term for a new meaning...

It's fine if the answer is "I'll fix it in the next pr"

parley/src/layout/line_break.rs
```diff
@@ -122,34 +142,101 @@ impl LineState {
...
+        let (leading_above, leading_below) = if quantize {
+            // We mimic Chrome in giving 'below' the larger leading half.
+            // Although the comment in Chromium's NGLineHeightMetrics::AddLeading function
+            // in ng_line_height_metrics.cc claims it's for legacy test compatibility.
+            // So we might want to think about giving 'above' the larger half instead.
+            let above = (leading * 0.5).floor();
+            let below = leading.round() - above;
+            (above, below)
+        } else {
+            (leading * 0.5, leading - leading * 0.5)
```
DJMcNab:
```suggestion
            let above = leading * 0.5;
            // We calculate like this to limit floating point error.
            (above, leading - above)
```

DJMcNab (approved):
This looks good - thanks!

parley/src/layout/line_break.rs
tomcur: Yeah, it comes from `RunMetrics::line_height`, which is calculated from the style.

parley/src/layout/line_break.rs
tomcur: Note these still are the semantics from before this PR (I opened #697 to simplify our model and get us closer to CSS). But yes to your second point: `below` will be `line_height - above`.

Our `block_above + block_below` currently is effectively the maximum of the line height and the height of what I think CSS calls the "content area." The content area fits inline boxes, and the text runs' ascents and descents, ignoring the text runs' leadings.

Chrome uses the max of these two heights for selection geometry, but it's not actually the line's true layout height. I think this is where our "clamping to 0" of negative leading comes from, as ignoring negative leading makes it such that glyphs don't trivially overflow the area.

parley/src/layout/line_break.rs
tomcur: I think this is correct, but it's described somewhat complexly. I believe the semantics are:

- text contributes its `line_height` (from the style) to the line box (for now always aligned by baseline)
- text contributes its `ascent + descent` to the "content area"

Leading is a metric derived from those values.

#697 does a pass over this.

parley/src/layout/line_break.rs
tomcur: Yeah, I agree. This was copied from a different section (so the comment is not really introduced in this PR).

In the follow-up #697 I really didn't think about the word "strut" specifically, but a ctrl+f says that 7 occurrences are removed (and none introduced)!

nicoburns: > an invisible glyph of zero width

The more general meaning I think is a box that is zero-sized in one dimension (and thus invisible), but takes up space in the other dimension (CSS also uses the concept of "strut" for regular boxes with `visibility: collapse` in some parts of the spec).

parley/src/layout/line_break.rs
DJMcNab: Apologies - I believe the main thing I was complaining about was the use of the term "strut", which is just completely wrong here.

PR #697, opened 2026-07-17
Calculate line box sizing following CSS

tomcur:
Following on #696 (which consolidated line height calculations into one place).

This calculates line metrics following the CSS notion of a "line box." The line box is sized to fit its inline content, see [CSS 2.2 § 10.8 about line height](https://www.w3.org/TR/CSS22/visudet.html#line-height).

We have two types of inline content: text runs, whose height is specified to be a specific line height (and note the typographic `ascent + descent` can be larger than the line height, which then is "negative leading"); and inline boxes. This content is aligned together (and as we did previously, we for now just align the content by their baselines). We perform this aligning by keeping track of the line box's "virtual baseline." The line box is then sized to fit its inline content, which is just a simple extent `over` and `under` its virtual baseline.

This drops `LineMetrics::{ascent, descent, leading}`; I'm not sure those concepts really survive at the line level, as they are typographic metrics of a font run, and a line can contain many font runs (or zero font runs!). Before this PR, they were just the running `max`es of the inline contents' ascents, descents and leadings, but that meant these values did not correspond to any one run anymore, and lost their typographic meaning. They were also getting mixed up with details of inline boxes, which don't have a notion of leading at all.

Instead, the existing `LineMetrics::block_{min,max}_coord` are now non-overlapping line boxes (i.e., each line is as tall as necessary to fits the heights of its inline content); the new `LineMetrics::content_block_{min,max}_coord` are the best-effort "content area" of the line (roughly the "paint area"). The content area is identical for inline boxes, but for text runs takes ascent and descent into account instead of line height. That matches both Chrome and Firefox, I believe (and matches out current tests). CSS says:

> CSS 2.2 does not define what the content area of an inline box is (see [10.6.1](https://www.w3.org/TR/CSS22/visudet.html#inline-non-replaced) above) and thus different UAs may draw the backgrounds and borders in different places.

`LineMetrics::baseline` will lie somewhere in between `LineMetrics::block_{min,max}_coord`.

Chrome uses the union of the content block and line block for its per-line selection geometry, which I've updated our `geometry_with` and cursor geometry to. Firefox, on the other hand, uses the per-run `ascent + descent` for its selection geometry.

The updated snapshots show that line boxes are now sized to fit their inline boxes.

The new test exercises that we now handle inline boxes and negative leading correctly. Compare (16px font size, 12px line height, 14px inline box):

| Old Parley | New Parley | Chrome |
| --- | --- | --- |
|lines_negative_leading_inline_box_grows_line_box-0 | lines_negative_leading_inline_box_grows_line_box-0 | ksnip_20260717-195943 |

<details>
<summary>This benches between neutral and slightly negative (these two PRs versus main), but from some local tests I believe we can speed this up if we precompute the rounded metrics and store them in <code>RunMetrics</code>. (Don't want to inflate the diff unnecessarily at the moment.)</summary>

```
$ cargo bench --bench main -- compare ../target/benchmarks/main -t 8

Default Style - arabic 20 characters               [   8.8 us ...   8.9 us ]      +0.84%
Default Style - latin 20 characters                [   4.2 us ...   4.3 us ]      +1.04%*
Default Style - japanese 20 characters             [   8.1 us ...   8.1 us ]      +0.77%
Default Style - arabic 1 paragraph                 [  47.4 us ...  47.8 us ]      +0.69%
Default Style - latin 1 paragraph                  [  16.4 us ...  16.6 us ]      +1.02%*
Default Style - japanese 1 paragraph               [  69.5 us ...  69.8 us ]      +0.49%
Default Style - arabic 4 paragraph                 [ 198.7 us ... 201.2 us ]      +1.24%*
Default Style - latin 4 paragraph                  [  61.5 us ...  62.8 us ]      +1.97%*
Default Style - japanese 4 paragraph               [  98.4 us ...  98.8 us ]      +0.40%
Styled - arabic 20 characters                      [   9.8 us ...   9.9 us ]      +0.79%
Styled - latin 20 characters                       [   5.4 us ...   5.5 us ]      +2.13%*
Styled - japanese 20 characters                    [   8.6 us ...   8.7 us ]      +1.37%*
Styled - arabic 1 paragraph                        [  49.9 us ...  50.4 us ]      +0.89%
Styled - latin 1 paragraph                         [  20.8 us ...  21.1 us ]      +1.12%*
Styled - japanese 1 paragraph                      [  75.2 us ...  76.0 us ]      +1.08%*
Styled - arabic 4 paragraph                        [ 217.9 us ... 219.6 us ]      +0.80%
Styled - latin 4 paragraph                         [  81.5 us ...  82.9 us ]      +1.67%*
Styled - japanese 4 paragraph                      [ 106.6 us ... 107.9 us ]      +1.19%*
```
</details>


parley/src/layout/line.rs
```diff
@@ -144,13 +133,35 @@ pub struct LineMetrics {
...
+    ///
+    /// For horizontal text, this would be the top of the line's content.
+    pub content_block_min_coord: f32,
+    /// Maximum coordinate of the line's content in the direction orthogonal to line direction.
+    ///
+    /// See [`LineMetrics::content_block_min_coord`] for a caveat about glyphs overflowing the
+    /// content area.
+    ///
+    /// For horizontal text, this would be the bottom of the line's content.
+    pub content_block_max_coord: f32,
```
tomcur: Maybe we can find better names for these. Very open to that! I've opted to open the PR keeping `block_{min,max}_coord` around to reduce the breakage, but a name closer to terminology of the CSS specs could be something like `line_box_{min,max}_coord`.

parley/src/editing/cursor.rs
```diff
@@ -446,9 +448,9 @@ fn cursor_rect<B: Brush>(cluster: &Cluster<'_, B>, at_end: bool, size: f32) -> B
     let metrics = line.metrics();
     BoundingBox::new(
         line_x as f64,
-        metrics.block_min_coord as f64,
+        f32::min(metrics.block_min_coord, metrics.content_block_min_coord) as f64,
```
tomcur, line 474: This is the "union" that I mentioned in the PR description, where Chrome calculates the "union" of the line box and content area for selection geometry. If there is large leading, Chrome includes that leading in the selection (i.e, it selects the line box, so there are no gaps in between the selection boxes of lines). If the leading's negative, Chrome does not include the leading in the selection box (so glyphs don't overflow the selection).

Firefox has the somewhat more straightforward behavior of just selecting each run's content area (`ascent + descent`), similar to how both browsers size spans.

nicoburns:
@tomcur btw, I recommend the CSS 2.2 spec. The contents are mostly the same as the original CSS2 spec you linked to, but the formatting is much nicer (IMO)! (EDIT: actually in this case, the content seems a bit different)

https://www.w3.org/TR/CSS22/visudet.html#line-height

parley/src/editing/cursor.rs
nicoburns: Does that mean if there is a large inline box in the line, that it's height will be included in the selection geometry for the whole line?

tomcur:
> @tomcur btw, I recommend the CSS 2.2 spec. The contents are mostly the same as the original CSS2 spec you linked to, but the formatting is much nicer (IMO)! (EDIT: actually in this case, the content seems a bit different)
> 
> https://www.w3.org/TR/CSS22/visudet.html#line-height

Much nicer! Links updated!

parley/src/editing/cursor.rs
tomcur: Yeah, that seems to be what Chrome's doing. 

See: https://developer.mozilla.org/en-US/play?id=5%2BujwfWo3mgB0GcSRGbVorrGvILEzkHi4aGtbFA1WBMxcG2OvWnmpdnxcJuaxSJhBXLuqqsGSkYuxMEC

Firefox selects the per-run content area instead.

(But note that while Chrome grows the selection to include the inline box, it doesn't actually extend the selection horizontally to cover it. I'm not sure what we do currently.)

Note in Firefox that the selection of the white space either side of the inline box (the white space that doesn't have a span style) is 4px high, following the font size set on the `div`!

DJMcNab:
This is something I need to really care about us getting right, so have scheduled in some time on Monday for it.

The "normal" Vs "not normal" difference is my biggest risk point. But also, this kind of needs to have a conversation about what problems/use cases we're trying to solve.

tomcur:
> The "normal" Vs "not normal" difference is my biggest risk point.

I haven't looked at the exact details yet, but perhaps `LineHeight::MetricsRelative(1.)` already encodes "normal", though that variant takes [`skrifa::metrics::Metrics::leading`](https://docs.rs/skrifa/0.44.0/skrifa/metrics/struct.Metrics.html#structfield.leading) into account (which is the font's "recommended additional spacing between lines.") I haven't checked what what browsers do, but the CSS spec doesn't seem to mention that leading for "normal": https://drafts.csswg.org/css-inline/#inline-height.

If we need to ignore it, we could perhaps add another `LineHeight` variant that sets `RunMetrics::line_height` to `metrics.ascent + metrics.descent`. (And note our runs here are "shaped runs," so we know glyphs in the run come from a single font. Though, if we do synthesis like emboldening, should that influence these metrics and require `parley` to split runs further still?)

parley/src/editing/selection.rs
```diff
@@ -531,16 +534,26 @@ impl Selection {
...
-            // end of this line is selected. It's based on the ascent and
-            // descent so it doesn't change with the line height.
+            // end of this line is selected. It's based on the content block so
+            // it doesn't change with the line height.
             //
             // TODO: the width of this whitespace should be the width of a space
             // (U+0020) character.
             let newline_whitespace = if line.break_reason() == BreakReason::Explicit {
-                (metrics.ascent as f64 + metrics.descent as f64) * NEWLINE_WHITESPACE_WIDTH_RATIO
+                // The newline's metrics are those of the last run.
```
DJMcNab: I'd expect it to be the font for the newline? I guess it never gets an actual selected font?

parley/src/layout/line.rs
DJMcNab: Whilst we're here, it might be worth expanding what this is relative to (e.g. global, local...)

parley/src/layout/line_break.rs
```diff
@@ -64,181 +64,90 @@ impl LineState {
...
-    /// leading.
-    text_descent: f32,
-    /// Of the text on the line currently being built, the maximum intrinsic line height (i.e.
-    /// including the full leading) of any run seen so far. This may be *smaller* than
-    /// `text_ascent + text_descent` when the line height is smaller than the font metrics (negative
-    /// leading), in which case lines are allowed to overlap.
-    text_line_height: f32,
-    /// Of the inline boxes on the line currently being built, the maximum ascent (distance above
-    /// the baseline) of any box seen so far.
+    /// The extents from the line box's baseline.
```
DJMcNab, line 75:
```suggestion
    /// The vertical extents from the line box's baseline.
```
might be better.

parley/src/layout/line_break.rs
```diff
@@ -64,181 +64,90 @@ impl LineState {
...
-            text_line_height: 0.0,
-            // Box extents default to negative infinity so that a line with no inline boxes never
-            // grows the line via the box-overflow terms in `line_height`.
-            box_ascent: f32::NEG_INFINITY,
-            box_descent: f32::NEG_INFINITY,
+            // NOTE: these should be `f32::NEG_INFINITY`, but that's currently causing
+            // `tests::test_builders::builders_empty` to fail with a `NaN`.
+            //
+            // And even more ideally, a line's initial extents should be sourced from the primary
+            // font.
```
DJMcNab, line 105: *this* is the strut of the root inline box, right?

parley/src/layout/line_break.rs
```diff
@@ -1301,29 +1250,26 @@ impl<'a, B: Brush> BreakLines<'a, B> {
...
+            self.state.line_y.round() as f32
+        } else {
+            self.state.line_y as f32
+        };
+        line.metrics.baseline = top + line_box_extents.over;
+        line.metrics.block_min_coord = top;
+        line.metrics.block_max_coord = if quantize {
+            // TODO: perhaps this should be something like the following, to ensure quantized line
+            // boxes tile exactly without gaps or overlap. However, that would cause some
+            // line height asserts to fail in the tests.
```
DJMcNab, line 1268: Do we know what Chromium does here?

Is the thinking that we'll change the tests in a follow-up, but this is kept this way to de-risk.

DJMcNab (approved):
This seems like a much more reasonable platform to work from in getting this fully correct (with first available font and friends!). Thanks!

parley/src/editing/selection.rs
tomcur: I believe if we have a `BreakReason::Explicit`, there was something like a (CR)LF and that'll be part of a run. We mark a newline character as not contributing to shaping, so I believe it should just always be part of a run of the primary font.

Of course this is still hacky in that we don't take that font's actual space character width.

Just in case, I did point an LLM at this, and it noted we should take the last *logical* run of line (and not visual run as we do here), so I fixed that.

parley/src/layout/line_break.rs
tomcur: Right! That matches my understanding.

parley/src/layout/line.rs
tomcur: I've done a quick pass, but I'm sure it can be improved further!

parley/src/layout/line_break.rs
tomcur: We'd have to check. I think it won't cause snapshot changes, but I believe the failures are about measured line box height vs specified and reported line height.

PR #707, opened 2026-07-23
Keep justification out of `ShapedText`

waywardmonkeys:
## Summary

Keep justification state in `parley` rather than mutating the font-derived data in `parley_engine::ShapedText`.

Justification now stores per-cluster advance adjustments in `LayoutData`. Re-line-breaking or re-aligning discards that derived state directly instead of running the justification algorithm again with negative adjustments.

This removes:

- `unjustify`
- `is_aligned_justified`
- The `UNDO_JUSTIFICATION` const-generic path
- Justification’s use of `ShapedText::clusters_mut()`

## Motivation

`ShapedText` is the output of shaping, but justification previously modified `ClusterData::advance` in place. To reuse a layout, Parley had to remember that this happened and later “unjustify” it by repeating the alignment calculation in reverse.

Besides crossing the shaping/layout ownership boundary, that left multiple answers to how wide justified text was:

- `Cluster::advance()` included justification.
- Cached run and line advances did not.
- Rendering depended on whether a glyph was stored inline.

With this change, shaping output remains unchanged throughout justification, repeated alignment, and re-line-breaking.

## Implementation

`LayoutData` now owns a lazily allocated vector of per-cluster justification adjustments. It is empty for layouts where no spaces have been expanded.

The final adjusted advance is used consistently by:

- `Cluster::advance()`
- Emitted glyph advances
- Cached line-item/run advances
- `LineMetrics::advance`
- Hit testing, cursor placement, and accessibility through their existing cluster/run accessors

Line breaking continues to read natural shaped advances directly. Each line retains its natural advance so clearing justification restores cached geometry without performing an inverse calculation.

For non-inline clusters, the adjustment is added to the copied final glyph in the cluster. This keeps positioned glyphs consistent with the cluster, run, and line geometry without modifying `ShapedText`.

## Tests

Added regressions covering:

- `ShapedText` remaining unchanged after justification
- Repeated justification
- Re-line-breaking a justified layout
- Agreement between line, run, cluster, and emitted glyph advances
- Latin and Arabic text
- Restoring the copied metrics of an empty final line after a trailing newline

Existing realignment, bidi, text-indent, word-spacing, hit-testing, and rendering tests continue to pass.

## Performance

Added permanent Tango benchmarks for full justified layout and repeated realignment of an already line-broken paragraph.

Paired comparison:

| Benchmark | Before | After | Change |
|---|---:|---:|---:|
| Realign — Arabic | 408.4 ns | 214.1 ns | −47.58% |
| Realign — Latin | 320.2 ns | 184.8 ns | −42.30% |
| Justified — Arabic, 1 paragraph | 44.3 µs | 44.3 µs | +0.06% |
| Justified — Latin, 1 paragraph | 15.1 µs | 15.1 µs | +0.23% |
| Justified — Arabic, 4 paragraphs | 178.9 µs | 179.2 µs | +0.16% |
| Justified — Latin, 4 paragraphs | 58.6 µs | 58.6 µs | −0.06% |

Repeated runs consistently showed a roughly 42–48% improvement for realignment. Full-layout results remained effectively neutral; small shifts among the shortest default and styled cases moved between runs and did not indicate a path-specific regression.

## Next steps

This does not yet make `ShapedText` immutable from Parley’s perspective. Letter and word spacing still mutate clusters and glyphs through `clusters_and_glyphs_mut()`.

A follow-up should:

1. Resolve the existing letter-spacing/ligature semantics, including whether letter spacing disables ligatures.
2. Move letter and word spacing into layout-owned per-cluster state.
3. Refactor the test canonicalizer so it no longer needs `clusters_mut()`.
4. Remove the hidden `clusters_mut()`, `glyphs_mut()`, and `clusters_and_glyphs_mut()` escape hatches from `parley_engine::ShapedText`.

That would complete the ownership boundary established here: `parley_engine` owns immutable shaping output, while `parley` owns all layout-time advance adjustments.

### DIFF
--- a/parley/src/layout/alignment.rs
+++ b/parley/src/layout/alignment.rs
@@ -5,8 +5,6 @@ use super::{BreakReason, data::LineItemData};
 use crate::data::LayoutData;
 use crate::style::Brush;
 
-use parley_engine::shape::ClusterData;
-
 /// Alignment of a layout.
 #[derive(Copy, Clone, Default, PartialEq, Eq, Debug)]
 #[repr(u8)]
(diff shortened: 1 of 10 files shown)


waywardmonkeys:
LLM assisted, GPT 5.6, Sol, xhigh. Human reviewed and audited as usual.

nicoburns (approved):
Makes sense to me. I guess the trade-off here is extra memory usage. But I reckon that's probably worth it.

waywardmonkeys:
The real pay off is when the other stuff that is violating layering is also removed as laid out above as next steps.

tomcur:
Because of pre-existing bugs ([#parley > use grapheme clusters over shaping clusters @ 💬](https://xi.zulipchat.com/#narrow/channel/205635-parley/topic/use.20grapheme.20clusters.20over.20shaping.20clusters/near/611987773)), the existing justification is likely not entirely correct. I'm currently working to try to solve those grapheme / cluster details, and it's not quite clear yet how this PR would interact with that. Bruce and I have spoken out-of-band, and decided to wait with this.

waywardmonkeys:
This causes problems for Tom, so I will not be pursuing this (and I no longer use Parley).

PR #710, opened 2026-07-24
Add `parlance::BidiLevel`

tomcur:
The motivation is to have a `repr(transparent)` wrapper for bidi levels to hold documentation and useful methods (e.g., letting users call `level.is_rtl()` instead of manually writing `!level.is_multiple_of(2)`, or having our APIs vary between `level() -> u8` and `is_rtl() -> bool` methods), while allowing bytemucking between `BidiLevel` <=> `u8` for interop.

`parley_engine` now returns `BidiLevel`, but I've kept `parley`'s API unchanged.

### DIFF
--- a/parlance/src/bidi.rs
+++ b/parlance/src/bidi.rs
@@ -38,3 +38,50 @@ pub enum BidiOverride {
     /// Force right-to-left.
     Rtl,
 }
+
+/// Bidirectional text embedding level.
+///
+/// These are numbers indicating how deeply bidirectional embeddings are nested in the text, and the
+/// default direction of text on that level. Even levels are left-to-right, odd levels are
+/// right-to-left. Normally, the minimum level is 0 (left-to-right), and the maximum level,
+/// according to [UAX #9 § 3.1.1 BD2][uax-bd2], is 125.
+///
+/// See [UAX #9 § 3.1][uax-definitions] for more information.
+///
+/// [uax-definitions]: https://unicode.org/reports/tr9/#Definitions
+/// [uax-bd2]: https://unicode.org/reports/tr9/#BD2
+///
+// NOTICE: If the representation changes, be sure to check the `bytemuck` marker trait
+// implementations.
+//
+// TODO: it would be quite nice for this to implement
+// <https://doc.rust-lang.org/stable/core/iter/trait.Step.html>, once stabilized.
+#[derive(Clone, Copy, Default, PartialEq, Eq, PartialOrd, Ord, Hash, Debug)]
+#[repr(transparent)]
+pub struct BidiLevel(u8);
+
+impl BidiLevel {
+    /// Construct a new bidi level.
+    #[inline(always)]
+    pub const fn new(level: u8) -> Self {
+        Self(level)
+    }
+
+    /// Get the numeric bidi level.
+    #[inline(always)]
+    pub const fn to_u8(self) -> u8 {
+        self.0
+    }
+
+    /// Whether this level is left-to-right.
+    #[inline(always)]
+    pub const fn is_ltr(self) -> bool {
+        self.0.is_multiple_of(2)
+    }
+
+    /// Whether this level is right-to-left.
+    #[inline(always)]
+    pub const fn is_rtl(self) -> bool {
+        !self.is_ltr()
+    }
+}
--- a/parlance/src/bidi.rs
+++ b/parlance/src/bidi.rs
@@ -92,4 +92,22 @@ impl BidiLevel {
     pub const fn is_rtl(self) -> bool {
         !self.is_ltr()
     }
+
+    /// Get the next odd bidi level.
+    ///
+    /// When the return value overflows (`self.to_u8() >= 255`) this panics in debug mode. The
+    /// return value wraps in release mode.
+    #[inline(always)]
+    pub const fn next_odd(self) -> BidiLevel {
+        BidiLevel::new((self.to_u8() + 1) | 1)
+    }
+
+    /// Get the next even bidi level.
+    ///
+    /// When the return value overflows (`self.to_u8() >= 254`) this panics in debug mode. The
+    /// return value wraps in release mode.
+    #[inline(always)]
+    pub const fn next_even(self) -> BidiLevel {
+        BidiLevel::new((self.to_u8() + 2) & !1)
+    }
 }
--- a/parlance/src/bidi.rs
+++ b/parlance/src/bidi.rs
@@ -95,8 +95,8 @@ impl BidiLevel {
 
     /// Get the next odd bidi level.
     ///
-    /// When the return value overflows (`self.to_u8() >= 255`) this panics in debug mode. The
-    /// return value wraps in release mode.
+    /// When the return value overflows (`self.to_u8() >= 255`) this panics when overflow checks are
+    /// enabled. Otherwise, the return value wraps.
     #[inline(always)]
     pub const fn next_odd(self) -> Self {
         BidiLevel::new((self.to_u8() + 1) | 1)
@@ -104,8 +104,8 @@ impl BidiLevel {
 
     /// Get the next even bidi level.
     ///
-    /// When the return value overflows (`self.to_u8() >= 254`) this panics in debug mode. The
-    /// return value wraps in release mode.
+    /// When the return value overflows (`self.to_u8() >= 254`) this panics when overflow checks are
+    /// enabled. Otherwise, the return value wraps.
     #[inline(always)]
     pub const fn next_even(self) -> Self {
         BidiLevel::new((self.to_u8() + 2) & !1)
--- a/parlance/src/bidi.rs
+++ b/parlance/src/bidi.rs
@@ -98,7 +98,7 @@ impl BidiLevel {
     /// When the return value overflows (`self.to_u8() >= 255`) this panics in debug mode. The
     /// return value wraps in release mode.
     #[inline(always)]
-    pub const fn next_odd(self) -> BidiLevel {
+    pub const fn next_odd(self) -> Self {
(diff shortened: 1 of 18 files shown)


parley_engine/src/analysis.rs
```diff
@@ -70,21 +70,15 @@ impl Analysis {
...
-    pub fn paragraph_level(&self) -> u8 {
+    pub fn paragraph_level(&self) -> BidiLevel {
         self.paragraph_level
     }
-
-    /// Whether the paragraph's resolved base direction is right-to-left.
-    #[inline(always)]
-    pub fn is_rtl(&self) -> bool {
-        !self.paragraph_level.is_multiple_of(2)
-    }
```
tomcur, line 87: Removing `Analysis::is_rtl` in favor of `Analysis::paragraph_level().is_rtl()`, as per https://github.com/linebender/parley/pull/708#discussion_r3645691315.

In that same thread, Bruce pointed out we have some inconsistency in the code base, sometimes calling this "base level", sometimes "paragraph level". [The spec](https://unicode.org/reports/tr9/) uses both words, but uses "paragraph level" a bit more. The code uses "base level" more. In any case, we probably should still rename this, either to something like `base_bidi_level` or `paragraph_bidi_level`.

tomcur:
We could similarly implement `bytemuck` traits for `NormalizedCoord` (see https://github.com/linebender/parley/pull/679#discussion_r3567241390 and https://github.com/linebender/parley/pull/679#discussion_r3606425309).

parley_engine/src/bidi.rs
```diff
@@ -902,3 +902,13 @@ impl BracketStack {
...
+
+#[inline(always)]
+const fn next_odd(level: BidiLevel) -> BidiLevel {
+    BidiLevel::new((level.to_u8() + 1) | 1)
+}
+
+#[inline(always)]
+const fn next_even(level: BidiLevel) -> BidiLevel {
+    BidiLevel::new((level.to_u8() + 2) & !1)
+}
```
nicoburns: These could optionally be methods on the type

nicoburns (approved):
Very nice. Good use of parlance.

parlance/src/bidi.rs
```diff
@@ -38,3 +38,50 @@ pub enum BidiOverride {
     /// Force right-to-left.
     Rtl,
 }
+
+/// Bidirectional text embedding level.
+///
+/// These are numbers indicating how deeply bidirectional embeddings are nested in the text, and the
+/// default direction of text on that level. Even levels are left-to-right, odd levels are
+/// right-to-left. Normally, the minimum level is 0 (left-to-right), and the maximum level,
+/// according to [UAX #9 § 3.1.1 BD2][uax-bd2], is 125.
```
DJMcNab, line 47: Incidentally, this being 125 does give us a bit to muck about with. I don't know of any use for that (maybe for whether rule L1 would apply to this?).

parlance/src/bidi.rs
```diff
@@ -38,3 +38,50 @@ pub enum BidiOverride {
...
+// NOTICE: If the representation changes, be sure to check the `bytemuck` marker trait
+// implementations.
+//
+// TODO: it would be quite nice for this to implement
+// <https://doc.rust-lang.org/stable/core/iter/trait.Step.html>, once stabilized.
+#[derive(Clone, Copy, Default, PartialEq, Eq, PartialOrd, Ord, Hash, Debug)]
+#[repr(transparent)]
+pub struct BidiLevel(u8);
+
+impl BidiLevel {
```
DJMcNab, line 63: Does it make sense for max level to be stored here, either a u8 or BidiLevel constant?

parlance/src/bidi.rs
```diff
@@ -38,3 +38,50 @@ pub enum BidiOverride {
...
+    /// Construct a new bidi level.
```
DJMcNab, line 72: I'd claim that we should debug assert that this is less than or equal to 125, but not blockingly so.

parlance/src/impl_bytemuck.rs
```diff
@@ -39,6 +39,14 @@ unsafe impl Contiguous for GenericFamily {
     const MAX_VALUE: u8 = GenericFamily::MAX_VALUE;
 }
 
+// Safety: The struct is `repr(transparent)`, wrapping a `u8`.
+//
+// While generally BidiLevels have a maximum of 125, no value is unsound.
+unsafe impl Pod for BidiLevel {}
+
+// Safety: The struct is `repr(transparent)`, wrapping a `u8`.
+unsafe impl Zeroable for BidiLevel {}
```
DJMcNab, line 48: We *probably* should have the usual safety tests for these. But also this is so simple it seems hard to imagine it going wrong! I'm not even sure what the tests would look like? Maybe even just that the size is 1 to force this to be revisited?

parley/src/layout/alignment.rs
```diff
@@ -97,7 +97,7 @@ fn align_impl<B: Brush, const UNDO_JUSTIFICATION: bool>(
     options: AlignmentOptions,
 ) {
     // Whether the text base direction is right-to-left.
-    let is_rtl = layout.base_level & 1 == 1;
+    let is_rtl = layout.base_level.is_rtl();
```
DJMcNab: This comment!

parley/src/layout/run.rs
```diff
@@ -110,7 +110,7 @@ impl<'a, B: Brush> Run<'a, B> {
 
     /// Returns `true` if the run has right-to-left directionality.
     pub fn is_rtl(&self) -> bool {
-        self.shaped.bidi_level & 1 != 0
+        self.shaped.bidi_level.is_rtl()
     }
```
DJMcNab, line 114: I guess nominally it might be better to just return the bidi level here, but that's not done here to avoid breaking changes?

parley/src/tests/test_analysis.rs
```diff
@@ -26,8 +27,11 @@ impl TestContext {
         self
     }
 
-    fn expect_bidi_embed_level_list(self, expected: Vec<u8>) -> Self {
+    fn expect_bidi_embed_level_list(self, expected: &[u8]) -> Self {
         let actual = self.layout_context.analysis.bidi_levels();
+        // For ergonomics, we allow tests to pass in levels like `[0, 1, 2, 2, 1, 1, 1]`, and
+        // convert here. (Could also bytemuck... but that'd require pulling in the dependency here).
+        let expected = Vec::from_iter(expected.iter().map(|level| BidiLevel::new(*level)));
```
DJMcNab: I'd lean towards bytemucking here. Although my inclination would actually be to go the other way; comparing two printed arrays of `BidiLevel(1), BidiLevel(2)` etc. seems miserable.

parley/src/tests/test_analysis.rs
```diff
@@ -26,8 +27,11 @@ impl TestContext {
         self
     }
 
-    fn expect_bidi_embed_level_list(self, expected: Vec<u8>) -> Self {
+    fn expect_bidi_embed_level_list(self, expected: &[u8]) -> Self {
```
DJMcNab, line 29: _🔥_

parley_engine/src/bidi.rs
```diff
@@ -92,11 +89,11 @@ impl BidiResolver {
             len += 1;
         }
         self.base_level = match base_direction {
-            BaseDirection::Auto => Self::default_level(&self.initial_types),
-            BaseDirection::Ltr => 0,
-            BaseDirection::Rtl => 1,
+            BaseDirection::Auto => BidiLevel::new(Self::default_level(&self.initial_types)),
```
DJMcNab: I'd probably make default_level return a BidiLevel?

parley_engine/src/bidi.rs
DJMcNab: The lack of capping to 125 here would be my main hesitance, but I think on balance I agree (also with it being optional/up to your discretion)

DJMcNab (approved):
Nothing blocking, some code style nits.

I'm really happy to see this type properly documented. I did try and start on that task, but it got too big for what I had time to do.

This probably should have changelog entries (what is parley_engine's changelog story?)

parlance/src/bidi.rs
tomcur: That sounds sensible. I've added `BidiLevel::MAX`, but perhaps you have feedback on the naming.

In particular, it's not the greatest value `BidiLevel` itself can represent.

parlance/src/bidi.rs
tomcur: Perhaps that makes sense. This ties back into https://github.com/linebender/parley/pull/710#discussion_r3661314159.

Deferring this for now.

parley/src/layout/alignment.rs
tomcur: Removed it :)

parley/src/layout/run.rs
tomcur: Yeah, I intentionally did not change the `parley` API here.

That said, we probably should break it. (But not in this PR.)

parley/src/tests/test_analysis.rs
tomcur: Done!

parlance/src/impl_bytemuck.rs
tomcur: I've added the size test. I'm also not sure whether we can do better.

parley_engine/src/bidi.rs
tomcur: I've added the methods, with some documentation about wrapping.

tomcur:
> This probably should have changelog entries (what is parley_engine's changelog story?)

I've added a changelog entry for Parlance. I think we should write something in the changelog for `parley_engine` when we do the initial release (more or less describing the state it's in at that point).

PR #712, opened 2026-07-25
Retain Unicode joining facts through shaping

waywardmonkeys:
Layout policy needs to distinguish scripts whose letters participate in cursive joining without treating every `Join_Causing` character as evidence of a cursive script. In particular, U+200D ZERO WIDTH JOINER has `Joining_Type=Join_Causing` and can occur in Latin or emoji text.

Pack `Joining_Type` into `parley_data::Properties`, and generate a sorted lookup of ISO 15924 tags for scripts containing `Left_Joining`, `Right_Joining`, or `Dual_Joining` characters. Excluding `Join_Causing` and `Transparent` keeps `Zyyy` from being classified solely because it owns ZWJ and inherited marks.

Expose `script_has_joining_characters` in `parley_engine` for its `parlance::Script` vocabulary, and retain the resolved `script` in each `ShapedRun`. `Item` and `ShapedRun` derive their convenience accessor from that script instead of caching duplicate state, so public record construction remains straightforward and there is no script/classification invariant to maintain.

The lookup deliberately reports a Unicode fact. It does not inspect a particular string or decide letter-spacing policy.


waywardmonkeys:
This was done with assistance by GPT 5.6, Sol, xhigh. I've been through it though and made alterations.

waywardmonkeys:
This is needed for Underwood. I also have a patch that uses this within `parley` as Parley hasn't handled letter spacing vs Arabic correctly in the past. I'll submit that as a separate PR.

PR #713, opened 2026-07-25
Preserve cursive shaping under letter spacing

waywardmonkeys:
CSS Text requires user agents that cannot perform cursive elongation to avoid inserting gaps between a cursive script’s typographic letter units. Such text is effectively treated as one tracking unit per word, while spacing can still apply to non-cursive content such as spaces. Parley instead added `letter_spacing` to every shaped cluster, which broke Arabic and other cursive connections.

Use the resolved-script fact retained by `ShapedRun` to suppress letter spacing for non-space clusters in scripts with joining characters. Spaces still receive both `letter_spacing` and `word_spacing`, and non-cursive runs remain tracked normally. Because the decision uses the resolved script rather than the presence of a `Join_Causing` character, a Latin run containing U+200D ZERO WIDTH JOINER remains trackable.

CSS Text and CSS Fonts also specify that non-default letter spacing disables optional ligatures, while low-level `font-feature-settings` take precedence. Add default `liga=0` and `clig=0` shaping features only when tracking is actually applied and the author did not supply those tags. Required and cursive shaping features remain untouched.

References:

- CSS Text Level 3, Tracking and Cursive Scripts: https://www.w3.org/TR/css-text-3/#cursive-spacing
- CSS Fonts Level 4, Feature and variation precedence: https://www.w3.org/TR/css-fonts-4/#feature-variation-precedence

Add coverage for default and explicitly authored ligatures, cursive words and spaces, and Latin text containing ZWJ. Extend `parley_bench` with nonzero-letter-spacing cases so this path remains measurable.

Final Tango comparisons against the parent commit were flat within run-to-run noise for tracked Arabic (+0.03% and -0.64%) and Latin (+0.44% and -0.97%). Tracked Japanese measured +2.79% and +4.19% because shaping now processes the two required optional-ligature overrides; the zero-spacing suite showed no stable regression across repeated runs.


waywardmonkeys:
This builds on top of #712. It was done with GPT 5.6, Sol, xhigh.

It exists because I figured there should be a use case in tree for what I needed from #712. My code does not rely on this at all as it isn't using `parley` itself.

waywardmonkeys:
This causes problems for Tom, so I will not pursue this.

PR #716, opened 2026-07-28
Move AccessKit integration to examples. Upgrade AccessKit to 0.25.

DataTriny:
As discussed at RustWeek 2026, the current API of the AccessKit integration is not a great fit for many consumers.

The conclusion we reached was that removing the integration entirely would allow users to design their own according to their needs. The existing integration can be moved to the vello_editor example to serve as a reference implementation or a starting point.

This require making some items public though.

The changes here are mostly mechanical. I am guessing that most of the changes made to parley's API will be controversial, so this pull request should be considered as a way to open the discussion, not necessarily as something to merge as is.

See https://github.com/slint-ui/slint/issues/2895 for my current motivation.


DJMcNab (commented):
We spoke about this in the most recent text office hours. We're definitely in favour, but @nicoburns wants to rebase this, as he's the one who has changed the editor example quite significantly (if I remember correctly - the transcript hasn't been released).

It also needs just a little bit of thought from a reviewer about the new public API being exposed.

DataTriny:
@nicoburns We can update accesskit to 0.25 while at it. Would you want me to do it? There shouldn't be any API change  relevant to parley.

nicoburns:
> We can update accesskit to 0.25 while at it. There shouldn't be any API change relevant to parley.

Yes - done!

nicoburns (approved):

PR #717, opened 2026-07-28
Don't panic when fontconfig exposes no fonts

ogoffart:
While populating the generic family map, `SystemFonts::new` unwrapped the result of `font_sort`. On a host whose fontconfig configuration exposes an empty font set (common in minimal containers and embedded Linux images that ship fontconfig but no font files) `font_sort` returns `NoMatch` for the first generic family, so the unwrap panicked inside `Collection::new`, before any fallback font could be registered.

Treat an error from `font_sort` as "this generic family has no members" and skip it. A collection built on a font-less system then exposes no system fonts, which is the correct result and lets the caller register its own fonts afterwards.


tomcur (approved):
Good catch. The fix looks reasonable to me; thank you!

PR #720, opened 2026-07-30
Ignore empty word break ranges, clarify ranges must fall on character boundaries

tomcur:
Empty ranges must be ignored, as otherwise the range the iterator is pointing can get out of sync with the character.

### DIFF
--- a/parley_engine/src/analysis.rs
+++ b/parley_engine/src/analysis.rs
@@ -337,6 +337,16 @@ pub(crate) fn analyze_text(
             if self.cursor >= self.text_len {
                 return None;
             }
+
+            // Ignore empty ranges.
+            while self
+                .word_break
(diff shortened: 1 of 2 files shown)


nicoburns (commented):
Would it be better to store this as a flat Vec of indices (`usize` rather than `Range<usize>`)?

tomcur:
> Would it be better to store this as a flat Vec of indices (`usize` rather than `Range<usize>`)?

I think probably not, because of gaps using `WordBreak::Normal`. (If we didn't allow gaps, we'd either have to special-case `&[]` or always require an allocation, even if the entire text is `WordBreak::Normal`.) The alternative would be foregoing storing and either monomorphizing or doing `dyn` dispatch.

nicoburns (approved):

PR #721, opened 2026-07-30
Consistently pass through the root style for empty text

tomcur:
On top of https://github.com/linebender/parley/pull/720, as otherwise that code hits an underflow.

Empty text still needs a style, e.g., to be able to size cursors. This PR plumbs the root style through. I hit this when making font selection infallible, which tripped up the `builders_empty` test.

(This might also be useful for sizing "struts".)

Because with this change these empty layouts now actually have a style, that means font selection now succeeds in the `builders_empty` test, where it previously failed. Not having a font was the only reason there weren't any runs in the changed test, because, for empty text, `parley` actually shapes a space character, which should lead to a run.

I've chosen to remove checking that runs are empty from the test, as I think it's mostly checking the wrong thing at the moment. If we ever stop synthesizing a space character for empty text, we should revisit this.

### DIFF
--- a/parley/src/builder.rs
+++ b/parley/src/builder.rs
@@ -187,6 +187,22 @@ impl<'b, B: Brush> StyleRunBuilder<'b, B> {
             self.cursor == self.len,
             "StyleRunBuilder requires runs that cover the full text"
         );
+        if self.lcx.style_runs.is_empty() {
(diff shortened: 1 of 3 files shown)


nicoburns (approved):
Feels like we ought to do something more principled here, but happy to go with this for now.

PR #722, opened 2026-07-30
`parley_engine`: Abort shaping if no font is available, attempt a last resort scan in `parley`

tomcur:
This makes the font selection callback `parley_engine` takes infallible. That's important, because a font is needed to be able to give font metrics to a `ShapedRun`. As my intention is to move to a `ShapedText` model where the entire source text is covered, to simplify the API and reduce some bookkeeping, that means there shouldn't be gaps in `ShapedRun`s.

`parley` is updated to always pass a font. This PR proposes the very simple/naive protocol of just taking any font from the collection as a final "fallback font", which is used if the font query itself doesn't return any font. If there are no fonts in the collection at all, shaping is completely skipped.

#720 and #721 fixed some pre-existing bugs around empty text, which this infallible font selection surfaced (see #721 in particular).

### DIFF
--- a/parley/src/shape/mod.rs
+++ b/parley/src/shape/mod.rs
@@ -4,7 +4,6 @@
 //! Text shaping implementation using `harfrust`for shaping
 //! and `icu` for text analysis.
 
-use alloc::vec::Vec;
 use parley_engine::shape::{CharCluster, Coverage};
 use parley_engine::{Analysis, AnalysisDataSources, FontInstance, ShapeOptions, Shaper};
 
@@ -39,15 +38,7 @@ pub(crate) fn shape_text<'a, B: Brush>(
     }
 
     // Do nothing if there is no text or styles (there should always be a default style)
-    let last_resort_font;
-    if text.is_empty() || styles.is_empty() || {
-        // Get any font (and only at this point, because if we returned due to empty text or styles,
-        // we don't need to build the font).
-        last_resort_font = any_font(fcx);
-
-        // If `true`, there isn't even a single font we could use to shape.
-        last_resort_font.is_none()
-    } {
+    if text.is_empty() || styles.is_empty() {
         // Process any remaining inline boxes whose index is greater than the length of the text
         for box_idx in 0..inline_boxes.len() {
             // Push the box to the list of items
@@ -55,7 +46,6 @@ pub(crate) fn shape_text<'a, B: Brush>(
         }
         return;
     }
-    let last_resort_font = last_resort_font.unwrap();
 
     let mut fq = fcx.collection.query(&mut fcx.source_cache);
 
@@ -114,15 +104,8 @@ pub(crate) fn shape_text<'a, B: Brush>(
 
         let style_index = char_style_indices[item.range.char_range.start];
         let style = &styles[usize::from(style_index)];
-        let mut font_selector = FontSelector::new(
-            &mut fq,
-            rcx,
-            styles,
-            style_index,
-            item.script,
-            style.locale,
-            &last_resort_font,
-        );
+        let mut font_selector =
+            FontSelector::new(&mut fq, rcx, styles, style_index, item.script, style.locale);
 
         let shaped_runs_range = scx.shape_item(
             text,
@@ -136,7 +119,7 @@ pub(crate) fn shape_text<'a, B: Brush>(
                 char_style_indices,
             },
             #[inline(always)]
-            |char_cluster| Some(font_selector.select_font(char_cluster, analysis_data_sources)),
+            |char_cluster| font_selector.select_font(char_cluster, analysis_data_sources),
             &mut layout.data.shaped_text,
         );
         for shaped_run_idx in shaped_runs_range {
@@ -158,6 +141,14 @@ pub(crate) fn shape_text<'a, B: Brush>(
     }
 }
 
+/// The font to use if a font query doesn't return any font candidates at all.
+#[derive(Debug)]
+enum LastResortFont {
+    Unresolved,
+    Resolved(FontInstance),
+    Unavailable,
+}
+
 struct FontSelector<'a, 'b, B: Brush> {
     query: &'b mut Query<'a>,
     fonts_id: Option<usize>,
@@ -169,14 +160,11 @@ struct FontSelector<'a, 'b, B: Brush> {
     features: &'a [FontFeature],
 
     /// The font to use if [`Self::query`] doesn't return any font.
-    last_resort_font: &'b QueryFont,
+    last_resort_font: LastResortFont,
 }
 
 impl<'a, 'b, B: Brush> FontSelector<'a, 'b, B> {
     /// Construct a new `FontSelector`.
-    ///
-    /// If `query` ends up not returning a font for a query, the `last_resort_font` is returned
-    /// instead.
     fn new(
         query: &'b mut Query<'a>,
         rcx: &'a ResolveContext,
@@ -184,7 +172,6 @@ impl<'a, 'b, B: Brush> FontSelector<'a, 'b, B> {
         style_index: u16,
         script: Script,
         locale: Option<Language>,
-        last_resort_font: &'b QueryFont,
     ) -> Self {
         let style = &styles[style_index as usize];
         let fonts_id = style.font_family.id();
@@ -210,7 +197,7 @@ impl<'a, 'b, B: Brush> FontSelector<'a, 'b, B> {
             attrs,
             variations,
             features,
-            last_resort_font,
+            last_resort_font: LastResortFont::Unresolved,
         }
     }
 
@@ -218,7 +205,7 @@ impl<'a, 'b, B: Brush> FontSelector<'a, 'b, B> {
         &mut self,
         cluster: &mut CharCluster,
         analysis_data_sources: &AnalysisDataSources,
-    ) -> FontInstance {
+    ) -> Option<FontInstance> {
         let style_index = cluster.style_index();
         let is_emoji = cluster.is_emoji();
         if style_index != self.style_index || is_emoji || self.fonts_id.is_none() {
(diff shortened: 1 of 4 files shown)


nicoburns:
What happens if the collection is empty?

tomcur:
> What happens if the collection is empty?

That's handled the same way as if the text was empty: shaping is completely skipped. https://github.com/linebender/parley/pull/722/changes#diff-61f3c5230660ff9e41294f99bee211592e8b099810136ffba274e7ff93b9381eL40-R51

parley/src/shape/mod.rs
```diff
@@ -4,24 +4,26 @@
...
 
 use fontique::{self, Query, QueryFamily, QueryFont};
 use parlance::Script;
 
 #[allow(clippy::too_many_arguments)]
 pub(crate) fn shape_text<'a, B: Brush>(
     rcx: &'a ResolveContext,
-    mut fq: Query<'a>,
+    fcx: &'a mut FontContext,
+    // mut fq: Query<'a>,
```
taj-p:
```suggestion
```

parley/src/shape/mod.rs
```diff
@@ -4,24 +4,26 @@
 //! Text shaping implementation using `harfrust`for shaping
```
taj-p, line 4: Funny branch name "tomcur:push-tqzwkknvsuyv" 😆 😛

parley/src/shape/mod.rs
```diff
@@ -267,9 +287,34 @@ impl<'a, 'b, B: Brush> FontSelector<'a, 'b, B> {
...
+            .unwrap_or(self.fallback_font.clone())
     }
 }
 
+/// Just select any font from the font collection.
+///
+/// If this returns `None`, there are no fonts available at all. This can be used to have a font at
+/// hand for shaping, in case more sophisticated font querying returns no fonts.
+fn any_font(fcx: &mut FontContext) -> Option<FontInstance> {
+    let name = fcx.collection.family_names().next()?.to_owned();
```
taj-p: It's unfortunate we have to pay a String allocation for every invocation just to immediately drop it 🤔 . 

I'm also unsure whether we want to enumerate every family and look up its contents to determine the first "populated" font (to address the below). 

It might be best to add an API into fontique that returns the first (if any) font. An alternative solution might look at amending the query to have some "pre-run check" / "last resort" mechanism to encapsulate the "please always load at least one font"

parley/src/shape/mod.rs
```diff
@@ -267,9 +287,34 @@ impl<'a, 'b, B: Brush> FontSelector<'a, 'b, B> {
...
+    let font = fcx
+        .collection
+        .family_by_name(&name)?
+        .default_font()?
+        .clone();
```
taj-p: I'm not sure this is entirely correct because a family can have all its fonts unregistered by a consumer wanting to minimize their memory consumption. So, I think a consumer can register a few families, drain the first, and this will return None (preventing all shaping from occurring).

AI wrote me this unit test:

```rs
#[cfg(test)]
mod tests {
    use std::sync::Arc;

    use fontique::{Blob, Collection, CollectionOptions, FontInfoOverride};

    use super::*;

    #[test]
    fn any_font_skips_empty_families() {
        let font_path = parley_dev::font_dirs()
            .nth(1)
            .unwrap()
            .join("Roboto-Regular.ttf");
        let font_data = Blob::new(Arc::new(std::fs::read(font_path).unwrap()));
        let mut collection = Collection::new(CollectionOptions {
            shared: false,
            system_fonts: false,
        });

        for family_name in ["First family", "Second family"] {
            collection.register_fonts(
                font_data.clone(),
                Some(FontInfoOverride {
                    family_name: Some(family_name),
                    ..Default::default()
                }),
            );
        }

        // Empty whichever family `any_font` will inspect first. Family names remain registered
        // after their last font is removed, so the other family is still a valid fallback.
        let first_family_name = collection.family_names().next().unwrap().to_owned();
        let first_family = collection.family_by_name(&first_family_name).unwrap();
        for font in first_family.fonts() {
            assert!(collection.unregister_font(
                first_family.id(),
                font.width(),
                font.style(),
                font.weight(),
            ));
        }
        assert!(
            collection
                .family_by_name(&first_family_name)
                .unwrap()
                .fonts()
                .is_empty()
        );
        let remaining_family_name = if first_family_name == "First family" {
            "Second family"
        } else {
            "First family"
        };
        assert!(
            collection
                .family_by_name(remaining_family_name)
                .unwrap()
                .default_font()
                .is_some()
        );

        let mut fcx = FontContext {
            collection,
            source_cache: Default::default(),
        };
        assert!(
            any_font(&mut fcx).is_some(),
            "a later family still contains a usable font"
        );
    }
}

```

parley/src/shape/mod.rs
```diff
@@ -36,15 +38,27 @@ pub(crate) fn shape_text<'a, B: Brush>(
...
+        fallback_font.is_none()
+    } {
         // Process any remaining inline boxes whose index is greater than the length of the text
         for box_idx in 0..inline_boxes.len() {
             // Push the box to the list of items
             layout.data.push_inline_box(box_idx);
         }
         return;
     }
+    let fallback_font = fallback_font.unwrap();
```
taj-p: The name "fallback font" is a bit ambiguous because the concept also exists in fontique with a different meaning. I think "last resort font" or similar could be a better name.

parley/src/shape/mod.rs
tomcur: That's [a `jj` thing](https://docs.jj-vcs.dev/latest/github/#using-a-generated-bookmark-name) (if you're too lazy to actually specify a name).

parley/src/shape/mod.rs
tomcur: I agree.

As we'll be moving to a `Shaper::shape_text` signature where `parley_engine` owns the item loop, another option is for the `select_font` callback to remain fallible (`-> Option<FontInstance>`), and to instead document that returning `None` aborts shaping.

We'll still want some sort of last-resort font mechanism, but we then only need to instantiate that font if the query itself fails. If looking up a last-resort *also* fails, then unless the user is doing strange things, the first query should've failed anyway, so we won't have wasted much shaping effort.

> It might be best to add an API into fontique that returns the first (if any) font. An alternative solution might look at amending the query to have some "pre-run check" / "last resort" mechanism to encapsulate the "please always load at least one font"

That could be good. I suppose that warrants somewhat more thought about what exactly that should look like. (For example, a user might want to choose a specific font, e.g., to render with [Unicode's last resort font](https://github.com/unicode-org/last-resort-font).)

parley/src/shape/mod.rs
tomcur: > I'm not sure this is entirely correct because a family can have all its fonts unregistered by a consumer

Ah... thank you. That's unfortunate. I didn't realize that was possible!

parley/src/shape/mod.rs
tomcur: Agree. Renamed this to "last resort font".

parley/src/shape/mod.rs
tomcur: I've rebased to be on top of #725, and updated `parley`'s `any_font` to use `fontique`'s normal querying mechanisms, so should now be able to find a font. It first tries the first of each generic family.

If the first of each generic family also fails to find a find, it iterates using the `fontique::Collection::family_ids` from #725. That does still need to allocate due to the mutable borrow, but is only hit in the unlikely case there are no fonts from text families available at all.

If we also take the idea from https://github.com/linebender/parley/pull/722#discussion_r3698628486, we only ever need to call this if the normal query fails to return anything.

parley/src/shape/mod.rs
taj-p: Ohhhh TIL. Thank you 😊

parley/src/shape/mod.rs
taj-p: > That does still need to allocate due to the mutable borrow, but is only hit in the unlikely case there are no fonts from text families available at all.

For how I consume Parley (and on Wasm if no fallbacks are explicitly specified), I think this will be hit every time 😅

tomcur: Right... it's a bit unfortunate that `Collection`'s methods take `&mut self`... Perhaps with what I suggested in https://github.com/linebender/parley/pull/722#discussion_r3698628486 this is still mostly fine, because `any_font` will then only be called if the original query failed to return a font.

There are two further possibilities, though:

1. we could indeed add something to `fontique` to just get a font. Generalizing that a bit, it would probably be an iterator over fonts. The downside is that if someone were to consume the full iterator, I believe it'd materialize all the fonts. That could just be a documented caveat;
2. or, as a band-aid until we know what to do about a more general last-resort mechanism, we could update `any_font` to iterate `for n in 0.. { if let Some(family_id) = fcx.collection.family_ids().nth(n) { ... } }`, the downside being it has now become an `O(n^2)` walk, but without allocation, and probably just succeeding on the first family anyway (unless the user is doing strange things).

parley/src/shape/mod.rs
taj-p: > As we'll be moving to a Shaper::shape_text signature where parley_engine owns the item loop, another option is for the select_font callback to remain fallible (-> Option<FontInstance>), and to instead document that returning None aborts shaping.

I like this option!

parley/src/shape/mod.rs
taj-p: https://github.com/linebender/parley/pull/722#discussion_r3698628486 looks good!

We can also optimise the `any_font` implementation somewhat.

1. Only issue a query if there are generic families / fallbacks to query
2. Greedily assume the first family has a font
3. Store the allocation on context so we're just paying a copy instead of an allocation if we fall through to the "copy all family ID"s path

```rs
fn any_font(fcx: &mut FontContext, scratch: &mut Vec<fontique::FamilyId>) -> Option<QueryFont> {
    scratch.clear();

    // First try the system's preferred generic text families.
    //
    // At this point, the font itself doesn't really matter: we mostly just need any font to work
    // with. We take the first from each generic family, as on most systems that should "just work"
    // and give a sensible font, while allowing us to skip extending our query's candidate font
    // families with all the generic families registered on the system.
    let sans = fcx
        .collection
        .generic_families(GenericFamily::SansSerif)
        .next();
    let serif = fcx.collection.generic_families(GenericFamily::Serif).next();
    let monospace = fcx
        .collection
        .generic_families(GenericFamily::Monospace)
        .next();
    // Plus any font covering Latin.
    let latin = fcx
        .collection
        .fallback_families(fontique::FallbackKey::new(
            Script::from_bytes(*b"Latn"),
            None,
        ))
        .next();
    let preferred = [sans, serif, monospace, latin];

    if preferred.iter().any(Option::is_some) {
        let mut query = fcx.collection.query(&mut fcx.source_cache);
        query.set_families(preferred.into_iter().flatten());
        if let Some(font) = first_query_font(&mut query) {
            return Some(font);
        }
    }

    // Greedy path: Just check the first ID.
    let first_family = fcx.collection.family_ids().next()?;
    {
        let mut query = fcx.collection.query(&mut fcx.source_cache);
        query.set_families(core::iter::once(first_family));
        if let Some(font) = first_query_font(&mut query) {
            return Some(font);
        }
    }

    // Fallback path: Collect the remaining family IDs in the reusable buffer.
    scratch.extend(
        fcx.collection
            .family_ids()
            .filter(|family| *family != first_family),
    );
    if scratch.is_empty() {
        return None;
    }
    let mut query = fcx.collection.query(&mut fcx.source_cache);
    query.set_families(scratch.drain(..));
    first_query_font(&mut query)
}
```

That said, a typical use case is a long lived `FontContext` that lives on a server - which may see tens of thousands of fonts - where unused fonts are unregistered over time. So, I fear that even this optimised implementation has gaps. 

I think your proposal to only call `any_font` when `select_font` returns None is likely best.

tomcur: This version looks the nicest to me so far. I'll update the PR tomorrow, including only calling on `None` (and aborting shaping if `any_font` also fails).

Perhaps the correct solution longer-term would be to allow users to hand some explicit last resort font to either `fontique` or `parley`.

taj-p (approved):
LGTM 🎉  once comments addressed

parley/src/shape/mod.rs
tomcur: > I think your proposal to only call `any_font` when `select_font` returns None is likely best.

Done! This did require changing the shape of `any_font` yet again, because we only have access to `&mut Query` at that point, and can't easily get access to the `Collection` anymore.

`any_font` now just queries generic families. To do the greedy check we'd have to look up the `[sans, serif, monospace, latin]` family IDs before going into shaping, and storing them, but that'd perhaps get a bit messy. We also *could* introduce something like `QueryFamily::All`, but perhaps at that point it becomes better to just have real last-resort support in `fontique` (or `parley`).

If this shape for `any_font` really doesn't spark joy, we could choose to fully remove `any_font` and accept shaping gets aborted. In any case, `any_font` probably should be removed once we have actually good last-resort behavior.

PR #723, opened 2026-07-31
Fix variable font wght synthesis for shaping

taj-p:
When a caller declares a variable font with weight `X`, it could be shaped with weight `Y` because shaping uses the font's default weight, rather than the specified `X` value. 

This PR ensures that Fontique synthesizes the correct weight when the default and declared weight conflict.

See the provided unit test. 

| Before Fix | After Fix |
| -------- | ------------|
| style_font_weight_uses_variable_axis_default-0 |  style_font_weight_uses_variable_axis_default-0 |


fontique/src/font.rs
```diff
@@ -94,13 +94,17 @@ impl FontInfo {
...
+        // when the axis default and requested weight are not the same, synthesize
+        // the requested weight for correct shaping.
+        if let Some(weight_axis) = self.axes.iter().find(|axis| axis.tag == Tag::new(b"wght")) {
+            if weight_axis.default != weight.value() {
                 synth.vars[len] = (Tag::new(b"wght"), weight.value());
                 len += 1;
-            } else if weight.value() > self.weight.value() {
-                synth.embolden = true;
             }
+        } else if weight.value() > self.weight.value() {
```
behdad, line 111: This condition needs tweaking IMO. Should only embolden if requested weight is at least 200/300 CSS units more, not just 1, or 100 for that matter.

behdad: Also, doesn't the embolden() get a strength by any chance?

nicoburns: There is a strength parameter on the actual embolden impl in `kurbo` (exposed on `vello`). Don't think we have it wired through the text parts of the stack which I think might still be using more of an RBIZ model (yes/no) for embolden.

taj-p: I think Chromium uses a "at least 200 units more" approach to trigger embolden:

https://chromium.googlesource.com/chromium/src/%2B/39f1c9a7161786d7c54f394dfbfca8941f78fd8b/third_party/blink/renderer/platform/fonts/skia/font_cache_skia.cc#311

I'm happy to add that change in as a separate PR?

taj-p:
cc @dinox0r - this fixes that bug you found 🙏

tomcur (commented):
This makes sense to me, with the caveat that I'm still learning about fonts!

I think we should document that if the two values disagree, the font is non-conformant. The spec says the two values must be identical: https://learn.microsoft.com/en-us/typography/opentype/spec/dvaraxistag_wght#additional-information.

I believe the same can happen for `slnt`. The font can have a non-zero `slnt` but claim it is `FontStyle::Normal`. Request `FontStyle::Normal`, and the font will then actually render slanted. Here the spec again says attributes must agree https://learn.microsoft.com/en-us/typography/opentype/spec/dvaraxistag_slnt#additional-information.

I do wonder if we should instead, in `from_font_ref`, do a fix-up of the font's attributes based on the variable axes' default values. If the attributes are changed, they no longer agree with what the font claimed, but as the font is non-conformant anyway, that's perhaps fine.

https://github.com/linebender/parley/blob/9a56b698cb2a649549b3956c6822c600c76c6b56/fontique/src/font.rs#L219-L235

taj-p:
> I do wonder if we should instead, in `from_font_ref`, do a fix-up of the font's attributes based on the variable axes' default values. If the attributes are changed, they no longer agree with what the font claimed, but as the font is non-conformant anyway, that's perhaps fine.

I think this might be somewhat orthogonal. I think this is the question of whether `from_font_ref` should try fix up / normalize font properties - rather than report what the font file says. I don't have context on whether we have a position / philosophy here.

It is orthogonal because even if we fix up the attributes in `from_font_ref`, we'd still require these changes to synthesis generation because `apply_override` can change weight. Consider:

1. `from_font_ref` reads Roboto Flex as weight 400 with wght default 300; weight is fixed up to 300.
2. `apply_override` changes its matching weight to 100.
3. A caller requests weight 100.
4. Shaping still defaults to wght=300, so synthesis must emit wght=100.

taj-p:
> I think we should document that if the two values disagree, the font is non-conformant. The spec says the two values must be identical: https://learn.microsoft.com/en-us/typography/opentype/spec/dvaraxistag_wght#additional-information.

Happy to add this! Where were you thinking of putting it? In `synthesis` or `FontInfoOverride` or somewhere else?

tomcur (approved):
> even if we fix up the attributes in `from_font_ref`, we'd still require these changes to synthesis generation because `apply_override` can change weight

Ah, right!

> Happy to add this! Where were you thinking of putting it? In `synthesis` or `FontInfoOverride` or somewhere else?

I think `synthesis` is right, saying that we need to read the variable axis' default if present, because it can differ from the font's weight attribute if 1) a `FontInfoOverride` has been set or 2) the font itself is non-conformant (as in, if the font is specifying different values in the OS/2 and fvar tables).

PR #727, opened 2026-08-03
`parley_engine`: Always shape full text, ensuring `ShapedText` covers the full text

tomcur:
To shape, you now call `Shaper::shape_text`, which owns the itemization loop. Together with font selection being infallible (#722), `ShapedText` is then guaranteed to represent the full source text contiguously. That's useful to simplify some bookkeeping. See https://github.com/linebender/parley/pull/715#discussion_r3680380707 for some prior discussion on this. Because users previously looped items themselves and would call `Shaper::shape_item`, that meant items could be skipped or submitted out-of-order.

Because control flow moved into `parley_engine`, some structures are introduced. `Shaper::shape_text` takes an iterator of items encoding the span and holding the item's `ShapeOptions` (previously, `parley` constructed the options inside the loop and passed them as a parameter to `Shaper::shape_item`). Font selection is now a trait instead of a callback, with one method for selection and one method called at the start of each segment.

Note, "segment" is a new name, and encodes maximal spans within the user's items that have constant bidi level and script. The distinction between items and segments becomes more important later on, because browsers reset grapheme segmentation at item boundaries (where shaping options change), but not at segment boundaries.

Performance on the Japanese benchmarks is improved quite a bit, mostly because we're smarter about reusing the `FontSelector`. Performance on Arabic and Latin is about +1.5%, but addressing the TODO I left on the the item iterator in `parley` appears to make this a net small performance win on those scripts (i.e., iterating style runs and inline boxes to figure out where item boundaries are, instead of an `O(chars)` walk). That's for a separate PR: I just had an LLM address that TODO for an initial measurement, and have not reviewed that code at all.

<details>
<summary>Benchmarks with the state as in this PR</summary>

```
$ cargo bench --bench main -- compare ../target/benchmarks/main -t 8.

Default Style - arabic 20 characters               [   9.0 us ...   9.2 us ]      +1.42%*
Default Style - latin 20 characters                [   4.5 us ...   4.5 us ]      +1.62%*
Default Style - japanese 20 characters             [   8.6 us ...   8.0 us ]      -7.29%*
Default Style - arabic 1 paragraph                 [  48.6 us ...  49.1 us ]      +1.12%*
Default Style - latin 1 paragraph                  [  17.4 us ...  17.7 us ]      +1.75%*
Default Style - japanese 1 paragraph               [  72.6 us ...  64.9 us ]     -10.66%*
Default Style - arabic 4 paragraph                 [ 202.9 us ... 205.2 us ]      +1.11%*
Default Style - latin 4 paragraph                  [  65.7 us ...  66.3 us ]      +0.82%
Default Style - japanese 4 paragraph               [ 102.4 us ...  90.8 us ]     -11.28%*
Styled - arabic 20 characters                      [  10.1 us ...  10.2 us ]      +1.15%*
Styled - latin 20 characters                       [   5.6 us ...   5.7 us ]      +1.68%*
Styled - japanese 20 characters                    [   9.3 us ...   8.5 us ]      -8.89%*
Styled - arabic 1 paragraph                        [  51.0 us ...  51.7 us ]      +1.25%*
Styled - latin 1 paragraph                         [  21.9 us ...  22.3 us ]      +1.94%*
Styled - japanese 1 paragraph                      [  78.8 us ...  71.1 us ]      -9.77%*
Styled - arabic 4 paragraph                        [ 222.8 us ... 226.6 us ]      +1.72%*
Styled - latin 4 paragraph                         [  84.9 us ...  85.4 us ]      +0.57%
Styled - japanese 4 paragraph                      [ 111.6 us ... 101.0 us ]      -9.55%*
```

</details>

<details>
<summary>Experimental: benchmarks after addressing the item iterator TODO</summary>

```
$ cargo bench --bench main -- compare ../target/benchmarks/main -t 8.

Default Style - arabic 20 characters               [   9.0 us ...   8.9 us ]      -1.04%*
Default Style - latin 20 characters                [   4.5 us ...   4.5 us ]      -0.29%
Default Style - japanese 20 characters             [   8.7 us ...   7.8 us ]      -9.50%*
Default Style - arabic 1 paragraph                 [  48.6 us ...  48.3 us ]      -0.51%
Default Style - latin 1 paragraph                  [  17.5 us ...  17.5 us ]      +0.09%
Default Style - japanese 1 paragraph               [  72.5 us ...  63.3 us ]     -12.76%*
Default Style - arabic 4 paragraph                 [ 203.0 us ... 201.7 us ]      -0.64%
Default Style - latin 4 paragraph                  [  65.7 us ...  66.2 us ]      +0.64%
Default Style - japanese 4 paragraph               [ 102.5 us ...  88.8 us ]     -13.34%*
Styled - arabic 20 characters                      [  10.1 us ...  10.0 us ]      -0.78%
Styled - latin 20 characters                       [   5.7 us ...   5.6 us ]      -1.37%*
Styled - japanese 20 characters                    [   9.3 us ...   8.4 us ]      -9.73%*
Styled - arabic 1 paragraph                        [  51.3 us ...  50.7 us ]      -1.23%*
Styled - latin 1 paragraph                         [  21.9 us ...  21.9 us ]      +0.03%
Styled - japanese 1 paragraph                      [  78.9 us ...  69.4 us ]     -12.10%*
Styled - arabic 4 paragraph                        [ 222.8 us ... 220.8 us ]      -0.92%
Styled - latin 4 paragraph                         [  84.9 us ...  84.8 us ]      -0.15%
Styled - japanese 4 paragraph                      [ 111.7 us ...  97.8 us ]     -12.45%*
```

</details>

### DIFF
--- a/parley_engine/src/itemize.rs
+++ b/parley_engine/src/itemize.rs
@@ -51,7 +51,6 @@ pub struct Item_<'a> {
 
     /// The options to shape this item with.
     pub options: ShapeOptions<'a>,
-
     // TODO: we probably should allow users to pass in some data (like we allow passing
     // style_indices elsewhere), which we copy onto `ShapedRun`. That allows users to easily
     // correlate `ShapedRun`s with some data they themselves hold.
--- a/parley_engine/src/itemize.rs
+++ b/parley_engine/src/itemize.rs
@@ -52,6 +52,16 @@ pub struct Item<'a> {
     pub char_end: u32,
 
     /// The options to shape this item with.
+    //
+    // TODO: should users instead be allowed to build `options` at the `Segment` level (i.e.,
+    // through some callback)? The motivation is for users to have access to the segment's `Script`
+    // and be able to set font features based on that. If the entirety of `ShapeOptions` moves
+    // there, it would allow users to set, e.g., a font size per segment, even though it's not
+    // necessarily an item boundary; and note an item then doesn't mean a whole lot anymore. We
+    // could also allow only some options to be per-segment. In any case, we're probably moving
+    // towards a future where items reset grapheme segmentation, but segments do not (note Gecko and
+    // Blink also reset grapheme segmentation when something like font size changes, but not when
+    // the script changes).
     pub options: ShapeOptions<'a>,
     // TODO: we probably should allow users to pass in some data (like we allow passing
     // style_indices elsewhere), which we copy onto `ShapedRun`. That allows users to easily
--- a/parley_engine/src/itemize.rs
+++ b/parley_engine/src/itemize.rs
@@ -42,7 +42,10 @@ pub struct Segment {
 
 /// A span of text shaped with specific [`ShapeOptions`].
 ///
-/// While shaping text, items are divided into [`Segment`]s.
+/// An [`Item`] represents a sequence of constant [`ShapeOptions`], but cannot always be passed to
+/// the shaper as a single unit. Within an item, the script or bidirectional text embedding level
+/// may change, which requires further splitting the item into segments of constant script and bidi
+/// level (see [`Segment`]).
 #[derive(Debug)]
 pub struct Item<'a> {
     /// The character offset in the source text which this item ends.
--- a/parley_engine/src/itemize.rs
+++ b/parley_engine/src/itemize.rs
@@ -43,13 +43,24 @@ pub struct Item {
 /// An item produced by [`Analysis::itemize`].
 #[derive(Debug)]
 pub struct Item_<'a> {
-    /// The options to shape this item with.
-    pub options: ShapeOptions<'a>,
     /// The character offset in the source text which this item ends.
     ///
     /// This must be strictly greater than the previous item's end. For the first item, it must be
     /// greater than 0.
     pub char_end: u32,
+
+    /// The options to shape this item with.
+    pub options: ShapeOptions<'a>,
+
+    // TODO: we probably should allow users to pass in some data (like we allow passing
+    // style_indices elsewhere), which we copy onto `ShapedRun`. That allows users to easily
+    // correlate `ShapedRun`s with some data they themselves hold.
+    //
+    // This is potentially important for better correctness in `parley`: it itemizes based on
+    // `nearly_eq` of shaping-relevant styles like font size, i.e., it should then read that item's
+    // style to know the font size, even though it may be different for the run.
+    // /// Opaque data copied onto every `ShapedRun` produced from this item.
+    // pub user_data: u16,
 }
 
 /// Produces the items in a text via [`Self::next`]; created by [`Analysis::itemize`].
--- a/parley_engine/src/itemize.rs
+++ b/parley_engine/src/itemize.rs
@@ -8,7 +8,7 @@ use core::{ops::Range, str::CharIndices};
 use icu_properties::props::Script as IcuScript;
 use parlance::{BidiLevel, Script};
 
-use crate::{Analysis, CharInfo};
+use crate::{Analysis, CharInfo, ShapeOptions};
 
 /// A range of text.
 #[derive(Clone, Debug, PartialEq, Eq)]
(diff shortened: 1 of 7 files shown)


parley_engine/src/itemize.rs
```diff
@@ -40,8 +40,43 @@ pub struct Item {
     pub script: Script,
 }
 
-/// An iterator over items in text, produced by [`Analysis::itemize`].
-pub struct Itemizer<'a, F> {
+/// A span of text shaped with specific [`ShapeOptions`].
+///
+/// While shaping text, items are divided into [`Segment`]s.
```
taj-p:
```suggestion
/// An [`Item`] represents a sequence of constant `ShapeOptions`, but, cannot always be passed to
/// the shaper as a single unit. When script or Bidi change within an [`Item`], we must split the [`Item`] 
/// into segments of constant Bidi and script (see [`Segment`]). 
```

parley_engine/src/itemize.rs
```diff
@@ -20,9 +20,9 @@ pub struct TextRange {
     pub char_range: Range<usize>,
 }
 
-/// An item produced by [`Analysis::itemize`].
+/// A span of text inside an [`Item`] with constant script and bidirectional embedding level.
 #[derive(Clone, Debug)]
-pub struct Item {
+pub struct Segment {
```
taj-p, line 25: I struggled a bit with the new names in this file.

The flow today is roughly:

- callers pass `Item`s to `shape_text`
- `shape_text` uses an `Itemizer` to split each item range into `Segment`s
- each `Segment` produces one or more `ShapedRuns`

I found it surprising that an `Itemizer` produces `Segment`s rather than `Item`s. At the same time, `Segmenter` (as a rename) or similar doesn't quite feel right. Both `Item` and `Segment` also feel fairly generic 🤔

I'm wondering whether we should rename Item to `ShapingContextSpan` and `Segment` to `ItemizedRun`.

Then the flow would read:

- callers pass `ShapingContextSpan`s to `shape_text`
- the `Itemizer` splits them into `ItemizedRun`s
- each `ItemizedRun` produces one or more `ShapedRun`s

I don't feel too strongly here - especially considering the TODO comments which suggest the naming and semantics may change significantly as this migration evolves. Please defer to your own best judgement here.

parley_engine/src/shape/shaper.rs
```diff
@@ -33,6 +33,8 @@ pub struct ShapeOptions<'a> {
     pub variations: &'a [FontVariation],
     /// The per-character style indices.
     // TODO: rename to something like `user_data` (s.t. we don't assume it's a style per se).
+    // TODO: probably move this out of `ShapeOptions`, and supply it as a parameter on
+    // `Shaper::shape_text`.
```
taj-p, line 40: 👍

parley_engine/src/shape/shaper.rs
```diff
@@ -78,62 +80,109 @@ impl core::fmt::Debug for Shaper {
...
+                    &mut select_font,
+                    analysis.char_info(),
+                    shaped_text,
+                )
+                .is_err()
+                {
+                    // Abort on error. This happens iff `FontSelector::select_font` failed to return a
+                    // font. By aborting we ensure `ShapedText` covers the source text contiguously (as
+                    // we need a font to construct `ShapedRun`).
+                    return;
```
taj-p, line 152: Do we plan to return an `Error` from `shape_text` in the future?

taj-p (approved):
LGTM!

parley_engine/src/itemize.rs
tomcur: Yeah, very fair. I've decided not to fix it here yet. It's likely the exact semantics are going to change a bit more.

Two bits of additional context perhaps help.

First, the distinction between the "item" and "segment" is not really a technical necessity, but it follows browser behavior: browsers reset grapheme segmentation at their "item" boundaries (style changes), but they don't reset on script changes (and Blink also doesn't reset on bidi changes). This moves us in that direction. But, if my understanding is correct, style changes don't technically *need* to reset grapheme segmentation, so it's just a choice.

Second, an item is user-provided, whereas segments are derived from the text.

> At the same time, `Segmenter` (as a rename) or similar doesn't quite feel right. 

I agree, especially because it may be confused with the icu4x segmenters. One option is to perhaps name everything "items" again, where users can provide additional item boundaries and perhaps can specify whether they want those to reset grapheme segmentation or not. `ShapeOptions` could be provided by some callback called by `parley_engine` with the item ranges (i.e., the user's boundaries, plus constant bidi and script).

parley_engine/src/shape/shaper.rs
tomcur: Hm, perhaps we should.

PR #728, opened 2026-08-04
Add `parlance::Script::is_cursive`

tomcur:
This encodes whether a script is cursive according to [CSS Text 3](https://www.w3.org/TR/css-text-3/#script-groups).

Letter spacing in the form of gaps should not be applied between graphemes of these scripts, and perhaps optional ligation should not be turned off. This doesn't wire it up yet in `parley`. See also [CSS Text 3 § 6.4.4](https://www.w3.org/TR/css-text-3/#justify-cursive) about cursive justification by elongation.

Because this list is spec'd by CSS, it's probably general enough to live in `parlance`, though it does mean `parlance` grows a data table.

See the same list in
- Gecko: https://github.com/mozilla-firefox/firefox/blob/d0d8c04b85032c5bd92bbc60d0c73bfdfbeed6ce/intl/components/src/UnicodeProperties.h#L348-L355
- Blink: https://github.com/chromium/chromium/blob/41fe3640cac5688a644a7a04799d996b898a74c0/third_party/blink/renderer/platform/text/character.cc#L353-L363

### DIFF
--- a/parlance/src/script.rs
+++ b/parlance/src/script.rs
@@ -73,6 +73,30 @@ impl Script {
     pub fn parse(s: &str) -> Result<Self, ParseScriptError> {
         s.parse()
     }
+
+    /// Whether this is a cursive script according to [CSS Text 3 Appendix D][css-cursive-scripts].
+    ///
+    /// As per that specification, "these scripts do not admit gaps between their letters for either
+    /// justification or letter-spacing."
+    ///
+    /// See also [CSS Text 3 § 6.4.4][css-cursive-justification] on cursive justification.
+    ///
+    /// [css-cursive-scripts]: https://www.w3.org/TR/css-text-3/#script-groups
+    /// [css-cursive-justification]: https://www.w3.org/TR/css-text-3/#justify-cursive
+    #[inline(always)]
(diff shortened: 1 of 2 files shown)


parlance/src/script.rs
```diff
@@ -73,6 +73,30 @@ impl Script {
...
+        // Script identifiers from UAX #31: <https://www.unicode.org/reports/tr31/>.
+        matches!(
+            &self.raw,
+            b"Arab" | // Arabic
+            b"Rohg" | // Hanifi Rohingya
+            b"Mand" | // Mandaic
+            b"Mong" | // Mongolian
+            b"Nkoo" | // N'ko
+            b"Phag" | // Phags Pa
+            b"Syrc" // Syriac
```
tomcur, line 97: We *could* add these as associated constants on `Script`, but I didn't want to start us down that path yet (or at least not in a way that's `pub`).

DJMcNab (approved):
Seems reasonable to me. This is the first time I've heard of inter-character justification. I don't think I'd expect to see that in English text basically ever! (I guess if you have a single-word slogan thing, you *could* want it?)

PR #730, opened 2026-08-05
Add ranged style overlays to PlainEditor

AdrianEddy:
`PlainEditor` currently supports exactly one style for the whole text. That is right for plain editing, but many editors need to *display* per-range styling on top of the text without becoming rich-text editors: syntax highlighting, spellcheck squiggles, search-match emphasis, linter underlines. Today the only options are to fork the editor or to rebuild a parallel layout by hand.

This PR adds a style overlay:

```rust
editor.set_style_overlay(vec![
    (StyleProperty::Underline(true), 0..5),
    (StyleProperty::Brush(red), 6..11),
]);
```

- The overlay is a list of `(StyleProperty<'static, T>, Range<usize>)` pairs over the raw text, applied in `update_layout` between the default styles and the IME preedit underline — so the preedit underline always stays visible while composing.
- Entries are applied in order; later entries win where they overlap.
- Setting the overlay marks the layout dirty (like `edit_styles`); setting an overlay equal to the current one is a no-op.
- Ranges are not adjusted when the text is edited. The overlay is meant for *derived* styling that the application recomputes from the text, so tracking edits would be wasted work — the caller sets a fresh overlay after editing.
- Stale entries are safe in the meantime: ranges are clamped to the text length, and entries whose bounds no longer fall on char boundaries are skipped until a new overlay is set (covered by a test that edits the text under a retained overlay).
- Pairs with #729: with deferral enabled, batching edits and setting the new overlay before the next read makes the edit-plus-restyle update atomic — the single rebuild carries fresh styling. Without it, the overlay simply applies from the next rebuild after it is set.

`get_style_overlay` returns the current overlay. No behavior changes for editors that never call the new methods.

Tests cover: ranges resolving to the right clusters, layout invalidation and the equal-overlay no-op, preedit-underline precedence, overlay retained across edits, and out-of-bounds clamping.

<sup>This PR was generated by Claude</sup>


parley/src/editing/editor.rs
```diff
@@ -1040,6 +1046,38 @@ where
...
+    /// To update the overlay atomically with a batch of edits, defer layout
+    /// rebuilds (`set_defer_layout`, where available) and set the new overlay
+    /// before the next read; the single rebuild then carries fresh styling.
+    ///
+    /// Setting an overlay equal to the current one is a no-op.
+    pub fn set_style_overlay(&mut self, overlay: Vec<(StyleProperty<'static, T>, Range<usize>)>) {
+        if self.style_overlay == overlay {
+            return;
+        }
+        self.style_overlay = overlay;
```
DJMcNab: It seems surprising to me that we wouldn't want to allow reusing the allocation.

parley/src/editing/editor.rs
```diff
@@ -1040,6 +1046,38 @@ where
...
+    ///
+    /// Out-of-bounds ranges are clamped to the text length; reversed or empty ranges
+    /// are ignored; entries whose clamped bounds do not fall on char boundaries
+    /// (e.g. stale after an edit) are skipped until a new overlay is set.
+    ///
+    /// To update the overlay atomically with a batch of edits, defer layout
+    /// rebuilds (`set_defer_layout`, where available) and set the new overlay
+    /// before the next read; the single rebuild then carries fresh styling.
+    ///
+    /// Setting an overlay equal to the current one is a no-op.
```
DJMcNab: This doc comment is much too long, and reads as Claude-authored - it features Claude's weird obsession with implementation details.

DJMcNab (commented):
I'm not necessarily opposed to the idea. What do you think about only allowing it to set the `Brush` value? That makes it much lower-risk (i.e. avoiding people trying to hack a WYSIWG editor on top).

parley/src/editing/editor.rs
AdrianEddy: I changed the shape to `edit_style_overlay()`, like `edit_styles()`

AdrianEddy:
> I'm not necessarily opposed to the idea. What do you think about only allowing it to set the `Brush` value? That makes it much lower-risk (i.e. avoiding people trying to hack a WYSIWG editor on top).

My primary use case and motivation for this shape is building a syntax highlighting editor. For that, I'd need Brush and at least Bold and Italic support. In monospace fonts, bold and italics don't change the space each glyph occupies, but that's not really a contract in this code (ie. nothing is stopping users from using this with sans fonts)

I personally prefer to have more capabilities rather than less, but if you think that creates too much burden on parley, I'd be happy to reduce it to one of the following shapes (in order of my preference):
1. Full `StyleProperty` - current PR shape
2. Brush + Underline family + Strikethrough family + FontWeight + FontStyle + FontFamily + FontSize + FontWidth
3. Brush + Underline family + Strikethrough family + FontWeight + FontStyle
4. Brush only

PR #731, opened 2026-08-05
Suppress optional ligation in case of letter spacing

tomcur:
Following [CSS Text 3 § 7.2](https://www.w3.org/TR/css-text-3/#letter-spacing-property), if letter spacing is applied, optional ligatures should not form.

The same is true for [`text-justify: inter-character`](https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Properties/text-justify#inter-character), but `parley` doesn't currently implement that.

This disables the OpenType features for "common", "discretionary" and "historic" ligatures as given by [CSS Fonts 4 § 6.4](https://www.w3.org/TR/css-fonts-4/#font-variant-ligatures-prop).

This matches Gecko. Blink additionally disables contextual alternates (`calt`), but that's perhaps not desirable: it's my understanding disabling contextual alternates potentially makes text render wrong.


parley/src/shape/mod.rs
```diff
@@ -107,14 +132,28 @@ pub(crate) fn shape_text<'a, B: Brush>(
...
 
+        let style_features = rcx.features(style.font_features).unwrap_or(&[]);
+        let features = if !nearly_zero(style.letter_spacing) {
+            if style_features.is_empty() {
+                OPTIONAL_LIGATURES_OFF.as_slice()
+            } else {
+                features_scratch.clear();
+                features_scratch
+                    .extend(OPTIONAL_LIGATURES_OFF.iter().chain(style_features).copied());
+                features_scratch.as_slice()
```
DJMcNab, line 143: I believe that it is possible on the web to explicitly set these font-features to on, and for that to overwrite the default related to letter spacing.

Image

That is, we need a comment here saying that "later values are preferred", or "earlier values are preferred", depending on which is true. This would also allow consumers to choose to implement the blink behaviour, for example, if they wanted to.

parley/src/shape/mod.rs
```diff
@@ -90,6 +114,7 @@ pub(crate) fn shape_text<'a, B: Brush>(
         }
     };
 
+    let mut features_scratch = Vec::new();
```
DJMcNab: Would it make sense for this to be a smallvec (inline capacity 8?)? I don't quite have a good sense of how likely you are to have more than say 4 feature settings set?

DJMcNab (approved):
I think we probably want to test the behaviour when the ligatures are explicitly set, as well, rather than just the default.

parley/src/shape/mod.rs
tomcur: Done; I expect more than four doesn't happen very often, but I don't know.

parley/src/shape/mod.rs
tomcur: Makes sense! Added a comment here, and docstring on `ShapeOptions::features`.

tomcur:
> I think we probably want to test the behaviour when the ligatures are explicitly set, as well, rather than just the default.

Haha, makes sense. Added that as another case in in the "interaction" test.

PR #734, opened 2026-08-06
Preserve non-breaking spaces when collapsing whitespace

### LINKED ISSUE #684: WhiteSpaceCollapse::Collapse strips non-breaking spaces at span edges (`trim_start`/`trim_end` are Unicode-aware)
`WhiteSpaceCollapse::Collapse` is meant to implement CSS-style white-space collapsing, but it strips non-breaking spaces (U+00A0) at the edges of spans, because `push_uncommitted_text` trims with the Unicode-aware `str::trim_start()` / `str::trim_end()`:

https://github.com/linebender/parley/blob/main/parley/src/resolve/tree.rs#L94-L113

```rust
if self.is_span_first
    || (self.last_item_kind == ItemKind::TextRun
        && self.text.chars().last().is_some_and(|c| c.is_ascii_whitespace()))
{
    span_text = span_text.trim_start();   // <- trims U+00A0
}
if is_span_last {
    span_text = span_text.trim_end();     // <- trims U+00A0
}
```

`char::is_whitespace` (which `trim_start`/`trim_end` use) matches U+00A0 NO-BREAK SPACE, U+2007 FIGURE SPACE, etc. Per CSS, collapsible "document white space" is only spaces, tabs and segment breaks — a no-break space "prevents both collapsing and line breaks" ([CSS Text §4.1](https://www.w3.org/TR/css-text-3/#white-space-processing)). The mid-span collapsing a few lines below gets this right by using `is_ascii_whitespace`:

```rust
// Collapse spaces
let mut last_char_whitespace = false;
span_text.chars().filter_map(|c: char| {
    let this_char_whitespace = c.is_ascii_whitespace();
    ...
```

so only the edge trims are inconsistent.

## Effect

A span consisting of just `"\u{a0}"` collapses to the empty string, and leading/trailing NBSPs disappear from any span. The real-world hit is HTML renderers built on parley (found via blitz): the classic `<td>&#160;</td>` spacer-cell pattern (ubiquitous in HTML email — bar charts, colored progress segments, spacer rows) lays out as an empty cell, so the row collapses to zero height and the colored background vanishes.

## Repro

```rust
let mut builder = tree_builder_with(WhiteSpaceCollapse::Collapse, ...);
builder.push_text("\u{a0}");
let (layout, text) = builder.build();
assert_eq!(text, "\u{a0}"); // fails: text is ""
```

## Suggested fix

Trim only ASCII white space at the span edges, matching the mid-span filter:

```rust
span_text = span_text.trim_start_matches(|c: char| c.is_ascii_whitespace());
...
span_text = span_text.trim_end_matches(|c: char| c.is_ascii_whitespace());
```

(The `self.text.chars().last()` gate above already uses `is_ascii_whitespace`, so this also makes the gate and the trim agree.)

Happy to send a PR if that direction looks right.

subotac:
Use ASCII-only trimming at collapsed text boundaries, matching the existing internal whitespace-collapse logic. This preserves non-breaking spaces while continuing to remove surrounding ASCII whitespace.

Includes a focused regression test.

Fixes #684.

### DIFF
--- a/parley/src/resolve/tree.rs
+++ b/parley/src/resolve/tree.rs
@@ -106,10 +106,10 @@ impl<B: Brush> TreeStyleBuilder<B> {
                             .last()
                             .is_some_and(|c| c.is_ascii_whitespace()))
                 {
-                    span_text = span_text.trim_start();
+                    span_text = span_text.trim_start_matches(|c: char| c.is_ascii_whitespace());
                 }
                 if is_span_last {
-                    span_text = span_text.trim_end();
+                    span_text = span_text.trim_end_matches(|c: char| c.is_ascii_whitespace());
                 }
 
                 // Collapse spaces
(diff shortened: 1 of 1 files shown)


tomcur (approved):
This looks correct to me, thanks!

For posterity: [CSS Text 3 § 4.1](https://www.w3.org/TR/css-text-3/#white-space-rules) says spaces (`U+0020`), tabs (`U+0009`) and "segment breaks" (`\n`) are collapsible. Carriage returns are considered equal to spaces by the spec.

`char::is_ascii_whitespace` further includes form feeds (`U+000C`), but that's probably benign and matches what the other code was already doing anyway.

PR #735, opened 2026-08-06
Simplify whitespace collapsing code somewhat

tomcur:
This uses `str::trim_ascii_{start,end}` instead of the manual matching and `str::ends_with`. Behaviorally and performance-wise this should be the same (in particular, `ends_with` and `chars().last()` should both just do an O(1) `next_back` under the hood).

Follow-up to https://github.com/linebender/parley/pull/734, which corrected our whitespace collapsing.


nicoburns (approved):
We may eventually end up wanting trimming using a definition that more precisely matches the CSS definitions, but this seems fine for now.

PR #736, opened 2026-08-06
Make `append_atom_to_line` private

tomcur:
This can't usefully be called externally. The user has nothing they could pass to this method.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -33,7 +33,6 @@ This release has an [MSRV] of 1.88.
   Shaped clusters that cross grapheme boundaries are represented using the existing `Cluster::is_ligature_start` and `Cluster::is_ligature_continuation`; note these methods previously encoded graphemes as well.
   Shaped clusters' advances are split evenly over the grapheme clusters they overlap.
   `Run::cluster_range` now returns grapheme cluster indices relative to the run's shaped run.
-  `BreakerState::append_cluster_to_line` was replaced by `BreakerState::append_atom_to_line`.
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -33,7 +33,6 @@ This release has an [MSRV] of 1.88.
   Shaped clusters that cross grapheme boundaries are represented using the existing `Cluster::is_ligature_start` and `Cluster::is_ligature_continuation`; note these methods previously encoded graphemes as well.
   Shaped clusters' advances are split evenly over the grapheme clusters they overlap.
   `Run::cluster_range` now returns grapheme cluster indices relative to the run's shaped run.
-  `BreakerState::append_cluster_to_line` was replaced by `BreakerState::append_atom_to_line`.
```
tomcur, line 36: I've removed this part of the changelog entry from #715, as I think effectively no-one called this method before #715 either: it already was pointless and probably would just corrupt state.

nicoburns (approved):
I'm not using `append_cluster_to_line`, but Blitz is using `BreakerState::append_inline_box_to_line`. Hopefully that's still available.

tomcur:
> but Blitz is using BreakerState::append_inline_box_to_line. Hopefully that's still available

Yep!

DJMcNab:
It failed due to GitHub's incident today, but I realise that you should probably merge it even still Tom - I don't have a need to get this in that couple of hours later, and I haven't actually reviewed it myself, although I'm sure it's fine.

PR #737, opened 2026-08-07
Add benchmarks for word/letter spacing and repeated justification

tomcur:
LLM Contributions: Mostly generated.

This adds two benchmarks, motivated by a PR that's following.

One benchmark exercises layout when letter and word spacing are applied, which (especially in the PR that follows) does some additional work relative to the zero-spacing path.

The other benchmark exercises performance of re-justifying (similar to what was proposed in https://github.com/linebender/parley/pull/707).

**Changelog: None**

### DIFF
--- a/parley_bench/src/benches.rs
+++ b/parley_bench/src/benches.rs
@@ -47,6 +47,100 @@ pub fn defaults() -> Vec<Benchmark> {
         .collect()
 }
 
+/// Benchmark for nonzero word and letter spacing.
+pub fn spacing() -> Vec<Benchmark> {
+    const DISPLAY_SCALE: f32 = 1.0;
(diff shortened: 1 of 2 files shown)


parley_bench/src/benches.rs
```diff
@@ -47,6 +47,100 @@ pub fn defaults() -> Vec<Benchmark> {
         .collect()
 }
 
+/// Benchmark for nonzero word and letter spacing.
+pub fn spacing() -> Vec<Benchmark> {
```
tomcur, line 51: There's probably an argument for folding this on into the `styled` one.

(In fact, the `styled` one currently only adjusts styles that have no impact on itemization at all.)

taj-p (approved):

PR #740, opened 2026-08-08
Set shaping context across run boundaries

tomcur:
LLM Contributions: Investigation, generated initial test.

`harfrust` requires context outside of the run it's shaping to select correct forms. This now gives it that context.

The added test shapes the same string twice (one on each line), with the second character of the first line in a larger font size to create run boundaries. `parley` renders this as follows.

| Before | Now |
| --- | --- |
| shaping_context_across_items-0 | shaping_context_across_items-0 |

Blink also handles this correctly, but Gecko doesn't appear to select the correct glyphs. See what your browser does: https://developer.mozilla.org/en-US/play?id=YMnuttLOy0Ntg0XCzwIS%2F8Lf9OFabMHGOnadV9C8hAPXRUus2lzZZN0uaO9foK796qZoE0WdZF3ArOnt.

**Changelog**

> ### Added
>
> #### Parley
>
> - Shaping across style boundaries now uses the surrounding context to select the correct glyphs.


behdad:
Looks good to me.

taj-p (approved):

PR #741, opened 2026-08-09
Move `char_style_indices` out of the per-item `ShapeOptions`

tomcur:
LLM Contributions: Review.

The style indices are text-wide, whereas the options are per-item since #727. Fixes the TODO from https://github.com/linebender/parley/pull/727#discussion_r3741783084.

### DIFF
--- a/parley_engine/src/shape/shaper.rs
+++ b/parley_engine/src/shape/shaper.rs
@@ -115,6 +115,12 @@ impl Shaper {
         shaped_text.reserve(text.len());
 
         let char_count = analysis.char_info().len();
+        debug_assert_eq!(
+            char_style_indices.len(),
+            char_count,
+            "The number of character style indices must be equal to the character count"
+        );
+
         let mut previous_item_end = 0;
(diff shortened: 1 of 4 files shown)


parley_engine/src/shape/shaper.rs
```diff
@@ -92,6 +87,10 @@ impl Shaper {
     /// split on properties like shaping-relevant style changes (e.g., font size) or properties like
     /// language.
     ///
+    /// `char_style_indices` holds per-character style indices; these are copied onto
+    /// [`Character`][super::Character`] and [`ShapedCluster`][super::ShapedCluster]. A shaped
+    /// cluster's style index is that of its logically first constituent character.
```
tomcur, line 92: There's now an argument for dropping `style_indices` from `parley_engine` completely, as it doesn't actually read them. Since #727, the full text is shaped, so character indices exposed by `parley_engine` can be used by users to index into some per-character slice directly.

nicoburns (approved):

PR #743, opened 2026-08-12
Preserve (don't clamp) negative half-leading

nicoburns:
### Summary

When `line-height < ascent + descent`, half-leading is legitimately negative:

- Extents now default to `f32::NEG_INFINITY` so legitimately negative extents are preserved.
- Lines with no extent contributions are resolved to zero in finish_line.
- Out-of-flow inline boxes no longer contribute extents. 

### LLM Contributions

Code generated with Devin Ultra.

### Changelog Entry

> ### Fixed
>
> - Line boxes with negative half-leading (line-height smaller than ascent + descent) are no longer floored at the content height; out-of-flow inline boxes no longer floor a line's extents at zero.

### DIFF
--- a/parley/src/layout/line_break.rs
+++ b/parley/src/layout/line_break.rs
@@ -97,14 +97,29 @@ struct Extents {
 
 impl Default for Extents {
     fn default() -> Self {
+        // Initialized to negative infinity so that content with negative extents (e.g. text with
+        // negative half-leading from a small `line-height`) is not floored at zero. Lines with no
+        // contributing content resolve to zero via [`Self::or_zero`], keeping the sentinel
+        // internal.
+        //
+        // Ideally, a line's initial extents should be sourced from the primary font.
         Self {
-            // NOTE: these should be `f32::NEG_INFINITY`, but that's currently causing
-            // `tests::test_builders::builders_empty` to fail with a `NaN`.
-            //
-            // And even more ideally, a line's initial extents should be sourced from the primary
-            // font.
-            over: 0.,
-            under: 0.,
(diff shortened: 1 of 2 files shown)


DJMcNab (commented):
I believe this justifies a new test. Otherwise, this looks good to me.

If I didn't know that this code (and therefore its comments) will be changing again for proper strut support, I'd be more critical of the new comments - I'm not a huge fan of them.

parley/src/layout/line_break.rs
```diff
@@ -101,14 +97,29 @@ struct Extents {
 
 impl Default for Extents {
     fn default() -> Self {
+        // Initialized to negative infinity so that content with negative extents (e.g. text with
+        // negative half-leading from a small `line-height`) is not floored at zero. Lines with no
+        // contributing content resolve to zero via [`Self::or_zero`], keeping the sentinel
+        // internal.
+        //
+        // Ideally, a line's initial extents should be sourced from the primary font.
```
tomcur:
```suggestion
        // Content with small line heights and/or with inline boxes sitting entirely above or
        // below the baseline can cause the baseline to fall outside the line box; i.e., extents
        // can be negative. Hence, we initialize the space over and under the line box's
        // baseline to negative infinity. Lines with no contributing content resolve to zero via
        // [`Self::or_zero`], keeping the sentinel internal.
        //
        // Ideally, a line's initial extents should be sourced from the primary font.
```

tomcur (approved):
LGTM! One comment suggestion to explain more clearly when the extents are negative.

PR #747, opened 2026-08-14
Add rich text editor API and example

geom3trik:
LLM Contributions: LLM used to generate the example (based on previous editor example) and some parts of the `RichEditor` (thoroughly checked by comparison with `PlainEditor`).

Adds a new `RichEditor`, the rich text counterpart to `PlainEditor`, and an example demonstrating its use (based on the editor example).

**Changelog**

> ### Added
>
> - New `RichEditor` and `RichEditorDriver`.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -42,6 +42,20 @@ dependencies = [
 ]
 
 [[package]]
+name = "accesskit_atspi_common"
+version = "0.19.1"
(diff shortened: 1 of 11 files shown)


geom3trik:
Just a thought, if it were decided that a rich editor should instead replace the plain editor, since the latter is a subset of the former, I could refractor this PR to modify the existing plain editor and example and the diff would be a lot smaller and easier to review.

PR #749, opened 2026-08-15
Honour region subtags

### LINKED ISSUE #748: Shaping drops script/region subtags, so zh-Hans and zh-Hant select the same OpenType langsys
Shaping currently feeds HarfRust only the primary language subtag of `parlance::Language`:

```rust
// parley_engine/src/shape/shaper.rs (also parley 0.11 `src/shape/mod.rs`)
let language = options
    .language
    .as_ref()
    .and_then(|lang| lang.language().parse::<harfrust::Language>().ok());
```

So `zh-Hans`, `zh-Hant`, `zh-TW`, and `zh-HK` all become `"zh"`. HarfRust maps bare `"zh"` to OpenType `ZHS`, and Traditional Chinese `locl` / `ZHT` / `ZHH` forms are never selected.

This is not a HarfRust limitation. Its own tests already distinguish the tags:

- `"zh"` / `"zh-Hans"` → `ZHS`
- `"zh-Hant"` / `"zh-tw"` → `ZHT`
- `"zh-HK"` → `ZHH`

Font fallback is unaffected: `FontSelector` still passes the full `Language` into `fontique::FallbackKey`, and `canonical_locale` treats `Hant` / `TW` / `HK` correctly.

## Suggested fix

Parse the canonical tag instead of the primary subtag:

```rust
lang.as_str().parse::<harfrust::Language>().ok()
```

`Language::as_str()` is already `language[-Script][-REGION]`, which is what HarfRust's `Language` parser expects.

taj-p:
LLM Contributions: Regression testing, CI changes (GPT 5.5 Sol).

Fixes https://github.com/linebender/parley/issues/748.

| Before | After |
| ------ | ------ |
| Subtag not honoured (same glyphs throughout) | Subtag honoured |
|  before_subtag |  after_subtag |

**Changelog**

> ### Fixed
>
> - Honours script and region sub tags for correct shaping of, for example, zh-Hans and zh-Hant.


taj-p:
If someone approves, please merge yourself. If you cLGTM with tests removed, I'll remove tests and then merge.

nicoburns:
I'm personally not a fan of the "create a custom font to test" pattern. If we're going to test this, I'd much rather we find some real fonts to test with. That costs a little repo size, but it makes for a more realistic test and less code to maintain. If we wanted to keep sizes down then we could always subset those fonts.

taj-p:
> I'm personally not a fan of the "create a custom font to test" pattern. If we're going to test this, I'd much rather we find some real fonts to test with. That costs a little repo size, but it makes for a more realistic test and less code to maintain. If we wanted to keep sizes down then we could always subset those fonts.

That's fine by me - preferring realism over non-native speakers being able to easily determine correctness is a choice I am happy with. I'll update the tests accordingly.

nicoburns (approved):

DJMcNab:
Sticking a 15MB font file into the repo is at the very least a surprising choice to make without discussion (e.g. at text office hours). It's not the end of the world, but personally I'm inclined to force push a revert; this is a repository size cost we would otherwise have to pay until the end of time.

This is under the assumption that the current font is not subsetted. If this is subsetted, it seems crazy to me that 15MB is the minimal size, but I don't know if there's some table which can't actually be subsetted in play here.

petamoriken:
Since Web Platform Tests subsets the Noto Font and uses it for testing, I think it is probably possible to reduce the file size.
https://github.com/web-platform-tests/wpt/tree/0cc6a7e1904a69dae98c86412b15bc81256279f8/fonts/noto/cjk

taj-p:
> Sticking a 15MB font file into the repo is at the very least a surprising choice to make without discussion (e.g. at text office hours). It's not the end of the world, but personally I'm inclined to force push a revert; this is a repository size cost we would otherwise have to pay until the end of time.
> 
> This is under the assumption that the current font is not subsetted. If this is subsetted, it seems crazy to me that 15MB is the minimal size, but I don't know if there's some table which can't actually be subsetted in play here.

I'm happy if you want to force push a revert and I can add this PR back with a subset. Maybe worth force pushing earlier than later?

DJMcNab:
Done, this has now been force-push reverted.

petamoriken:
Is this reverted PR being worked on again?

DJMcNab:
Re-opened as #756.

PR #751, opened 2026-08-18
Add documentation for running WPT tests against Parley via Blitz

nicoburns:
Documents how to run WPT tests against Parley with Blitz.

LLM Contributions: An initial draft was generated by LLM. An editing pass has been done.


DJMcNab (changes requested):
We spoke about this in office hours yesterday. I think that documenting this is fine, but the actual running docs should just point to a real doc in Blitz.
So this should be approximately:
Clone Blitz, replace parley with your local version, follow *this document* from Blitz's repo.

nicoburns:
@DJMcNab Docs moved to Blitz repo, link and short paragraph added to readme here.

DJMcNab (approved):
Maybe I should read the linked doc, but!

PR #753, opened 2026-08-18
Fix infinite loop in `break_remaining` with `CustomOutOfFlow` inline boxes

### LINKED ISSUE #752: Infinite loop in BreakLines::break_remaining (0.11) — specific inline/block nesting + page CSS, layout never terminates (ok on 0.10)
### Summary

After upgrading parley **0.10.0 → 0.11.1** (no other changes), line breaking hangs indefinitely on a specific inline/block fragment from baidu.com's homepage. `break_all_lines`'s `while self.break_next().is_some() {}` never makes progress: the worker thread spins at 100% CPU forever (verified still spinning after 20+ minutes).

Found via [DioxusLabs/blitz](https://github.com/DioxusLabs/blitz) (we render server-side screenshots through blitz). Bisected cleanly against blitz commits:

| blitz commit | parley | result |
|---|---|---|
| `2fa6434d` (#724) | 0.10.0 | renders the page in ~4s ✅ |
| `b96c9c24` (#726, only #725 in between — a pure dependency bump, zero blitz code change) | 0.11.1 | never finishes, 100% CPU, still spinning at 150s+ ❌ |

### Reproducer

Full single-file HTML (270KB — a 1.4KB body fragment plus the page's original `<head>` CSS): https://gist.github.com/yinnho/aaf7b1b3e0089aab2275924cba4a66bb

Rendered through blitz `HtmlDocument::from_html` + `document.resolve()` at 1280x800 viewport → never returns. The triggering fragment is baidu's IME menu:

```html
<span class="tools"><span id="mHolder"><div id="mCon"><span>
                                    输入法</span></div><ul id="mMenu"><li><a href="javascript:;">
                                    手写</a></li>...</ul></span></span>
```

(block `<div>`/`<ul>` nested inside inline `<span>`s; `#mMenu` is hidden by page CSS).

### What it is *not*

I minimized/varied the fragment — the hang survives all of these, so it is **not** text-content-dependent:

- CJK replaced with ASCII (`x` runs) → still hangs
- whitespace runs collapsed → still hangs
- isolating just the IME-menu block → still hangs
- the `<input type=submit>` next to it alone → renders fine

Synthetic rebuilds of the same DOM shape with simplified CSS do **not** hang, so some concrete CSS constraint from the page (width/height/font on `.tools`/`#mHolder`/`#mCon`/`#mMenu`) is part of the trigger. I did not minimize the CSS further.

### Stack (macOS aarch64, release build, `sample` output)

```
blitz_dom::layout::inline::compute_inline_layout_inner
parley::layout::layout::Layout<B>::break_all_lines
parley::layout::layout::Layout<B>::break_all_lines        (recursive)
parley::layout::line_break::BreakLines<B>::break_remaining
parley::layout::line_break::BreakLines<B>::break_next_line_or_box
parley::layout::line_break::BreakerState::clone           (heavy sampling weight here)
```

### Environment

- parley 0.11.1 (crates.io), fontique 0.11.1
- caller: blitz `b96c9c24` (packages/blitz-dom inline layout)
- macOS 15 aarch64 (release build); originally observed on the full baidu.com DOM (686KB after JS rendering)

Happy to run further minimization or test a fix branch if useful.

nicoburns:
Fix infinite loop in `break_remaining` with `CustomOutOfFlow` inline boxes. `break_next` yields on `CustomOutOfFlow`, boxes without advancing past them, so if `break_remaining` ignores these yields (as it currently does), it will loop forever with `break_next` continually returning the same box.

Fix: Advance past the box as zero-sized out-of-flow content instead.

Fixes #752

LLM Contributions: Fix generated with Devin Ultra.

No changelog entry because this will get backported as 0.11.2. Changelog added on v0.11.x branch and then forward-ported to `main`.

### DIFF
--- a/parley/src/layout/line_break.rs
+++ b/parley/src/layout/line_break.rs
@@ -1085,10 +1085,12 @@ impl<'a, B: Brush> BreakLines<'a, B> {
         self.state.layout_max_advance = max_advance;
         self.state.line_max_advance = max_advance;
         while let Some(yield_data) = self.break_next() {
-            // `break_next` yields on `CustomOutOfFlow` boxes without advancing past them,
-            // leaving placement (and advancing the breaker state) to the caller. Here there
-            // is no custom placement logic, so treat the box as zero-sized out-of-flow
-            // content and advance past it to guarantee progress.
+            // When `break_next` encounters a `CustomOutOfFlow`, it yields a `YieldData::InlineBoxBreak`
+            // without advancing the item iteration past the box. This allows embedders to implement custom
+            // placement logic to place the box. However, it means that we must also handle placing the box
+            // and advancing the iteration here to avoid an infinite loop.
+            //
+            // So we place the box as a zero-sized out-of-flow box to guarantee progress.
             if let YieldData::InlineBoxBreak(_) = yield_data {
                 self.state.append_inline_box_to_line(
                     self.state.line.x,
--- a/parley/src/layout/line_break.rs
+++ b/parley/src/layout/line_break.rs
@@ -1084,7 +1084,20 @@ impl<'a, B: Brush> BreakLines<'a, B> {
         // println!("\nBREAK ALL");
         self.state.layout_max_advance = max_advance;
         self.state.line_max_advance = max_advance;
-        while self.break_next().is_some() {}
+        while let Some(yield_data) = self.break_next() {
+            // `break_next` yields on `CustomOutOfFlow` boxes without advancing past them,
+            // leaving placement (and advancing the breaker state) to the caller. Here there
+            // is no custom placement logic, so treat the box as zero-sized out-of-flow
+            // content and advance past it to guarantee progress.
+            if let YieldData::InlineBoxBreak(_) = yield_data {
+                self.state.append_inline_box_to_line(
+                    self.state.line.x,
+                    0.0,
+                    0.0,
+                    self.layout.data.quantize,
+                );
+            }
+        }
         self.finish();
     }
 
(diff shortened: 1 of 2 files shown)


parley_tests/tests/issues.rs
```diff
@@ -114,3 +117,30 @@ Third line that ends with newlines\n\n";
     env.with_name("max_context_with_mandatory_breaks")
         .check_layout_snapshot(&layout);
 }
+
+/// Test that `break_all_lines` terminates when the layout contains a
+/// `CustomOutOfFlow` inline box.
+///
+/// `break_next` yields on `CustomOutOfFlow` boxes without advancing past them,
+/// so `break_remaining` must advance past the box itself or it will loop forever.
```
taj-p:
```suggestion
```

parley/src/layout/line_break.rs
```diff
@@ -1084,7 +1084,20 @@ impl<'a, B: Brush> BreakLines<'a, B> {
         // println!("\nBREAK ALL");
         self.state.layout_max_advance = max_advance;
         self.state.line_max_advance = max_advance;
-        while self.break_next().is_some() {}
+        while let Some(yield_data) = self.break_next() {
+            // `break_next` yields on `CustomOutOfFlow` boxes without advancing past them,
+            // leaving placement (and advancing the breaker state) to the caller. Here there
+            // is no custom placement logic, so treat the box as zero-sized out-of-flow
+            // content and advance past it to guarantee progress.
```
taj-p: Is this trying to say that `InlineBoxBreak` is a special case where the caller just expects us to advance past inline box breaks, so we do it for them here? [super nit]: Could you please update this documentation - it's a little hard to parse

taj-p (approved):

PR #757, opened 2026-08-26
Simplify template

DJMcNab:
LLM Contributions: None.

Based on feedback in [#parley > Parley PR template](https://xi.zulipchat.com/#narrow/channel/205635-parley/topic/Parley.20PR.20template/with/616079356), the current template is too long.
People also seem unable to understand the syntax of using square brackets for optional arguments. I had been under the impression that this was common, but I've not observed anyone actually use it correctly. As such, I've removed it from the template.

Additionally, people using LLM tools to open their PRs is currently indistinguishable from people using e.g. IDEs or other tools which don't expose it. I'm hinting towards this template in the CHANGELOG file to at least have a chance for LLMs to follow this (I've observed LLMs having a drive to make CHANGELOG entries).

**Changelog: None**

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,25 +1,17 @@
-<!--
-Thanks for opening a PR for this Linebender project!
--->
+<!-- Please ensure that you have reviewed our LLM ("AI") policy at https://linebender.org/wiki/llm-policy/.
 
-<!-- If you used any LLM ("AI") tools in making this PR, please ensure that you have reviewed our
-LLM policy at https://linebender.org/wiki/llm-policy/.
-Please list how you used it (e.g. writing code, reviewing, fact finding) below.
-We expect you to have fully understood and self-reviewed your contributions to Linebender.
-
-If you did not use any LLM tools, please replace "Unspecified" with 'None'. -->
+If you did not use any LLM tools, please replace `Unspecified` with `None`. -->
 LLM Contributions: Unspecified.
 
-_Describe what problem this is solving, and how it's solved. Use your own words - descriptions authored by LLMs are not allowed._
+_Use your own words - descriptions authored by LLMs are not allowed._
 
 <!--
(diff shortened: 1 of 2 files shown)


nicoburns (approved):
Seems like an improvement (much less verbose! :)). I would recommend moving LLM Contributions to the bottom (or just above changelog).

I generally don't like the changelog section. Seems fiddly. And I generally consider changelog maintenance a maintainer activity rather than a contributor activity. I would pretty much never want to block a PR landing on changelog.

DJMcNab:
> I would recommend moving LLM Contributions to the bottom (or just above changelog).

Could you expand on this? The current position is aiming to ensure that reviewers can have the context about how likely the rest of the pr is to be nonsense. I guess maybe your thinking is that the "use your own words" should hopefully cover that?
It might be that this field is just a "how respectful of our time will you be" filter, and so prs which ignore it can slide down the queue. And as such, we can have it further down anyway?

> I generally consider changelog maintenance a maintainer activity rather than a contributor activity.

My experience with doing changelog maintenance is that there is no easy heuristic for "does this pr need a changelog" (or more so, there's no way for a PR to explicitly say "this doesn't need a changelog"). This means that it's easy for prs which really should have had changelog entries to just not. The main idea of this scheme is that people can opt into writing a changelog in a low friction way (without fighting merge conflicts).
I also don't trust LLMs to write changelog entries due to their lacking theory of mind.

(As an aside - I wonder if there's a world where we ask the approver to write a changelog entry? One problem is that I don't see how that could be "enforced" (that is, remembered, as we probably wouldn't want to hard block on it).)

> I would pretty much never want to block a PR landing on changelog.

Fwiw, this whole scheme is trying to lower the amount of work that does get blocked on the changelog.

I'd like to keep the experiment going for now. But I'll admit, it requires me finding more time to work on Gazeto.

DJMcNab:
We can make follow ups later.

PR #760, opened 2026-08-31
Correctly justify with consecutive trailing spaces

tomcur:
LLM Contributions: Investigation, test.

When justifying a line, this now correctly handles runs of consecutive trailing spaces. This is follow-up to https://github.com/linebender/parley/pull/738#discussion_r3738289581.

There's some more that we could/should do here (separately!). E.g., I think we simply do not need to check the break reason, and that would also get us to handle `BreakReason::Emergency` correctly for free. We also don't currently handle consecutive trailing spaces correctly for either justification or hanging when there's a formatting change inside that run; something like the following.

```html
<div style="background: #efefef; width: 95px; white-space-collapse: preserve; text-align: justify;">AA BB CC  <span style="font-size: 2em;"> </span>DD</div>
```

**Changelog**

> ### Fixed
>
> - Lines aligned with justification now correctly handle runs of consecutive trailing spaces.


tomcur (commented):
Without the fix, the snapshot looks as follows.

Image

nicoburns (approved):

parley_tests/tests/basic.rs
```diff
@@ -631,6 +631,27 @@ fn justify_consecutive_spaces() {
...
+    let builder = env.ranged_builder(text);
+
+    // TODO: this should also correctly hang the whitespace when the spaces are not in the same run (e.g.,
+    // because of a font or font size change). However, we don't currently handle that correctly.
+    //
+    // builder.push(StyleProperty::FontSize(15.9), 10..11);
+    let mut layout = builder.build(text);
+    layout.break_all_lines(Some(max_advance));
+    layout.align(Alignment::Justify, AlignmentOptions::default());
+    env.check_cluster_snapshot(&layout, text, 12.0);
```
DJMcNab, line 652: I didn't know this existed!

DJMcNab (commented):
Great to see this fixed. We've been carrying a patch for this for quite a while 😅

PR #761, opened 2026-09-01
Fix LineHeight not forcing a new itemizer item

alexeygolev:
LLM Contributions: None.

For ranges with only the `StyleProperty::LineHeight` being different from the "current" style during `shape_text` shape-relevant style properties change splitting a new itemizer item wasn't created so line height change was being lost in the shaping result.

I noticed that I can't get "spans" of text with the line height that is different from the base style to actually have that line height. It was actually worse in the current published version (there seems to be an off by one bug concerning the line heights that prevented the correct line height application to spans even when accompanied by other style properties change). I looked through the codebase and through documentation/comments and couldn't find anything that can hint at why line height wasn't in "shaping-relevant style properties" set. I checked how it works in CSS just to make sure I'm not going crazy and it works as I'd expected it to work.

It's a tiny change. I added a test to make sure it works the way I think it should. The `builders_style_runs_match_ranged` can't capture this because it compares builders outputs and all builders ultimately use the same item-splitting code.

**Changelog**

> ### Added
>
> - LineHeight change is now respected as shaping-relevant style property change.

### DIFF
--- a/parley/src/shape/mod.rs
+++ b/parley/src/shape/mod.rs
@@ -155,7 +155,8 @@ pub(crate) fn shape_text<'a, B: Brush>(
                         || style.font_variations != item_style.font_variations
                         || style.font_features != item_style.font_features
                         || !nearly_eq(style.letter_spacing, item_style.letter_spacing)
-                        || !nearly_eq(style.word_spacing, item_style.word_spacing);
+                        || !nearly_eq(style.word_spacing, item_style.word_spacing)
+                        || !style.line_height.nearly_eq(item_style.line_height);
                 }
 
                 if split {
--- a/parley/src/tests/test_builders.rs
+++ b/parley/src/tests/test_builders.rs
@@ -685,3 +685,30 @@ fn builders_crlf_across_run_boundary_counts_as_single_line_break() {
         "styled CRLF should match styled LF line count"
     );
 }
+
+/// Test that a `LineHeight` difference alone forces a new itemizer item.
+#[test]
+fn line_height_alone_forces_a_new_item() {
+    let text = "Hello world";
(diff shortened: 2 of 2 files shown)


nicoburns:
This seems to be not quite the right fix (and causes regressions in Blitz WPT tests). The issue is the run boundaries are also shaping boundaries, but line-height shouldn't cause a shaping boundary. My LLM suggests change line-height to resolve per-atom instead of per-run instead.

@tomcur do you have thoughts?

tomcur:
> The issue is the run boundaries are also shaping boundaries, but line-height shouldn't cause a shaping boundary.

That's correct: e.g., Gecko and Blink also don't treat line height as shaping boundaries (graphemes form across line height changes).

In general we probably should do a pass over the granularity at which `parley` reads styles. Reading, e.g., font size at the level at which items were split is correct, because we use `nearly_eq` to decide whether to split, but color could be read at the `ShapedCluster` level. Line height perhaps could also be at the `ShapedCluster` level, but perhaps the `Atom` level is more straightforward because of assumptions around selection and graphemes.

nicoburns:
> color could be read at the ShapedCluster level

What is Parley reading `color` for? I thought that was just passthrough...

tomcur:
> What is Parley reading `color` for? I thought that was just passthrough...

I don't think `parley` is reading such things, but a consumer may be impacted by the granularity at which `parley` reports styles.

alexeygolev:
Happy to work on this more if you see a viable direction. I'm using my fork in a project for now but it would be nice to have this done in a way that makes sense.

tomcur:
Please see #839.

PR #764, opened 2026-09-03
Make glyph run iteration fast again

### LINKED ISSUE #763: 20x regression in `GlyphRunIter::next` performance
`GlyphRunIter::next` is 20x slower on `main` than in `0.11` (causes an end-to-end Blitz layout to be 3-4x slower!)

There are two issues:

- `Cluster` has gotten much larger and a lot of time is being spent in `memcpy`
- `next()` has O(n²) time complexity as it calls `.skip()` on the output of `.flatMap()` which causes it to perform a linear scan from the beginning of the list at each call (as there is no random access to the output of flatMap)

These issues compound as every cluster in the linear scan is copied at every iteration!

nicoburns:
Fixes: https://github.com/linebender/parley/issues/763 (20x regression in glyph iteration performance)

- Replaces `.flatMap().skip()` iteration with a custom iterator
- The new iterator:
   - Iterates over atoms, keeping 1 peeked atom in state so it can compare runs between the current and next atom
   - Internally handles ltr vs rtl iteration (benchmarked as being faster due to less code bloat)
   - Iterates by reference to avoid excessive memory copies

**LLM Contributions**: Generated with Fable 5.1 Low (which also identified the issue and benchmarked)

**Changelog**: None (fixes a regression that has never been published)

### DIFF
--- a/parley/src/layout/line.rs
+++ b/parley/src/layout/line.rs
@@ -11,7 +11,7 @@ use crate::style::Brush;
 use crate::{InlineBox, InlineBoxKind};
 
 use core::ops::Range;
-use parley_engine::{Atom, Atoms, Glyph, ShapedSlice};
+use parley_engine::{Atom, Atoms, Glyph};
 
 /// Line in a text layout.
 #[derive(Copy, Clone)]
@@ -298,14 +298,14 @@ impl<'a, B: Brush> GlyphRun<'a, B> {
     }
 }
 
-/// Cursor over the atoms of a run in visual order, resumable across [`GlyphRunIter::next`] calls.
+/// Peekable iterator over the atoms of a run in visual order, resumable across
+/// [`GlyphRunIter::next`] calls.
 ///
-/// This walks the run's shaped data ([`ShapedSlice`]/[`Atom`]) directly rather than
-/// materialising [`Cluster`](crate::Cluster)s and their composed glyph iterators. It computes
-/// the same glyph counts and advances as [`Run::glyphs_in`] over the same atoms.
+/// This walks the run's shaped data ([`Atom`]s) directly rather than materialising
+/// [`Cluster`](crate::Cluster)s and their composed glyph iterators. It computes the same glyph
+/// counts and advances as [`Run::glyphs_in`] over the same atoms.
 #[derive(Clone)]
-struct AtomCursor<'a> {
-    slice: ShapedSlice<'a>,
+struct AtomIter<'a> {
     atoms: Atoms<'a>,
     spacing: EffectiveSpacing,
     is_rtl: bool,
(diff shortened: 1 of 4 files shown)


parley_bench/src/benches.rs
```diff
@@ -172,39 +205,45 @@ pub fn styled() -> Vec<Benchmark> {
...
+        .collect()
+}
 
-                        with_contexts(|font_cx, layout_cx| {
-                            let mut builder =
-                                layout_cx.ranged_builder(font_cx, text, DISPLAY_SCALE, QUANTIZE);
-                            builder.push_default(FontFamily::from(FONT_FAMILY_LIST));
+/// Benchmark for iterating the positioned glyph runs and glyphs of a styled layout, as a renderer
+/// would.
+pub fn styled_items() -> Vec<Benchmark> {
```
tomcur: Nice! Nit: maybe the name should change to something like `iterate_styled_items`, so it's more immediately obvious this benches iteration of items rather than building a layout.

parley/src/layout/line.rs
```diff
@@ -296,10 +298,81 @@ impl<'a, B: Brush> GlyphRun<'a, B> {
...
+            for glyph in self.slice.shaped_cluster_glyphs(cluster_idx) {
+                glyph_count += 1;
+                advance += glyph.advance;
+            }
+        }
+        if glyph_count != 0 && !self.spacing.is_zero() {
+            advance += self.spacing.gaps(atom).total();
+        }
+        (glyph_count, advance)
+    }
```
tomcur: I believe this can use `Atom::advance` directly, so this doesn't have to sum manually.

```suggestion
    fn glyphs(&self, atom: &Atom<'a>) -> (usize, f32) {
        let glyph_count: usize = atom
            .shaped_clusters()
            .iter()
            .map(|cluster| usize::from(cluster.glyph_len()))
            .sum();
        let mut advance = atom.advance();
        if glyph_count != 0 && !self.spacing.is_zero() {
            advance += self.spacing.gaps(atom).total();
        }
        (glyph_count, advance)
    }
```

parley/src/layout/line.rs
```diff
@@ -296,10 +298,81 @@ impl<'a, B: Brush> GlyphRun<'a, B> {
...
+                self.atoms.next()
+            };
+        }
+        self.peeked
+    }
+
+    #[inline]
+    fn advance(&mut self) {
+        self.peeked = None;
+    }
```
tomcur, line 348: This name clashes with how `advance` is used elsewhere. It also doesn't quite advance on its own. Maybe `consume`?

```suggestion
    /// Consume the atom returned by [`Self::peek`], so the next peek moves on.
    #[inline]
    fn consume(&mut self) {
        self.peeked = None;
    }
```

parley/src/layout/line.rs
tomcur: Additionally, the method is named `glyphs` but doesn't return any glyphs; perhaps it should be renamed to something like `measure`.

tomcur (approved):
Nice catch! Really good that this now tests glyph iteration in the benchmarks, too.

The implementation mostly looks good to me, though while the struct is called `AtomCursor`, it doesn't feel very cursor-like to me. It reads more like a peekable iterator (and note that `Atoms` itself is a cursor). Nothing really blocking here, but I have some suggestions inline.

I'm conflicted on the added test. I think *nearly* all failures that it can catch will have snapshots failing anyway, leaving quite some test code that we now need to keep in sync only for very specific failure modes.

taj-p:
Yay!! Thanks Nico 🙏 🚀

PR #769, opened 2026-09-04
Expand glyph benchmark to check itemizing vs non-itemizing styles

tomcur:
LLM Contributions: Mostly generated, with a manual pass especially over the docs.

Styles that affect fonts affect itemization, and cause splitting into separately shaped runs. Other styles like underline only affect the style spans.

These have different performance characteristics, so this expands the benchmark to also bench both in isolation. This also adds a "uniform" Latin bench, where there are no styles at all; that'll be a common case and effectively is about raw glyph throughput.

This PR is motivated by iteration of glyphs over style spans being somewhat slow, so I'm preparing two follow-up PRs.

**Changelog: None**

### DIFF
--- a/parley_bench/src/benches.rs
+++ b/parley_bench/src/benches.rs
@@ -219,15 +219,12 @@ pub fn styled() -> Vec<Benchmark> {
 ///
 /// The cases fall into three groups:
 ///
-/// - Latin with uniform styling, giving one glyph run per shaped run, each as long as the shaped
-///   run.
-/// - Latin with a style alternating every few characters. Glyph runs break wherever the style
-///   index changes, but only a subset of style properties also splits the shaped run: underline
-///   leaves the text in one shaped run carrying many glyph runs, whereas bold selects a different
-///   font and so gives every glyph run a shaped run of its own.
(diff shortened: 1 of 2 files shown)


parley_bench/src/benches.rs
```diff
@@ -215,40 +216,145 @@ pub fn styled() -> Vec<Benchmark> {
...
+        },
+    )];
+
+    // Underline splits the glyph runs only; bold also splits the shaped runs.
+    for (label, chunk_len, style) in [
+        ("non-splitting", 1, StyleProperty::Underline(true)),
+        ("non-splitting", 16, StyleProperty::Underline(true)),
+        ("splitting", 1, StyleProperty::FontWeight(FontWeight::BOLD)),
+        ("splitting", 16, StyleProperty::FontWeight(FontWeight::BOLD)),
+    ] {
```
tomcur, line 251: On re-reading, perhaps there are better names... these could also be "non-itemizing" vs "itemizing" (but also see https://github.com/linebender/parley/pull/727#discussion_r3741774966).

taj-p (approved):

PR #771, opened 2026-09-04
perf: Write an explicit glyph iterator for `Run::glyphs_in` instead of composing

tomcur:
LLM Contributions: Mostly generated.

The old `Run::glyphs_in` builds quite the tower of composed iterators. This instead makes the iterator a much smaller type, makes it nameable, and improves performance nicely.

# Performance

Together with the #770, this benches as follows. See #770 for some more detailed numbers, including a comparison against v0.11.0.

```
$ cargo bench --bench main -- compare ../target/benchmarks/main -t 8. --filter "*Glyph*"

Glyph Runs - latin 4 paragraph, uniform                      [   5.7 us ...   2.2 us ]     -61.12%*
Glyph Runs - latin 4 paragraph, non-splitting every 1 chars  [  61.2 us ...   7.2 us ]     -88.28%*
Glyph Runs - latin 4 paragraph, non-splitting every 16 chars [   9.5 us ...   2.6 us ]     -72.42%*
Glyph Runs - latin 4 paragraph, splitting every 1 chars      [  27.8 us ...  10.7 us ]     -61.59%*
Glyph Runs - latin 4 paragraph, splitting every 16 chars     [   7.1 us ...   2.7 us ]     -62.01%*
Glyph Runs - arabic 4 paragraph, mixed                       [  14.5 us ...   5.4 us ]     -62.61%*
Glyph Runs - latin 4 paragraph, mixed                        [   8.9 us ...   3.1 us ]     -64.90%*
Glyph Runs - japanese 4 paragraph, mixed                     [   9.1 us ...   3.4 us ]     -63.23%*
```

**Changelog: None**

### DIFF
--- a/parley_engine/src/shape/atom.rs
+++ b/parley_engine/src/shape/atom.rs
@@ -136,10 +136,7 @@ impl<'a> ShapedSlice<'a> {
     /// Get an iterator over glyphs in visual left-to-right order of the shaped cluster with the
     /// given index.
     #[inline]
-    pub fn shaped_cluster_glyphs(
-        &self,
-        cluster_idx: u32,
-    ) -> impl Iterator<Item = Glyph> + Clone + use<'a> {
+    pub fn shaped_cluster_glyphs(&self, cluster_idx: u32) -> ShapedClusterGlyphs<'a> {
         debug_assert!(
             self.clusters.0 <= cluster_idx && cluster_idx < self.clusters.1,
             "cluster index out of this slice's range"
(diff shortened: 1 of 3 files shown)


parley_engine/src/shape/atom.rs
DJMcNab, line 1: I guess the main advantage of this is just to have this `Glyphs` iterator be nameable, just like the other one?

DJMcNab (approved):
Neat trick to limit the amount of per-iteration work!

parley_engine/src/shape/atom.rs
tomcur: Yeah!

tomcur: As an aside, while doing this work I noticed the old code composing iterators was really fragile performance-wise. It depended heavily on what the compiler/LLVM was able to do, which probably was heavily related to the compiler's decisions around inlining.

PR #776, opened 2026-09-06
Add `Layout::calculate_content_widths` benchmarks 

tomcur:
LLM Contributions: Review.

The content widths function will need some work to correctly match line breaking (see, e.g., https://github.com/linebender/parley/pull/762), so let's start measuring its performance.

**Changelog: None**

### DIFF
--- a/parley_bench/benches/main.rs
+++ b/parley_bench/benches/main.rs
@@ -6,7 +6,8 @@
 use tango_bench::tango_benchmarks;
 
 use parley_bench::benches::{
-    defaults, iterate_glyph_runs, long_line, repeated_justification, spacing, styled,
+    content_widths, defaults, iterate_glyph_runs, long_line, repeated_justification, spacing,
+    styled,
 };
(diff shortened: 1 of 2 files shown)


taj-p (approved):
LGTM!

nicoburns:
Note that `calculate_content_widths` is already ~2x slower than `0.11`. LLM attributes this to it having to iterate over two arrays (characters/shaped_clusters) instead of one fused array in `0.11`.

tomcur:
> Note that `calculate_content_widths` is already ~2x slower than `0.11`. LLM attributes this to it having to iterate over two arrays (characters/shaped_clusters) instead of one fused array in `0.11`.

Ah, that's good to know, thanks! Incidentally the new https://github.com/linebender/parley/pull/785 decreases timings by roughly 40% on most cases, so may recover quite a bit of it. Seems to come from a different area, so perhaps it's possible to do better yet.

PR #777, opened 2026-09-07
Simplify spacing calculation in the line breaker

tomcur:
`EffectiveSpacing` itself already checks for the fast path.

LLM Contributions: Review.

**Changelog: None**


conor-93 (approved):
Nice, saves building `EffectiveSpacing` twice too, in the non-zero flow

DJMcNab (approved):

PR #781, opened 2026-09-08
Don't let soft break opportunities shadow mandatory breaks in complex-script runs

### LINKED ISSUE #768: Mandatory line break (U+000A) silently dropped inside Thai, Khmer and Lao runs
## Summary

A mandatory line break (`U+000A`, UAX #14 class `BK`) is **silently dropped** when it sits inside a run of Thai, Khmer or Lao. The layout returns one line where two were asked for, with no error and no warning.

Reproduces on `main` (`11711d8`) and on the published `0.11.1`, **with and without the `complex-scripts` feature**.

## Reproduction

No font is named, so this should be portable.

```rust
use parley::{FontContext, LayoutContext, StyleProperty};

fn lines(text: &str) -> usize {
    let mut fcx = FontContext::new();
    let mut lcx: LayoutContext<[u8; 4]> = LayoutContext::new();
    let mut b = lcx.ranged_builder(&mut fcx, text, 1.0, false);
    b.push_default(StyleProperty::FontSize(32.0));
    let mut l = b.build(text);
    l.break_all_lines(None);
    l.lines().count()
}

fn main() {
    for (text, label) in [
        ("AB\nC",      "latin"),
        ("กก\nกก",      "thai"),
        ("ក្ម\nក្ម",      "khmer"),
        ("ກກ\nກກ",      "lao"),
        ("日本\n語",    "japanese"),
        ("กกA\nB",      "thai, latin before the LF"),
        ("กก \nกก",     "thai, space before the LF"),
    ] {
        let (got, want) = (lines(text), text.matches('\n').count() + 1);
        println!("{:<28} {got}/{want}  {}", label, if got == want { "ok" } else { "WRONG" });
    }
}
```

### Expected

Every row `2/2`. `U+000A` is class `BK`; the break is mandatory and unconditional.

### Actual

```
latin                        2/2  ok
thai                         1/2  WRONG
khmer                        1/2  WRONG
lao                          1/2  WRONG
japanese                     2/2  ok
thai, latin before the LF    2/2  ok
thai, space before the LF    2/2  ok
```

`Line::break_reason()` on the merged line returns `BreakReason::None`, not `Explicit` — the boundary appears to be lost rather than deprioritised.

## What I ruled out

- **Not width-dependent.** `break_all_lines(None)`, `Some(f32::MAX)`, `Some(10000.0)` and `Some(984.0)` all behave identically.
- **Not font-dependent.** Reproduces with no font named, and with Ayuthaya, Thonburi, Helvetica and Arial Unicode MS.
- **Not the `complex-scripts` feature.** Identical with it on and off. (Feature activation was verified behaviourally, not by inspecting flags: an unspaced Thai paragraph in a 420 px box gives 1 line with the feature off and 5 with it on.)
- **Not the input bytes.** Byte-dumped: a literal `0x0A`, no `U+2028`.
- **Not `icu_segmenter`.** Queried directly at the same version `cargo tree -i` shows parley linking (**v2.3.0, a single shared node**), it reports the mandatory boundary correctly in every case. For `"ฉันกำลังถือแมงมุม\nตัวใหญ่ไว้ในมือ"` the boundary set contains both `51` (dictionary boundary) and `52` (mandatory, after the LF).

## Trigger

Not simply "the character before the LF" — `"กกA\nB"` and `"Aก\nB"` both pass. Thai adjacent to the LF appears **necessary but not sufficient**; a single Thai character does not trigger it, a longer run does. A space or a Latin character immediately before the LF prevents it. `\r\n` also survives, presumably because CR takes a separate rule path.

I have not diagnosed it further than reproducing from the public API. It looks like the sort of thing #676 describes — a complex-script sub-range's opportunities being spliced back over the global list and overwriting the boundary at the range's end — but that is a guess, not a diagnosis.

## Workaround

`StyleProperty::TextWrapMode(TextWrapMode::NoWrap)` produces the correct 2 lines with `BreakReason::Explicit`, split at the authored `\n`. `WordBreak::BreakAll` also works.

## Possibly why it survived

The mandatory-break tests in `src/tests/test_analysis.rs` — `test_mandatory_break_in_text` (`"ABC DEF\nG"`), `test_newline`, `test_two_newlines` — are Latin-only. Nothing appears to exercise `\n` inside a complex-script run, so a regression here would not be caught.

## Environment

- parley `main` @ `11711d8f67d3c033d07ff7b5db55021da183d1a2`, and `0.11.1` from crates.io
- `icu_segmenter` 2.3.0, `compiled_data`
- rustc 1.95.0, macOS 15 (arm64). **Not tested on another platform** — but the failing path is text analysis with no OS involvement, and the segmentation data is baked in, so I would expect it to be platform-independent.

Happy to test a patch or add the cases to the test suite if that would help.

nicoburns:
Fixes #768

LLM Contributions: Debugged and fix generated by Fable 5.1 Low

## Changes made

- In `line_break.rs`: check the `is_newline` case before the `Boundary::Line` case so a mandatory break takes precedence over a line break opportunity.
- In `analysis.rs`: drop an ICU line boundary that lands on a mandatory-break character, so `\n` always gets `Boundary::Word`. This is applied before `line_break_override`, so overrides are unaffected. 
- Regression tests.

Either fix is sufficient to fix Parley's layout, but both are applied on principle. The analysis one is required for other consumers of Parley Engine.

**Changelog**

> ### Fixed
>
> - Mandatory line breaks (\n, \r\n, U+2028, U+2029) directly following Thai, Khmer or Lao text are no longer dropped.

### DIFF
--- a/parley/src/layout/line_break.rs
+++ b/parley/src/layout/line_break.rs
@@ -757,14 +757,9 @@ impl<'a, B: Brush> BreakLines<'a, B> {
                         let text_wrap_mode = self.state.line.text_wrap_mode;
                         self.state.line.text_wrap_mode = style.text_wrap_mode;
 
-                        if boundary == Boundary::Line && text_wrap_mode == TextWrapMode::Wrap {
-                            // We don't record boundaries when the advance is 0. As we do not want overflowing content to cause extra consecutive
-                            // line breaks. We should accept the overflowing fragment in that scenario.
-                            if self.state.line.x != 0.0 {
-                                self.state.mark_line_break_opportunity();
-                                // break_opportunity = true;
-                            }
-                        } else if is_newline {
+                        // A newline always ends the line, regardless of any soft break
+                        // opportunity that analysis may have recorded on it.
+                        if is_newline {
                             if max_height_exceeded {
                                 return self.max_height_break_data(line_height);
                             }
@@ -818,6 +813,14 @@ impl<'a, B: Brush> BreakLines<'a, B> {
                                 max_advance,
                                 line_indent,
                             );
+                        } else if boundary == Boundary::Line && text_wrap_mode == TextWrapMode::Wrap
+                        {
+                            // We don't record boundaries when the advance is 0. As we do not want overflowing content to cause extra consecutive
+                            // line breaks. We should accept the overflowing fragment in that scenario.
+                            if self.state.line.x != 0.0 {
+                                self.state.mark_line_break_opportunity();
+                                // break_opportunity = true;
+                            }
                         } else if
                         // This text can contribute "emergency" line breaks.
                         style.overflow_wrap != OverflowWrap::Normal
--- a/parley/src/layout/line_break.rs
+++ b/parley/src/layout/line_break.rs
@@ -757,8 +757,7 @@ impl<'a, B: Brush> BreakLines<'a, B> {
                         let text_wrap_mode = self.state.line.text_wrap_mode;
                         self.state.line.text_wrap_mode = style.text_wrap_mode;
 
-                        // A newline always ends the line, regardless of any soft break
-                        // opportunity that analysis may have recorded on it.
+                        // Take mandatory line breaks before checking for line wrapping.
                         if is_newline {
                             if max_height_exceeded {
                                 return self.max_height_break_data(line_height);
--- a/parley/src/tests/test_analysis.rs
+++ b/parley/src/tests/test_analysis.rs
@@ -5,6 +5,7 @@ use crate::{FontContext, LayoutContext, RangedBuilder, StyleProperty, WordBreak}
 use alloc::{vec, vec::Vec};
 use fontique::FontWeight;
 use icu_properties::props::{GraphemeClusterBreak, Script};
+use icu_segmenter::{LineSegmenter, options::LineBreakOptions};
 use parley_engine::Boundary;
 
 #[derive(Default)]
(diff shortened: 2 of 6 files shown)


parley/src/layout/line_break.rs
```diff
@@ -757,14 +757,9 @@ impl<'a, B: Brush> BreakLines<'a, B> {
...
-                        if boundary == Boundary::Line && text_wrap_mode == TextWrapMode::Wrap {
-                            // We don't record boundaries when the advance is 0. As we do not want overflowing content to cause extra consecutive
-                            // line breaks. We should accept the overflowing fragment in that scenario.
-                            if self.state.line.x != 0.0 {
-                                self.state.mark_line_break_opportunity();
-                                // break_opportunity = true;
-                            }
-                        } else if is_newline {
+                        // A newline always ends the line, regardless of any soft break
+                        // opportunity that analysis may have recorded on it.
```
tomcur: This comment is quite specific to the bug being fixed. Perhaps something like the following?

```suggestion
                        // Take mandatory line breaks before checking for line wrapping.
```

parley/src/tests/test_builders.rs
```diff
@@ -685,3 +685,40 @@ fn builders_crlf_across_run_boundary_counts_as_single_line_break() {
...
+        "กก\u{2028}กก",
+        "aa\nbb",
+    ] {
+        assert_eq!(
+            break_reasons(text),
+            vec![BreakReason::Explicit, BreakReason::None],
+            "{text:?} should produce exactly two lines with an explicit break",
+        );
+    }
+}
```
tomcur, line 725: I think this test isn't actually exercising the ICU4X bug, because `set_root_style` sets `WordBreak::BreakAll` and the ICU4X segmenter doesn't seem to reproduce the bug in that case.

With the other test that was added, I think this one could just be dropped.

parley/src/tests/test_analysis.rs
```diff
@@ -308,6 +308,40 @@ fn test_paragraph_separator_is_hard_break() {
         .expect_contributes_to_shaping_list(vec![true, false, true]);
 }
 
+/// ICU4X emits a soft break opportunity at the end of a complex-script run even
+/// when it is directly followed by a mandatory break character (violating UAX #14
+/// LB6). That opportunity must be dropped so the newline is not treated as a
+/// regular wrap opportunity. See <https://github.com/linebender/parley/issues/768>.
+#[test]
+fn test_mandatory_break_after_complex_script_run() {
```
tomcur, line 317: I think it would be valuable to update this test to check whether icu4x's segmenter bug still reproduces.

A comment should then say, once it no longer reproduces (and the test fails), we should revert the `is_line = !properties.is_mandatory_linebreak()` change back to `is_line = true`.

tomcur (commented):
The fix looks good, thanks! I've left some comments about the tests.

parley/src/tests/test_builders.rs
nicoburns: Updated to actually test the bug

parley/src/tests/test_analysis.rs
nicoburns: Done!

tomcur (approved):

DJMcNab:
cc @robertbastian - We think there's possibly an icu_segmenter bug which this is working around.

robertbastian:
Try updating to the neo LineSegmenter that was released in 2.3, that fixed quite a few bugs around complex scripts.

PR #783, opened 2026-09-08
Expand `Whitespace` variants

tomcur:
LLM Contributions: Investigation, review.

This proposes letting the enum cover all of Unicode `White_Space`, with the variants based on the sets that CSS Text 4 defines. Follow-up to https://github.com/linebender/parley/pull/762#discussion_r3923722770.

According the CSS Text 4, the set of collapsible white space is `Space | Tab | Newline`,  and according to the spec hangable is effectively `collapsible ∪ (space separator \ {non-breaking space})`; but note both Gecko and Blink only hang `collapsible ∪ {ideographic space}` (also note space, non-breaking space, and ideographic space are all part of the space separators `Zs` category).

`ControlWhitespace` are those that CSS Text 4 says should just be rendered as glyphs.

**Changelog: None**

(No changelog, as while `Whitespace` was migrated from `parley`, it wasn't `pub` and so hasn't been in a release yet.)

### DIFF
--- a/parley_engine/src/shape/cluster.rs
+++ b/parley_engine/src/shape/cluster.rs
@@ -624,7 +624,7 @@ mod tests {
 
     #[test]
     fn whitespace() {
-        // These all code points from `NUL` up to the last `White_Space` character as defined in
+        // Test all code points from `NUL` up to the last `White_Space` character as defined in
         // https://www.unicode.org/Public/UCD/latest/ucd/PropList.txt.
         for c in '\u{0000}'..='\u{3000}' {
             let ws = Whitespace::from_char(c);
--- a/parley_engine/src/shape/cluster.rs
+++ b/parley_engine/src/shape/cluster.rs
@@ -130,7 +130,6 @@ impl Whitespace {
             '\u{1680}' | '\u{2000}'..='\u{200A}' | '\u{202F}' | '\u{205F}' => {
                 Self::OtherSpaceSeparator
             }
-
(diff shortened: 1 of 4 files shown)


parley_engine/src/shape/cluster.rs
```diff
@@ -63,23 +63,83 @@ pub struct Char {
...
+    /// Note CSS Text 4 doesn't treat this space any differently from [`Self::OtherSpaceSeparator`],
+    /// but some applications do.
+    IdeographicSpace = 3,
+
+    /// The Unicode "space separators" (category `Zs`), except for the standard space, non-breaking
+    /// space and ideographic space.
+    ///
+    /// Applications may want to treat those three space separators differently from the general
+    /// class of space separators.
+    OtherSpaceSeparator = 4,
```
tomcur, line 100: Perhaps we want to carve out the "Ogham space mark" (U+1680) as well, now or in the future. That would grow this enum to requiring four bits to pack, but [CSS Text 4 § 4.3.2 Phase II  says those get trimmed (not as part of collapsing)](https://www.w3.org/TR/css-text-4/#white-space-phase-2). (I believe Blink doesn't do that, though.)

nicoburns: Gecko removes these, but Blink doesn't according to my LLM. Let's not worry too much about these for now? They seem to be for an ancient Irish script that I can't imagine is very common.

nicoburns (approved):

PR #786, opened 2026-09-08
TreeBuilder: fix (+ optimize) whitespace collapsing

nicoburns:
LLM Contributions: Generated with GPT 6 Astra High

This fixes a number of whitespace collapsing bugs, and also eliminates a per-span `String` allocation. 

Whitespace is now collapsed / trimmed where CSS Text says it should be:

- At the very start and very end of the Layout
- Immediately before and immediately after a preserved segment break (hard line break)
- But *not* after a preserved space or tab, which main dropped too and browsers keep: `<span style="white-space:pre">a </span> b` should be spaces wide.

Changes made include:

- Text now buffers into an intermediate buffer (stored in the `LayoutContext`).
- Whitespace collapsing happens in-place in the intermediate buffer rather than allocating a new `String`.
- The case of collapsed whitespace that is mixed `wrap`/`nowrap` is handled
- Buffered text is committed before switching whitespace modes, so text is processed under the mode in effect when it was pushed rather than the most recent one.
- `push_inline_box` flushes pending whitespace.

Note: the diff for this is pretty nasty. I found it easier to review the end state.

**Changelog**

> ### Fixed
>
> - Collapsible white space is now collapsed across style span and inline box boundaries instead of being trimmed at each span boundary, and is removed around hard line breaks.

### DIFF
--- a/parley/src/resolve/tree.rs
+++ b/parley/src/resolve/tree.rs
@@ -32,7 +32,12 @@ pub(crate) struct TreeStyleBuilder<B: Brush> {
     text: String,
     uncommitted_text: String,
     current_span: usize,
-    is_span_first: bool,
+    /// The span that a not-yet-committed collapsible whitespace sequence belongs to.
+    ///
+    /// Collapsible whitespace is only committed once it is known to be followed by content in the
+    /// same inline formatting context, so that it can collapse across span and inline box
+    /// boundaries and be removed at the end of the text.
+    pending_whitespace: Option<usize>,
     last_item_kind: ItemKind,
 }
 
@@ -52,7 +57,7 @@ impl<B: Brush> Default for TreeStyleBuilder<B> {
             text: String::new(),
             uncommitted_text: String::new(),
             current_span: usize::MAX,
-            is_span_first: false,
+            pending_whitespace: None,
             last_item_kind: ItemKind::None,
         }
     }
@@ -69,6 +74,8 @@ impl<B: Brush> TreeStyleBuilder<B> {
         self.white_space_collapse = WhiteSpaceCollapse::Preserve;
         self.text.clear();
         self.uncommitted_text.clear();
+        self.pending_whitespace = None;
+        self.last_item_kind = ItemKind::None;
 
         self.tree.push(StyleTreeNode {
             parent: None,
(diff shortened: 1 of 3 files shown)


parley/src/resolve/tree.rs
```diff
@@ -69,131 +86,194 @@ impl<B: Brush> TreeStyleBuilder<B> {
...
-                {
-                    span_text = span_text.trim_ascii_start();
+        let span = self.current_span;
+        match self.white_space_collapse {
+            WhiteSpaceCollapse::Preserve => {
+                if uncommitted_text.starts_with(is_segment_break) {
+                    // Collapsible whitespace immediately preceding a preserved segment break is
+                    // removed.
+                    self.pending_whitespace = None;
                 }
```
tomcur, line 129: This confused me for a bit, as it's collapsing whitespace in the `Preserve` branch, but I've since convinced myself it's correct.

Something like the following comment would've prevented my confusion.

```suggestion
            WhiteSpaceCollapse::Preserve => {
                if uncommitted_text.starts_with(is_segment_break) {
                    // Pending whitespace is always from a `WhiteSpaceCollapse::Collapse` span,
                    // and following CSS Text 4 § 4.3.1 Rule 1 must be removed if it immediately precedes
                    // a preserved segment break.
                    self.pending_whitespace = None;
                }
```

parley/src/resolve/tree.rs
```diff
@@ -13,13 +13,24 @@ struct StyleTreeNode<B: Brush> {
     parent: Option<usize>,
     style: ResolvedStyle<B>,
     style_id: Option<u16>,
+    /// The style's id when it is used for whitespace that wrapping is allowed after, which differs
+    /// from `style_id` only for spans that disable wrapping.
+    wrappable_style_id: Option<u16>,
```
tomcur, line 18: I think I understand why this is needed (e.g. the collapsed space in wrappable `aa ` + unwrappable ` bb` must still allow the line breaker to wrap).

With better transformation, this collapsed space could also become a zero-width space with the original style instead (that would require its own special handling in hanging, as it'd be a "synthetic" character and should be ignored for hanging).

(Nothing actionable here, just a note.)

parley/src/resolve/tree.rs
```diff
@@ -69,131 +86,194 @@ impl<B: Brush> TreeStyleBuilder<B> {
...
+                        rest = &rest[whitespace_len..];
+                        continue;
+                    }
+
+                    let text_len = rest
+                        .find(|c: char| c.is_ascii_whitespace())
+                        .unwrap_or(rest.len());
+                    self.flush_pending_whitespace();
+                    self.commit_text(span, &rest[..text_len]);
+                    rest = &rest[text_len..];
```
tomcur, line 157: Should this also check for a general `is_segment_break` before the `flush_pending_whitespace`? I.e., to collapse the whitespace before a non-ASCII break. We only treat ASCII breaks as collapsible, so something like `LS` gets through (which is probably fine), but it's then effectively a preserved mandatory break; the spec says to collapse whitespace before it.

The `Preserve` branch does check `is_segment_break`, and also `flush_pending_whitespace` collapses pending whitespace after an `is_segment_break`. I suppose if we don't collapse it before the break, it would normally end up hanging and not result in a visual difference, but I believe it's still a slight inconsistency.

```suggestion
                    let text_len = rest
                        .find(|c: char| c.is_ascii_whitespace())
                        .unwrap_or(rest.len());
                    if rest.starts_with(is_segment_break) {
                        // Collapsible whitespace immediately preceding a segment break is removed.
                        // ASCII `CR` and `LF` newlines are collapsed as whitespace, but `LS` and
                        // `PS` are treated as forced breaks.
                        self.pending_whitespace = None;
                    }
                    self.flush_pending_whitespace();
                    self.commit_text(span, &rest[..text_len]);
                    rest = &rest[text_len..];
```

tomcur (approved):
LGTM. A few comments inline, with a comment in the `Collapse` branch of `commit_uncommitted_text` that fixes what I think is an inconsistency.

parley/src/resolve/tree.rs
nicoburns: > I think I understand why this is needed (e.g. the collapsed space in wrappable aa  + unwrappable  bb must still allow the line breaker to wrap).

Yes, exactly.

> With better transformation, this collapsed space could also become a zero-width space with the original style instead (that would require its own special handling in hanging, as it'd be a "synthetic" character and should be ignored for hanging).

Not convinced that's a better representation (aren't we just introducing a problem for hanging with no gain?), but sure.

PR #788, opened 2026-09-08
Only count `Extended_Pictographic` for `is_emoji_or_pictograph`

nicoburns:
LLM Contributions: Investigation, profiling, and this PR generated by Fable 5.1 Low.

**Changelog: None**

## Summary

Since the transition from `swash` to `icu4x` for "analysis", we have been setting the `is_emoji_or_pictograph` bit for every character with either the `Emoji` or `Extended_Pictographic` properties. The UCD `Emoji` property marks ASCII digits `0-9` (and `#`, `*`) as `Emoji` (as they can form part of keycap emoji), so every digit cluster in Latin text is flagged "emoji".

`CharCluster::is_emoji()` therefore returns true for every digit cluster, and `parley`'s `FontSelector::select_font` takes its emoji branch on each one. This is slow because is causes `Query::set_families` to rebuilds the family list and re-resolve generic families, and `matches_with` to scan a widened candidate set.  Furthermore, if the next cluster is *not* an emoji (likely), then the same invalidation and slow family list rebuild occurs again for that cluster.

On a ~250KB English Wikipedia corpus (17.6k digit characters) this costs ~9-10ms of the build phase (most of the  ~12ms (+27%) regression introduced by the ICU4X migration commit)

## Fix

Only mark `Extended_Pictographic`s as emoji. This restores the pre-ICU4X semantics as swash's `is_emoji` was `is_extended_pictographic()`. There is an existing TODO in `parley_engine/src/shape/cluster.rs` which already tracks completing emoji detection (VS16 upgrade can key off the `VARIATION_SELECTOR` bit rather than `Emoji`).

Packed data regenerated with:

```
cargo run --locked -p parley_data_gen -- ./parley_data/src/generated --compression=5
cargo fmt -p parley_data
```


DJMcNab:
Like, I guess this is slightly better semantics than current main, but it's still moving from one wrong thing to another.

If you really think this is needed, I guess we can land it, although I'd rather focus on fixing the bug which makes Fontique so slow - I'm already working on bringing #620 into being landable (although other stuff keeps coming up to stop me focusing on that...)

nicoburns:
I would like to land this to be honest. This:

- Is a 2 line change (+ regenerating the data)
- Fixes a pretty bad bug (rendering ascii digits rendering as emoji), which we're actually seeing in the wild (much better to render emoji as text than text as emoji).
- Is a 20% performance boost to the slowest part of the pipeline

And also:

> other stuff keeps coming up to stop me focusing on that...

nicoburns:
Superceded by #811

PR #793, opened 2026-09-09
Add larger "document scale" benchmarks for each pipeline phase

nicoburns:
LLM Contributions: Generated with Fable 5.1 Low

Add a `document` benchmark group measuring build, break+align, content widths, and glyph iteration separately on the full-length text samples (~12–120KB) plus a ~250KB Latin document: the scale where the per-phase costs that the micro-bench axis comparisons exercise become measurable. Existing benchmark groups only cover excerpts of up to four paragraphs.

These are the benchmarks used for the historical sweep (see: https://xi.zulipchat.com/#narrow/channel/205635-parley/topic/Parley.20perf.200.2E11.20vs.20main.20.282026-09-08.29). So they're known to actually cover some of the perf issues that we care about.

**Changelog**: None

### DIFF
--- a/parley_bench/src/benches.rs
+++ b/parley_bench/src/benches.rs
@@ -5,7 +5,7 @@
 //!
 //! This module provides benchmarks for text layout and rendering.
 
-use crate::{ColorBrush, FONT_FAMILY_LIST, get_samples, with_contexts};
+use crate::{ColorBrush, FONT_FAMILY_LIST, get_documents, get_samples, with_contexts};
 use parley::{
     Alignment, AlignmentOptions, FontFamily, FontStyle, FontWeight, Layout, PositionedLayoutItem,
     StyleProperty,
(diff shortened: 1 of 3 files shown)


parley_bench/src/benches.rs
```diff
@@ -505,3 +505,74 @@ pub fn long_line() -> Vec<Benchmark> {
...
 }
+
+/// Benchmarks for document-scale inputs, exercising each layout pipeline phase
+/// separately.
+///
+/// The other groups cover short excerpts (at most four paragraphs) and mostly
+/// run the whole pipeline end-to-end. These use the full text samples
+/// (~12–120KB) plus a ~250KB Latin document, at a page-like 800px wrap width —
+/// the scale at which per-phase costs in build, line breaking, content widths,
+/// and glyph iteration diverge.
```
tomcur, line 516: The docstring isn't quite accurate as the other benches cover the separate parts of the pipeline now. I'd also remove the parts that are liable to becoming out-of-date.

```suggestion
/// These use the full text samples plus a ~250KB Latin document, at a page-like 800px
/// wrap width — the scale at which per-phase costs in build, line breaking, content
/// widths, and glyph iteration diverge.
```

tomcur (commented):
The benchmarks in `document` are already covered by other benchmarks. It would be interesting to put 250kB of text through those, but that could in principle be another `Sample`.

Maybe a huge source text combined with #800 could be nice?

PR #797, opened 2026-09-09
Don't create a line break opportunity at an out-of-flow inline box

nicoburns:
LLM Contributions: Generated 

This is a correctness fix. An out-of-flow box is not in-flow content, so unlike an in-flow InlineBox it introduces no line break opportunity.

**Changelog**

> ### Changed
>
> - OutOfFlow inline boxes no longer produce line-break opportunities


tomcur (approved):
LGTM!

PR #799, opened 2026-09-09
Reflow source text of benchmark samples

tomcur:
LLM Contributions: None.

The source text samples are hard-wrapped with `\n`. That means there are a ton of mandatory breaks in the samples, and that's confounding measuring our line wrapping, among other things. This removes those newlines, but keeps the paragraph newlines.

This is on top of #798, as otherwise it quite seriously impacts especially the Arabic benchmark timings in negative ways (as that sample's paragraphs are huge).

I considered using `WhiteSpaceCollapse::Collapse`, but that makes it harder to actually test paragraphs, and itself theoretically is a runtime workload worth measuring.

**Changelog: None**


taj-p (approved):
cLGTM if https://github.com/tomcur/parley/blob/a43a32a0c3994c859f2df4df05f2c10ab6eef431/parley_dev/assets/text_samples/README.md has its amendments column updated accordingly

tomcur:
> cLGTM if https://github.com/tomcur/parley/blob/a43a32a0c3994c859f2df4df05f2c10ab6eef431/parley_dev/assets/text_samples/README.md has its amendments column updated accordingly

Ah! Good point -- completely missed that README (maybe the first time not having a file tree in my editor has bitten).

PR #804, opened 2026-09-10
Expose Layout::clear

taj-p:
LLM Contributions: Claude Opus 5.

## Intent

We use a pool of [`Layout`] and I've found that, even after uninstalling a font, a `Layout` can retain the font in memory because it holds a strong reference to the underlying font data.

I would like to add a clear function to `Layout` so that I can be sure that, after uninstalling a font, its memory isn't going to be kept around due to some `Layout` holding onto it in our pool.

## Notes

- I created a new `parley/src/tests/test_layout.rs` file to house the test. I wanted to use a private API to enable equality comparison against `Layout::data`. I noticed `PartialEq` being implemented for `LayoutData` and `Brush` but not `Layout`, so I assumed that the omission was intentional?
- Anyway, to facilitate the test, I had to move a helper to a location that could be shared.

**Changelog**

> ### Added
> 
> - `Layout::clear` API to clear the contents of a layout. This is useful to ensure that all references are freed of an uninstalled font.

### DIFF
--- a/parley/src/layout/layout.rs
+++ b/parley/src/layout/layout.rs
@@ -48,11 +48,10 @@ impl<B: Brush> Layout<B> {
         Self::default()
     }
 
-    /// Clears the layout.
+    /// Clears the layout while retaining capacity.
     ///
-    /// When using a pool of [`Layout`] to reduce allocations for text layout,
-    /// [`Layout::clear`] is useful to ensure that no [`Layout`] maintains a strong reference
-    /// to an underlying font, which would prevent that fonts freeing.
+    /// This clears underlying strong references to fonts, which would otherwise
+    /// prevent those fonts from being freed until dropping or reusing this layout.
     pub fn clear(&mut self) {
         self.data.clear();
     }
--- a/parley/src/layout/layout.rs
+++ b/parley/src/layout/layout.rs
@@ -48,6 +48,22 @@ impl<B: Brush> Layout<B> {
         Self::default()
     }
 
+    /// Clears the layout.
+    ///
+    /// When using a pool of [`Layout`] to reduce allocations for text layout,
+    /// [`Layout::clear`] is useful to ensure that no [`Layout`] maintains a strong reference
+    /// to an underlying font, which would prevent that fonts freeing.
+    pub fn clear(&mut self) {
+        self.data.clear();
+        self.data.layout_max_advance = 0.0;
+        self.data.indent_amount = 0.0;
+        self.data.indent_options = IndentOptions::default();
+        #[cfg(feature = "accesskit")]
+        {
+            self.data.alignment = None;
+        }
+    }
+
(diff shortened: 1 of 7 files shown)


parley/src/layout/layout.rs
```diff
@@ -48,6 +48,22 @@ impl<B: Brush> Layout<B> {
...
+    /// to an underlying font, which would prevent that fonts freeing.
+    pub fn clear(&mut self) {
+        self.data.clear();
+        self.data.layout_max_advance = 0.0;
+        self.data.indent_amount = 0.0;
+        self.data.indent_options = IndentOptions::default();
+        #[cfg(feature = "accesskit")]
+        {
+            self.data.alignment = None;
+        }
```
nicoburns, line 64: I think these resets should be pushed data into `LayoutData::clear`

nicoburns (changes requested):

parley/src/layout/layout.rs
```diff
@@ -48,6 +48,22 @@ impl<B: Brush> Layout<B> {
         Self::default()
     }
 
+    /// Clears the layout.
+    ///
+    /// When using a pool of [`Layout`] to reduce allocations for text layout,
+    /// [`Layout::clear`] is useful to ensure that no [`Layout`] maintains a strong reference
+    /// to an underlying font, which would prevent that fonts freeing.
```
tomcur, line 55: Small typo:

```suggestion
    /// Clears the layout.
    ///
    /// When using a pool of [`Layout`] to reduce allocations for text layout,
    /// [`Layout::clear`] is useful to ensure that no [`Layout`] maintains a strong reference
    /// to an underlying font, which would prevent that font from being freed.
```

Or at your option, the docstring could describe the operational aspects.

```suggestion
    /// Clears the layout while retaining capacity.
    ///
    /// This clears underlying strong references to fonts, which would otherwise
    /// prevent those fonts from being freed until dropping or reusing this layout.
```

tomcur (approved):
LGTM! Approved, given Nico's point about resetting is addressed.

Left a small nit inline.

> * I noticed `PartialEq` being implemented for `LayoutData` and `Brush` but not `Layout`, so I assumed that the omission was intentional?

I don't know if it's intentional, but perhaps as `Layout` is public we don't necessarily want to guarantee `Layout: PartialEq`.

parley/src/layout/layout.rs
taj-p: Much better! Thank you!

parley/src/layout/layout.rs
taj-p: Moved in https://github.com/linebender/parley/pull/804/commits/ab6cf723fbe701709d70348ffe7e0145974dff33 - thank you!

DJMcNab:
(you should be able to dismiss Nico's review if it's been addressed)

nicoburns (approved):

PR #807, opened 2026-09-10
Benchmark layout cursor queries

tomcur:
LLM Contributions: Initial implementation, review.

This adds some benchmarks for `Cursor` navigation. (I'm working on `parley::layout::Cluster` improvements which should impact query and line breaking performance.)

We may want some more benchmarks (like selection geometry or hit-testing), so I started this in a new module to scope the possible future shared helpers.

**Changelog: None**

### DIFF
--- a/parley_bench/benches/main.rs
+++ b/parley_bench/benches/main.rs
@@ -10,7 +10,7 @@ use parley_bench::benches::{
     repeated_justification, spacing, styled,
 };
 use parley_bench::fontique_benches::system_fonts_init;
-use parley_bench::query_benches::caret_navigation;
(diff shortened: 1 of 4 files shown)


.bash_profile
taj-p, line 1: Can you please delete these dot files prior to merge

taj-p (approved):
Looks good to me so long as the dot files are removed first

.bash_profile
tomcur: Whoops; for some reason Claude Code has been leaving these for a while now (something to do with the sandbox), and `jj` automatically snapshots everything in its revisions... unfortunate combination :(

PR #810, opened 2026-09-10
Make `parley::Layout::Cluster` character-addressed

tomcur:
LLM Contributions: Investigation, initial implementation, review.

First: note this `Cluster` is effectively a grapheme, but with some caveats (see also
https://github.com/linebender/parley/pull/715/changes#r3693621362 for some prior discussion on its name).

`parley` currently addresses `Cluster` by line index, run index, and finally cluster/grapheme index within the run. That grapheme index is a bit troublesome, as grapheme positions are generally not needed unless a layout is actually queried, so it's not quite a first-class indexing mechanism. Currently, querying is quadratic in many places because of this.

This proposes replacing the cluster/grapheme index with the grapheme's starting character index. That stops us from needing to count graphemes in quite a few places. Note that we *could* store some grapheme map to get a similar speed-up, but that could itself become troublesome once we support graphemes spanning runs (i.e., one source grapheme may be in two or more runs, which isn't that trivial to map). Character indices don't have that problem, and don't require additional state. Most other places in the code use character indices already.

I've made a little bit of effort to ensure the AccessKit `NodeId`s remain stable under character insertion/deletion etc., but I've not gone too deep on that as the intention is to move that code out of `parley`: https://github.com/linebender/parley/pull/716.

# Performance

```
$ cargo bench --bench main -- compare ../target/benchmarks/main -t 4. --filter "*Break*"

Line Breaking - arabic 8000 characters, narrow              [ 150.9 us ... 134.9 us ]     -10.61%*
Line Breaking - arabic 8000 characters, wider               [ 111.3 us ...  91.0 us ]     -18.24%*
Line Breaking - arabic 8000 characters, unwrapped           [  94.1 us ...  75.9 us ]     -19.33%*
Line Breaking - latin 8000 characters, narrow               [ 106.6 us ... 100.8 us ]      -5.42%*
Line Breaking - latin 8000 characters, wider                [  60.4 us ...  57.7 us ]      -4.37%*
Line Breaking - latin 8000 characters, unwrapped            [  46.3 us ...  43.6 us ]      -5.73%*
Line Breaking - japanese 8000 characters, narrow            [ 151.8 us ... 134.1 us ]     -11.69%*
Line Breaking - japanese 8000 characters, wider             [ 113.4 us ...  95.9 us ]     -15.37%*
Line Breaking - japanese 8000 characters, unwrapped         [ 101.3 us ...  84.9 us ]     -16.25%*
Line Breaking - arabic 8000 characters, wrapped + spacing   [ 123.9 us ... 106.6 us ]     -13.92%*
Line Breaking - latin 8000 characters, wrapped + spacing    [  78.1 us ...  74.7 us ]      -4.42%*
Line Breaking - japanese 8000 characters, wrapped + spacing [ 125.6 us ... 110.6 us ]     -11.96%*
Page Break - arabic 8000 characters                         [  99.8 us ...  78.5 us ]     -21.31%*
Page Break - latin 8000 characters                          [  49.0 us ...  46.9 us ]      -4.12%*
Page Break - japanese 8000 characters                       [ 105.8 us ...  88.5 us ]     -16.35%*
```

```
$ cargo bench --bench main -- compare ../target/benchmarks/main -t 4. --filter "*Query*"

Query Cursor - arabic 8000 characters, next visual               [  47.7 us ...  35.7 us ]     -25.21%*
Query Cursor - latin 8000 characters, next visual                [ 333.7 us ...  42.0 us ]     -87.41%*
Query Cursor - japanese 8000 characters, next visual             [  64.6 us ...  37.3 us ]     -42.23%*
Query Cursor - arabic 8000 characters, line down + up            [  59.4 us ...  61.2 us ]      +3.07%*
Query Cursor - latin 8000 characters, line down + up             [ 325.1 us ... 234.6 us ]     -27.82%*
Query Cursor - japanese 8000 characters, line down + up          [  64.5 us ...  59.0 us ]      -8.47%*
Query Cursor - latin 8000 characters, line down + up, long lines [   3.1 ms ...   2.1 ms ]     -31.91%*
Query Cursor - latin 8000 characters, from byte index            [   7.3 us ... 795.0 ns ]     -89.04%*
```

**Changelog**

> ## Parley
> ### Changed
>
> - Breaking change: `ClusterPath::logical_index` was removed. You can resolve the path to a `Cluster` using `ClusterPath::cluster`. `ClusterPath` still provides the line and run indices, but should otherwise be treated as an opaque handle.
> - Breaking change: `Run::{logical_to_visual, visual_to_logical}` were removed. Use `Run::visual_clusters` or `Cluster::{next, previous}_visual` instead.

### DIFF
--- a/parley/src/layout/cluster.rs
+++ b/parley/src/layout/cluster.rs
@@ -22,7 +22,6 @@ use parley_engine::{Atom, Glyph, Grapheme, shape::Whitespace};
 ///
 /// [uax-grapheme]: https://www.unicode.org/reports/tr29/#Grapheme_Cluster_Boundaries
 pub struct Cluster<'a, B: Brush> {
-    pub(crate) path: ClusterPath,
     pub(crate) run: Run<'a, B>,
     /// The atom containing this grapheme.
     pub(crate) atom: Atom<'a>,
@@ -52,19 +51,10 @@ pub enum ClusterSide {
 impl<'a, B: Brush> Cluster<'a, B> {
     /// Returns the cluster for the given layout and byte index.
     pub fn from_byte_index(layout: &'a Layout<B>, byte_index: usize) -> Option<Self> {
-        if let Some((_, line)) = layout.line_for_byte_index(byte_index) {
-            for run in line.runs() {
-                if !run.text_range().contains(&byte_index) {
-                    continue;
-                }
-                for cluster in run.clusters() {
-                    if cluster.text_range().contains(&byte_index) {
-                        return Some(cluster);
-                    }
-                }
-            }
-        }
-        None
+        let (_, line) = layout.line_for_byte_index(byte_index)?;
+        line.runs()
+            .filter(|run| run.text_range().contains(&byte_index))
+            .find_map(|run| run.cluster_at_text_byte(byte_index))
     }
 
     /// Returns the cluster and side which is at the specified position in the given layout. If no cluster is
@@ -90,9 +80,9 @@ impl<'a, B: Brush> Cluster<'a, B> {
         y: f32,
         exact: bool,
     ) -> Option<(Self, ClusterSide)> {
-        let mut path = ClusterPath::default();
-        if let Some((line_index, line)) = layout.line_for_offset(y) {
-            path.line_index = line_index as u32;
+        let mut line_index = 0;
+        if let Some((index, line)) = layout.line_for_offset(y) {
+            line_index = index;
             let mut offset = line.metrics().offset + line.metrics().inline_min_coord;
             let last_run_index = line.len().saturating_sub(1);
             for item in line.items_nonpositioned() {
@@ -100,17 +90,12 @@ impl<'a, B: Brush> Cluster<'a, B> {
                     LineItem::Run(run) => {
                         let is_last_run = run.index as usize == last_run_index;
                         let run_advance = run.advance();
-                        path.run_index = run.index;
-                        path.logical_index = 0;
                         if x > offset + run_advance && (exact || !is_last_run) {
                             offset += run_advance;
                             continue;
                         }
-                        let last_cluster_index = run.cluster_range().len().saturating_sub(1);
-                        for (visual_index, cluster) in run.visual_clusters().enumerate() {
-                            let is_last_cluster = is_last_run && visual_index == last_cluster_index;
-                            path.logical_index =
-                                run.visual_to_logical(visual_index).unwrap_or_default() as u32;
+                        for cluster in run.visual_clusters() {
+                            let is_last_cluster = is_last_run && cluster.is_visual_last_in_run();
                             let cluster_advance = cluster.advance();
                             let edge = offset;
                             offset += cluster_advance;
@@ -135,7 +120,9 @@ impl<'a, B: Brush> Cluster<'a, B> {
             }
         }
         if y <= 0.0 && !exact {
-            Some((path.cluster(layout)?, ClusterSide::Left))
+            // Fall back to the first cluster of the line's first run.
+            let cluster = layout.get(line_index)?.item(0)?.run()?.clusters().next()?;
+            Some((cluster, ClusterSide::Left))
         } else {
             None
         }
@@ -151,10 +138,13 @@ impl<'a, B: Brush> Cluster<'a, B> {
         self.run
     }
 
-    /// Returns the path that contains the set of indices to reach the cluster
-    /// from a layout.
+    /// Returns the path to reach the cluster from a layout.
     pub fn path(&self) -> ClusterPath {
-        self.path
+        ClusterPath::new(
+            self.run.line_index,
+            self.run.index,
+            self.grapheme.char_range().start,
+        )
     }
 
     /// The first (logical) character of this cluster.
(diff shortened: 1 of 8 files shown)


DJMcNab:
> support graphemes spanning runs 

Can you remind me again what the situation where this can happen is? Would it be to do with regional indicators?

parley/src/layout/cluster.rs
```diff
@@ -52,19 +51,10 @@ pub enum ClusterSide {
 impl<'a, B: Brush> Cluster<'a, B> {
     /// Returns the cluster for the given layout and byte index.
```
DJMcNab: Whilst we're here, maybe `byte index in the source text`?

parley/src/layout/cluster.rs
```diff
@@ -52,19 +51,10 @@ pub enum ClusterSide {
...
-                        return Some(cluster);
-                    }
-                }
-            }
-        }
-        None
+        let (_, line) = layout.line_for_byte_index(byte_index)?;
+        line.runs()
+            .filter(|run| run.text_range().contains(&byte_index))
+            .find_map(|run| run.cluster_at_text_byte(byte_index))
```
DJMcNab: Nit: The use of `find_map` here after the filter feels a bit weird. If `find().and_then` wouldn't work, then we should possibly have a comment.

parley/src/layout/cluster.rs
```diff
@@ -151,10 +138,13 @@ impl<'a, B: Brush> Cluster<'a, B> {
         self.run
     }
 
-    /// Returns the path that contains the set of indices to reach the cluster
-    /// from a layout.
+    /// Returns the path to reach the cluster from a layout.
```
DJMcNab, line 141: This is a fairly existential doc! I don't think it's worth blocking this PR on it, but these docs could need a decent bit of improvement. The current one is [very sparse](https://docs.rs/parley/latest/parley/layout/struct.ClusterPath.html).

parley/src/layout/cluster.rs
```diff
@@ -257,16 +247,35 @@ impl<'a, B: Brush> Cluster<'a, B> {
             .flatten()
     }
 
+    /// Whether this is the visually first cluster of its run.
```
DJMcNab: This is a bit of an aside for this PR, but...

Does this mean 'leftmost' or 'visually first'? Because I would argue that in a fully RTL layout, "visually first" pretty defensibly would be rightmost 🙃

parley/src/layout/cluster.rs
DJMcNab, line 355: Lol - GitHub doesn't show this without expanding, so this function looks very wrong!

parley/src/layout/cluster.rs
```diff
@@ -279,16 +288,23 @@ impl<'a, B: Brush> Cluster<'a, B> {
...
+    fn next_in_run(&self) -> Option<Self> {
+        self.run
+            .cluster_containing_char(self.grapheme.char_range().end)
+    }
+
+    /// The cluster logically preceding this one within the same run, if any.
+    fn previous_in_run(&self) -> Option<Self> {
+        self.run
+            .cluster_containing_char(self.grapheme.char_range().start.checked_sub(1)?)
+    }
```
DJMcNab, line 301: I guess it doesn't really matter for private items, but I'd probably call these e.g. `next_logical_in_run`

parley/src/layout/cluster.rs
```diff
@@ -496,15 +502,16 @@ impl Affinity {
 pub struct ClusterPath {
     pub(crate) line_index: u32,
     pub(crate) run_index: u32,
-    pub(crate) logical_index: u32,
+    /// The cluster's first (logical) character index.
+    pub(crate) char_index: u32,
```
DJMcNab: Maybe `The index of the cluster's first (logical) character`?

DJMcNab (approved):
I've only skimmed the new tests and the accessibility stuff, but otherwise this all seems reasonable enough. If you think it needs a more thorough review, feel free to re-request.

parley/src/layout/cluster.rs
tomcur: Good point. "Leftmost" would clearly be better. I've rewritten this to now use the "line-left" and "line-right" terms from [CSS Text 4 § 6.3's line relative direction terms](https://www.w3.org/TR/css-writing-modes-4/#line-directions).

parley/src/layout/cluster.rs
tomcur: Much clearer!

parley/src/layout/cluster.rs
tomcur: Good point. Also makes querying a bit faster still!

```
Query Cursor - arabic 8000 characters, next visual [  38.6 us ...  33.3 us ]     -13.58%*
Query Cursor - latin 8000 characters, next visual  [  43.8 us ...  38.9 us ]     -11.18%*
Query Cursor - japanese 8000 characters, next visual [  40.3 us ...  35.0 us ]     -13.02%*
Query Cursor - arabic 8000 characters, line down + up [  61.4 us ...  58.4 us ]      -4.92%*
Query Cursor - latin 8000 characters, line down + up [ 228.2 us ... 227.6 us ]      -0.29%
Query Cursor - japanese 8000 characters, line down + up [  59.4 us ...  58.1 us ]      -2.11%*
Query Cursor - latin 8000 characters, line down + up, long lines [   2.0 ms ...   2.0 ms ]      +0.16%
Query Cursor - latin 8000 characters, from byte index [ 800.0 ns ... 725.4 ns ]      -9.33%*
```

tomcur:
> > support graphemes spanning runs
>
> Can you remind me again what the situation where this can happen is? Would it be to do with regional indicators?

Within a line, it can happen if font selection causes glyphs of a grapheme to be taken from different fonts. With our current charmap coverage-based font selection, this would only happen if the characters are from different scripts (e.g., a prepend character from some script followed by a character from another script). Regional indicator symbols would not split across runs as we've set them up now, but theoretically they could split across style spans (currently we just ignore style changes if they're inside a `grapheme ∩ font run`).

Grapheme can also split across lines if we follow Unicode's line and grapheme segmenters exactly: these segment such that a line breaks exist within graphemes (e.g., `space + U+0301` is a grapheme, and is allowed to break after the space).

parley/src/layout/cluster.rs
DJMcNab: We should probably do a full sweep over everything. Something to discuss in OH tomorrow!

DJMcNab:
Right. We don't support that currently though, I guess!

We also need it for emoji, unfortunately. For example `a[skin tone modifier]` is I think one grapheme with two interior font preference orders!

PR #811, opened 2026-09-14
Improve resolution of preferred presentation mode for Emoji

### LINKED ISSUE #744: Emoji font is selected for plain digits, #, *, © and ® when no requested family resolves
ASCII digits, `#`, `*`, `©` and `®` are shaped with the emoji font whenever no requested family resolves, because `select_font` appends `GenericFamily::Emoji` for any character with the Unicode `Emoji` property — not just those that default to emoji presentation.

Reported from Blitz via https://github.com/DioxusLabs/blitz/issues/688, where @nicoburns suggested it belongs here. The repro below is pure Parley, no Blitz.

## Repro

`parley` at `a0752c7` (main, 2026-08-10), Linux x86_64, fontconfig, with `NotoColorEmoji.ttf` installed.

Add a workspace member `examples/emoji_repro`:

```toml
# examples/emoji_repro/Cargo.toml
[package]
name = "emoji_repro"
edition = "2024"
publish = false

[dependencies]
parley = { workspace = true, features = ["system"] }
skrifa = { workspace = true, features = ["std"] }
```

```rust
// examples/emoji_repro/src/main.rs
use parley::{
    FontContext, FontFamily, GlyphRun, LayoutContext, PositionedLayoutItem, StyleProperty,
};
use skrifa::{MetadataProvider, raw::FontRef};

fn family_name(run: &GlyphRun<'_, [u8; 4]>) -> String {
    let font = run.run().font();
    let data = font.font.data.as_ref();
    match FontRef::from_index(data, font.font.index) {
        Ok(font_ref) => font_ref
            .localized_strings(skrifa::string::StringId::FAMILY_NAME)
            .english_or_first()
            .map(|s| s.chars().collect::<String>())
            .unwrap_or_else(|| "<no name table>".to_string()),
        Err(e) => format!("<unreadable: {e}>"),
    }
}

fn main() {
    let requested = std::env::args()
        .nth(1)
        .unwrap_or_else(|| "NoSuchFontAtAll".to_string());

    let text = "0123456789 abcdefghij #*+,-. \u{00a9}\u{00ae} \u{00e9}\u{00f1}";

    let mut layout_cx: LayoutContext<[u8; 4]> = LayoutContext::new();

    // `css:foo, bar` goes through CSS parsing (so generics resolve as generics);
    // anything else is pushed as a single *named* family.
    let family = match requested.strip_prefix("css:") {
        Some(source) => FontFamily::Source(source.into()),
        None => FontFamily::named(&requested),
    };
    builder.push_default(StyleProperty::FontFamily(family));
    builder.push_default(StyleProperty::FontSize(30.0));

    layout.break_all_lines(None);

    println!("requested font-family: {requested:?}");
    for line in layout.lines() {
        for item in line.items() {
            let PositionedLayoutItem::GlyphRun(glyph_run) = item else {
                continue;
            };
            let range = glyph_run.run().text_range();
            println!("{:<20}  {:?}", family_name(&glyph_run), &text[range]);
        }
    }
}
```

### Actual

```
$ cargo run -p emoji_repro
requested font-family: "NoSuchFontAtAll"
Noto Color Emoji      "0123456789"
Noto Sans             " abcdefghij "
Noto Color Emoji      "#*"
Noto Sans             "+,-. "
Noto Color Emoji      "©®"
Noto Sans             " éñ"
```

### Expected

All of it in `Noto Sans` — the same font script fallback already picks for the letters in the same run.

The affected set is exactly `0`–`9` `#` `*` `©` `®`: the `Emoji=Yes` characters outside the emoji blocks. `$ , . + - / %` are Common-script too and are unaffected, because they are not `Emoji=Yes`.

### Controls

```
$ cargo run -p emoji_repro -- "Noto Sans"
Noto Sans             "0123456789 abcdefghij #*+,-. ©® éñ"

$ cargo run -p emoji_repro -- "css:sans-serif"
Noto Sans             "0123456789 abcdefghij #*+,-. ©® éñ"

$ cargo run -p emoji_repro -- "css:NoSuchFontAtAll, sans-serif"
Noto Sans             "0123456789 abcdefghij #*+,-. ©® éñ"

$ cargo run -p emoji_repro -- "sans-serif"     # quoted in CSS, so a *named* family
Noto Color Emoji      "0123456789"
...
```

So the trigger is precisely **no requested family resolves**. As long as one does, it covers the digits and wins on complete coverage before the appended emoji generic is reached, and nothing looks wrong.

## Mechanism

`parley_engine/src/shape/cluster.rs:354` sets the cluster's `is_emoji` from `CharInfo::is_emoji_or_pictograph()`, which `parley_data_gen/src/lib.rs:41-42` builds as `Emoji ∪ Extended_Pictographic`. The Unicode `Emoji` property is `Yes` for `0`–`9`, `#`, `*`, `©` and `®` with no VS16 and no keycap sequence — those characters are `Emoji_Presentation=No`, i.e. they default to *text* presentation per [UTR #51](https://www.unicode.org/reports/tr51/#Presentation_Style).

`parley/src/shape/mod.rs:328-332` then does:

```rust
if is_emoji {
    let emoji_family = QueryFamily::Generic(GenericFamily::Emoji);
    self.query.set_families(fonts.chain(once(emoji_family)));
    self.fonts_id = None;
}
```

Because the emoji generic joins the **explicit family list**, it is consulted ahead of script fallback. When `fonts` is empty or unresolvable, the emoji family is the first thing that resolves, it covers the digit completely, `matches_with` stops there, and fallback — which serves every other character in the same run — never runs.

There is already a `TODO - make emoji detection more complete` at `parley_engine/src/shape/cluster.rs:351`, and two `#[allow(dead_code, reason = "To be used in more complete emoji checking, in select_font")]` bits (`VARIATION_SELECTOR_MASK`, `REGION_INDICATOR_MASK`) at `parley_engine/src/analysis.rs:197-206`, so this looks like known-incomplete ground rather than a surprise.

## Suggested fix

Drive the emoji-family append from **default emoji presentation** rather than from `Emoji`: `Emoji_Presentation=Yes`, or an explicit emoji presentation sequence (base + U+FE0F), or a keycap sequence (base + U+FE0F + U+20E3). A base followed by U+FE0E should stay on the text path.

That wants a *new* property bit rather than a redefinition of the existing one: `is_emoji_or_pictograph` also feeds cluster-level deletion in `parley/src/editing/editor.rs:297`, where the wider `Extended_Pictographic` set is the right one. The same change would also fix text-presentation pictographs like `☺` U+263A.

Related: #492 (configuring preferred emoji presentation) is the explicit-override half of this; this issue is the default-presentation half. #695 (renderability axis) would independently stop a renderer that cannot paint CBDT/COLR from being handed such a font, which is what turned this into *invisible* digits rather than merely differently-styled ones downstream.

Happy to open a PR for the `Emoji_Presentation` gating if you'd like it — say which shape you'd prefer for the new bit.

## Notes

Linux/fontconfig only so far. On macOS the emoji generic resolves to Apple Color Emoji and the digits are likely to paint as *something*, so this may present as wrong-font rather than missing-glyph there.

DJMcNab:
LLM Contributions: Unknown for #620, c-emoji. I used them for review, bootstrapping the new crate to follow https://linebender.org/wiki/formatting-scheme/, and getting/implementing the technique of sticking the properties into our existing property infrastructure, plus rebasing/reordering the series.

Adaptation of #620 by @fundon.

This uses a state machine based on @chansen's [c-emoji](https://github.com/chansen/c-emoji) to implement detection of Emoji sequences, as described in [tr-51](https://www.unicode.org/reports/tr51/tr51-29.html). This is Unicode 17 Emoji, as opposed to Unicode 18 (releasing tomorrow...), but I don't think the difference has a huge impact for us.

The state machine lives in a new crate `parley-emoji`, so as to follow c-emoji's license. We were unable to get its code relicensed under our MIT/Apache 2.0 license.

With this, we are *not* completely compliant with Unicode Emoji, but this gives us a significantly better jumping off point.
Things I've chosen to defer:
- Better testing. I have local changes which generate a suitable subset of Noto Color Emoji, but I think end-to-end testing needs more care.
  - Also not yet tested using the upstream Unicode data.
- Splitting a single grapheme cluster into its 'text presentation/emoji presentation' preferred sections. This can happen in three cases that I'm aware of:
  - Prepend character then Emoji, all within one grapheme cluster
  - Non-Emoji modifier base, then Fitzgerald skin tone modifier
  - ZWJ sequence containing VS15 text variation selectors.
- Having the Emoji font in the fallback chain, even if text presentation is preferred.
- Thinking through how the 'reject' case should be handled
- Doing the proper rendering of malformed tag sequences, which is required by spec, but which as far as I know, no system implements.
- [#font-tech > Appearance of unsupported emoji sequences](https://xi.zulipchat.com/#narrow/channel/296640-font-tech/topic/Appearance.20of.20unsupported.20emoji.20sequences/with/623471132).
- Verifying that the handling of RI sequences is correct - in particular, the way that three regional indicators in a row merge into grapheme clusters can lead to jank.
- Fix the editing behaviour to use this properly (see also #693)
- Documenting our level of compliance with the Unicode Emoji spec.
- Handling legacy modifier sequences with an intervening VS16. There's a TODO in the code for this.
- The 'emoji presentation locale' thing.

I don't believe any of those are blocking - this is still a better state than before.

Fixes #744.
Supersedes #620, #788, #692.

cc @kane50613 (wrt #745).

**Changelog**

> ### Fixed
>
> - Parley's detection of Emoji is now more robust.


parley_engine/src/analysis.rs
```diff
@@ -260,6 +275,44 @@ impl CharInfo {
         self.flags & Self::EMOJI_OR_PICTOGRAPH_MASK != 0
     }
 
+    /// Whether this character has the `Emoji` property ([UTS #51][]).
+    ///
+    /// Unlike [`is_emoji_or_pictograph`](Self::is_emoji_or_pictograph), this excludes
+    /// characters which are only `Extended_Pictographic`.
```
tomcur: Mega-nit (as the clause is restrictive):

```suggestion
    /// Unlike [`is_emoji_or_pictograph`](Self::is_emoji_or_pictograph), this excludes
    /// characters that are only `Extended_Pictographic`.
```

kane50613:
I'll rebase #745 once this is merged

parley_engine/src/shape/cluster.rs
```diff
@@ -61,6 +62,8 @@ pub struct Char {
     /// Indexes into the list of styles for the containing text run, to find the style applicable
     /// to this character.
     pub style_index: u16,
+    /// True if the character is an emoji presentation selector (VS15 or VS16).
+    pub is_emoji_presentation_selector: bool,
```
tomcur: Is this field needed, or should users just use the derived `CharCluster::emoji_presentation_style` / `CharCluster::is_emoji`?

(This field on `Char` itself is unread at the moment.)

parley/src/lib.rs
```diff
@@ -127,6 +127,7 @@ mod tests;
 
 pub use linebender_resource_handle::FontData;
 pub use parlance::BaseDirection;
+pub use parley_emoji as emoji;
```
tomcur: Could we keep `parley_emoji` private instead of re-exporting?

```suggestion
#[cfg(test)]
pub(crate) use parley_emoji as emoji;
```

(Or just import this in the tests.)

parley/src/tests/test_emoji.rs
```diff
@@ -0,0 +1,710 @@
...
+    assert_emoji(TestEntity {
+        sequence: &[
+            0x1F600, // GRINNING FACE
+        ],
+        categories: &[EmojiSegmentationCategory::EmojiPresentation],
+        style: EmojiPresentationStyle::Emoji,
+        is_emoji: true,
+        has_vs: false,
+    });
+}
```
tomcur, line 76: Are these tests in `parley` instead of `parley_emoji` for licensing reasons (i.e., to keep the test code `Apache-2.0 OR MIT`)?

parley_emoji/src/lib.rs
```diff
@@ -0,0 +1,59 @@
...
+//!
+//! In Parley, this impacts font selection for these clusters.
+//!
+//! # Features
+//!
+//! The following crate [feature flags](https://doc.rust-lang.org/cargo/reference/features.html#dependency-features) are available:
+//!
+//! - `std` (enabled by default): This is currently unused and is provided for forward compatibility.
+//!
+//! Note that Parley Emoji currently requires that an allocator is available (i.e. it uses [`alloc`][_alloc]).
```
tomcur: But currently it doesn't allocate, right? If so, maybe something like the following.

```suggestion
//! Note that Parley Emoji currently requires that an allocator is available (i.e. it depends on [`alloc`][_alloc]).
//! Currently, Parley Emoji does not actually allocate, but the dependency is kept so starting to allocate
//! in the future is not a breaking change.
```

parley/src/tests/test_emoji.rs
```diff
@@ -0,0 +1,710 @@
...
+        is_emoji: false,
+        has_vs: false,
+    });
+}
+
+// Keycap sequences
+
+// Lone keycap base; Encoded: 1
+#[test]
+fn long_keycap_base() {
```
tomcur:
```suggestion
// Lone keycap base; Encoded: 1
#[test]
fn lone_keycap_base() {
```

parley_emoji/LICENSE-MIT
```diff
@@ -0,0 +1,27 @@
+Copyright 2020 the Parley Authors
```
tomcur, line 1: I truly don't know, but should this be `2026`?

parley_engine/src/analysis.rs
tomcur: Mega-nit (as the clause is restrictive):

```suggestion
    /// Unlike [`is_emoji_or_pictograph`](Self::is_emoji_or_pictograph), this excludes
    /// characters that are only `Extended_Pictographic`.
```

parley_engine/src/analysis.rs
```diff
@@ -220,19 +227,27 @@ impl CharInfo {
...
+                | (is_region_indicator as u16) << Self::REGION_INDICATOR_SHIFT
+                | (is_control as u16) << Self::CONTROL_SHIFT
+                | (is_emoji_or_pictograph as u16) << Self::EMOJI_OR_PICTOGRAPH_SHIFT
+                | (contributes_to_shaping as u16) << Self::CONTRIBUTES_TO_SHAPING_SHIFT
+                | (force_normalize as u16) << Self::FORCE_NORMALIZE_SHIFT
+                | (is_grapheme_start as u16) << Self::GRAPHEME_START_SHIFT
+                | (is_emoji as u16) << Self::EMOJI_SHIFT
+                | (is_emoji_presentation as u16) << Self::EMOJI_PRESENTATION_SHIFT
+                | (is_emoji_modifier as u16) << Self::EMOJI_MODIFIER_SHIFT
+                | (is_emoji_modifier_base as u16) << Self::EMOJI_MODIFIER_BASE_SHIFT,
```
tomcur, line 250: I wonder whether these new emoji flags are required here. `CharInfo` is long-lived and a bit performance sensitive, so there's value in trying to keep it small. Of course there's room left, so storing these probably wouldn't *currently* cost us. So, don't consider this blocking (but then perhaps either refute my point or place a TODO comment to evaluate whether the fields can be pruned/moved out). 

From reading the code, it seems like the emoji properties are only needed for font selection during `CharCluster::fill`, which gets the source `&str`, so has access to the characters. It seems like `fill` could instead fetch the `Properties::get`, and then we don't have to wire them in via `CharInfo` (i.e., perhaps *all* the emoji flags could be removed here?). Note `CharCluster::contributes_to_shaping` already computes `Properties`.

tomcur (commented):
Overall this looks great. A few comments inline. I've not gone over the automaton itself, but I trust it's correct.

Please see especially my comment about the flags on `CharInfo`.

parley_data/src/lib.rs
```diff
@@ -153,6 +176,50 @@ impl Properties {
             Self::IS_MANDATORY_LINE_BREAK_BITS,
         ) != 0
     }
+
+    /// Returns whether the character has the `Emoji` property ([UTS #51][]).
+    ///
+    /// Unlike [`is_emoji_or_pictograph`](Self::is_emoji_or_pictograph), this excludes
+    /// characters which are only `Extended_Pictographic`.
```
tomcur: Same mega-nit as https://github.com/linebender/parley/pull/811#discussion_r4005438041 (the clause is restrictive).

```suggestion
    /// Unlike [`is_emoji_or_pictograph`](Self::is_emoji_or_pictograph), this excludes
    /// characters that are only `Extended_Pictographic`.
```

parley_engine/src/shape/cluster.rs
DJMcNab: Probably not needed. This is one of the consequences of this being the continuation of a long-running pr! I'll remove it.

parley/src/lib.rs
DJMcNab: Agreed - the re-export isn't needed. This is just another decision I didn't re-evaluate

parley/src/tests/test_emoji.rs
DJMcNab: Nope! Mainly just inertia - they were added here before both `parley_emoji` and `parley_engine` existed. I'll think about where they should go, and make sure the licensing is right.

parley_emoji/LICENSE-MIT
DJMcNab: I believe not. This is intentionally the same line as our existing LICENSE-MITs

parley_data/src/lib.rs
DJMcNab: Fwiw, I'm not sure I completely understand the comment - I still prefer the original form.

But I think I'll just remove the clause entirely - it's not really useful.

parley_engine/src/analysis.rs
DJMcNab: > Note `CharCluster::contributes_to_shaping` already computes `Properties`.

Can you expand on this? It seems like it's already calculated from the `properties` at the same time as the rest of the flags. I thought this meant that we were already fetching `Properties` "twice".

Given that this doesn't increase the size, I'm inclined to leave this, and evaluate moving this as a follow-up.

nicoburns:
In my ~60ms test case, this *does* recover the same (~9ms) performance improvement in font-fallback as #788. It also introduces a small 1-3ms of overhead of it's own, but that's probably expected.

parley_data/src/lib.rs
tomcur: Because the meaning was abundantly clear it doesn't really matter, but for posterity: what I've been taught is that, strictly speaking, "which" describes a property of something, whereas "that" subsets a group. (But it's very possible that language ship has sailed; I already only care about this stuff in technical writing anyway.)

parley_engine/src/analysis.rs
fundon: It's great to merge them together. I tried doing this before but encountered overflow with `u8`.

tomcur: What I meant with my comment about `contributes_to_shaping` specifically, is that `CharCluster:fill` fetches `Properties`, just to illustrate it could use those same properties to get at the flags.

Instead, the code now gets the flags during analysis and stores them. This then survives for as long as someone holds on to `Analysis`. I'm not sure, but I think the flags really are only needed transiently during `CharCluster::fill`.

parley_emoji/LICENSE-MIT
tomcur: That's fair (I did go looking to check it was the same year). I was unsure because we do set module copyright headers to the date of module creation.

parley_data/src/lib.rs
```diff
@@ -46,6 +57,10 @@ impl Properties {
     }
 
     /// Creates a new [`Properties`] from the given properties
+    #[allow(
+        clippy::too_many_arguments,
+        reason = "one argument per packed property; only called from the data generator and tests"
+    )]
```
tomcur, line 63:
```suggestion
    #[expect(
        clippy::too_many_arguments,
        reason = "one argument per packed property; only called from the data generator and tests"
    )]
```

tomcur (approved):
LGTM!

parley_tests/tests/emoji/mod.rs
```diff
@@ -0,0 +1,710 @@
...
+    assert_emoji(TestEntity {
+        sequence: &[
+            0x1F600, // GRINNING FACE
+        ],
+        categories: &[EmojiSegmentationCategory::EmojiPresentation],
+        style: EmojiPresentationStyle::Emoji,
+        is_emoji: true,
+        has_vs: false,
+    });
+}
```
DJMcNab: I've moved them in 296d9e64460a99d29631467e22aed1c8c00bbfca

parley_engine/src/analysis.rs
DJMcNab: We do use them once immediately in analysis, specificity in the check added in https://github.com/linebender/parley/pull/781

README.md
```diff
@@ -99,11 +99,13 @@ Licensed under either of
...
-Additionally, some files used for tests and benchmarks are under different licenses.
-See `./parley_dev/README.md` for details.
-
+Some code in this repository is not available under these licenses:
+
+- `parley_emoji` is licensed only under its MIT license (which has a different attribution to Parley's MIT License).
+  See `./parley_emoji/README.md` for details.
+- `parley_tests/tests/emoji` is licensed only under the Apache License, Version 2.0 (and not the MIT license).
+  See `./parley_tests/tests/emoji/README.md` for details.
+- Additionally, some files used for tests and benchmarks are under different licenses.
```
conor-93: Probably don't need "Additionally, " here in a list

.github/copyright.sh
```diff
@@ -7,7 +7,7 @@
 #   -g "!src/special_directory"
 
 # Check most standard Rust source files.
-output=$(rg "^// Copyright (19|20)[\d]{2} (.+ and )?the Parley Authors( and .+)?$\n^// SPDX-License-Identifier: Apache-2\.0 OR MIT$\n\n" --files-without-match --multiline -g "*.rs" -g "!parley_emoji/**" .)
+output=$(rg "^// Copyright (19|20)[\d]{2} (.+ and )?the Parley Authors( and .+)?$\n^// SPDX-License-Identifier: Apache-2\.0 OR MIT$\n\n" --files-without-match --multiline -g "*.rs" -g "!parley_emoji/**" -g "!parley_tests/tests/emoji/**" .)
```
conor-93, line 10: nit: Maybe not worth the slight headache but the only difference between the regexes here are the license identifiers and the target dir patterns, a helper could create those. Also applies to the printing of copyright header warnings.

parley_tests/tests/emoji/mod.rs
```diff
@@ -0,0 +1,711 @@
+// Copyright 2026 the Parley Authors and Google LLC
```
conor-93: Are we sure this is under the copyright of Google LLC? https://github.com/google/emoji-segmenter/blob/main/CONTRIBUTING.md says the author retains the copyright, so maybe put Christian Hansen here, as he authored [the tests](https://github.com/google/emoji-segmenter/pull/18/changes#diff-4c0ddae345d87fcf0aa681d1e998716d35ae7a892e0b1e71dfec2840c2324709) we're adapting here?

conor-93 (commented):
Just for https://github.com/linebender/parley/pull/811/changes/296d9e64460a99d29631467e22aed1c8c00bbfca

parley_tests/tests/emoji/mod.rs
DJMcNab: This header was based on the header from Google's Emoji Segmenter (see https://github.com/google/emoji-segmenter/blob/51dea64f6ecf6dbcd633845343e0c72b937d04e8/emoji_presentation_scanner.c#L3-L16).
The copyright headers we add don't actually have any weight.

Happy to change it to add Christian as well though.

.github/copyright.sh
DJMcNab: I really don't want to write a shell script to make a shell script. I'd suggest just making this a follow-up?

conor-93: Fair enough. I'd be happy just leaving it as-is then to be honest.

PR #812, opened 2026-09-14
Implement `WhiteSpaceCollapse::PreserveBreaks` whitespace collapsing

nicoburns:
LLM Contributions: Generated with Fable 5.1 Low. Human reviewed and iterated on.

Implements `WhiteSpaceCollapse::PreserveBreaks`. `PreserveBreaks` collapsed whitespace similar to `Collapse` except the "segment breaks" (hard newlines) are preserved.

This is the last remaining whitespace collapsing mode. We also need to implement `break-spaces`, but that only differs in terms of hanging. For collapsing it maps to `Preserve`.

**Changelog**

> ### Added
>
> - Support for `WhiteSpaceCollapse::PreserveBreaks`


parley/src/style/mod.rs
```diff
@@ -20,10 +20,26 @@ pub use styleset::StyleSet;
...
+impl WhiteSpaceCollapse {
+    /// Whether `c` is whitespace that this mode collapses.
+    pub(crate) fn is_collapsible(self, c: char) -> bool {
+        match self {
+            Self::Collapse => c.is_ascii_whitespace(),
+            Self::Preserve => false,
+            Self::PreserveBreaks => matches!(c, ' ' | '\t'),
+        }
+    }
 }
```
tomcur: I like this method, but maybe it should move to the `whitespace.rs` module. Admittedly it's then slightly unfortunate that the enum and the impl are in different modules; but `whitespace.rs` implements other such logic, and it's probably more important to try to keep those together.

parley/src/resolve/tree.rs
```diff
@@ -123,20 +123,19 @@ impl<B: Brush> TreeStyleBuilder<B> {
         match self.white_space_collapse {
             WhiteSpaceCollapse::Preserve => {
                 if uncommitted_text.starts_with(is_segment_break) {
-                    // Pending whitespace is always from a `WhiteSpaceCollapse::Collapse` span,
-                    // and following CSS Text 4 § 4.3.1 Rule 1 must be removed if it immediately precedes
-                    // a preserved segment break.
+                    // Pending whitespace is always from a `WhiteSpaceCollapse::Collapse` or
+                    // `WhiteSpaceCollapse::PreserveBreaks` span, and following CSS Text 4 § 4.3.1
+                    // Rule 1 must be removed if it immediately precedes a preserved segment break.
```
tomcur: Very nitty, but this reflow is unfortunate and reads like "Rule 1 must be removed".

Fine to leave this, but imo it'd be better to just break the max columns in this case.

```suggestion
                    // Pending whitespace is always from a `WhiteSpaceCollapse::Collapse` or
                    // `WhiteSpaceCollapse::PreserveBreaks` span, and following CSS Text 4 § 4.3.1 Rule 1
                    // must be removed if it immediately precedes a preserved segment break.
```

parley/src/tests/test_builders.rs
```diff
@@ -77,6 +77,34 @@ fn build_layout_with_style_runs(
     layout
 }
 
+#[test]
+fn tree_preserve_breaks_across_runs() {
```
tomcur: This test seems redundant given the `preserve_breaks_collapses_across_text_and_style_boundaries` test has a strict superset of cases.

tomcur (approved):

parley/src/style/mod.rs
```diff
@@ -20,10 +20,15 @@ pub use styleset::StyleSet;
 
 use crate::util::nearly_eq;
 
-#[derive(Debug, Clone, Copy)]
+/// How source whitespace is collapsed by a tree builder.
```
tomcur: This comment is quite specific. It should probably just say only the tree builder implements this for now.

parley/src/style/mod.rs
nicoburns: Moved!

PR #814, opened 2026-09-14
Make `WhiteSpaceCollapse` a style

nicoburns:
LLM Contributions: Generated with Fable 5.1 Low.

- Builds on https://github.com/linebender/parley/pull/812

Converts `WhiteSpaceCollapse` to be a proper style rather than being handled in a custom way on `TreeBuilder`. This makes the style available to layout (line breaking / content widths computation) and is a pre-requisite for implementing proper mode-aware whitespace hanging.

(this turned out to be  surprisingly straightforward)

**Changelog**

> ### Added
>
> - Breaking change: `WhiteSpaceCollapse` is now a style rather than a special method on the `TreeBuilder`


parley/src/layout/mod.rs
```diff
@@ -65,6 +65,8 @@ pub struct Style<B: Brush> {
     pub(crate) overflow_wrap: OverflowWrap,
     /// Per-cluster text-wrap-mode setting
     pub(crate) text_wrap_mode: TextWrapMode,
+    /// Per-cluster whitespace collapsing and hanging behavior.
```
tomcur, line 68: Orthogonal to this PR, as some other style options also say "per-cluster", but I'm not sure what that means exactly.

parley/src/style/mod.rs
```diff
@@ -20,12 +20,16 @@ pub use styleset::StyleSet;
 
 use crate::util::nearly_eq;
 
-/// How whitespace is collapsed.
-#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+/// Whitespace collapsing and hanging behavior.
+///
+/// Tree builders transform source text. Ranged and style-run builders accept already
+/// transformed text and use this property only during layout.
```
tomcur:
```suggestion
/// [`TreeBuilder`] transforms the source text according to this property. [`RangedBuilder`]
/// and [`StyleRunBuilder`] don't transform the text, and use this only for whitespace handling
/// during layout.
```

parley/src/builder.rs
```diff
@@ -38,6 +38,8 @@ impl BuilderOptions<'_> {
 }
 
 /// Builder for constructing a text layout with ranged attributes.
+///
+/// Text must already be whitespace-processed; source text and byte ranges are preserved.
```
tomcur: I think it may be better to say this builder does not transform the text, so `WhiteSpaceCollapse` is only used for whitespace handling during layout. Using the builder itself does not require the caller to have performed whitespace collapsing.

parley/src/builder.rs
```diff
@@ -103,6 +105,8 @@ impl<'b, B: Brush> RangedBuilder<'b, B> {
 
 /// Builder for constructing a text layout from a style table and
 /// indexed style runs.
+///
+/// Text must already be whitespace-processed; source text and byte ranges are preserved.
```
tomcur: See the comment on `RangedBuilder`.

parley/src/resolve/tree.rs
```diff
@@ -374,13 +373,62 @@ mod tests {
         let style = &styles[space.style_index as usize];
         assert_eq!(style.brush, 1);
         assert_eq!(style.text_wrap_mode, TextWrapMode::Wrap);
+        assert_eq!(style.white_space_collapse, WhiteSpaceCollapse::Collapse);
+    }
+
+    #[test]
+    fn nested_whitespace_modes_restore_and_reuse_styles() {
```
tomcur: I think this test doesn't really exercise something novel. It looks like it effectively restates existing whitespace collapsing tests plus exercises the general style system (pushing and popping of styles). I think I'd just drop it.

parley/src/tests/test_builders.rs
```diff
@@ -77,6 +77,49 @@ fn build_layout_with_style_runs(
     layout
 }
 
+#[test]
+fn tree_out_of_flow_boxes_commit_text_without_splitting_words_or_whitespace() {
```
tomcur: This could probably be merged into/next to the existing `collapsible_whitespace_crosses_spans_and_inline_boxes` test, which tests the same for in-flow inline boxes.

DJMcNab:
Have we had the discussion about what the longer-term plan is? The current proposal that one of the builders does whitespace collapsing and the others don't at all is extremely odd.

Also, has there been any thinking about how to handle some of the other line-specific cases - e.g. steps 1 and 3 of https://drafts.csswg.org/css-text-3/#white-space-phase-2?

(I don't think either of these concerns block this PR, although I think I'd want our documentation story to be extremely clear wrt to the first point).

parley/src/style/mod.rs
nicoburns: Went with something slightly different. Mostly because I'd already rewritten this comment before I saw this.

parley/src/layout/mod.rs
nicoburns: Good point, but yeah let's handle this separately.

tomcur:
> Have we had the discussion about what the longer-term plan is? The current proposal that one of the builders does whitespace collapsing and the others don't at all is extremely odd.

I agree. I think longer term we probably want to think about general text transformations and mapping back to the original text. There's a slight difference between whitespace collapse and some other text transformations, though: in browsers, if you transform e.g. to upper case, then when you copy the text, you copy the original without that transformation. Collapsed whitespace, on the other hand, is gone also when you copy.

Then there are also inline objects, which at some point we might insert as Unicode Object Replacement Characters into the analyzed string for correct bidi reordering and such.

parley/src/resolve/tree.rs
nicoburns: Dropped!

parley/src/tests/test_builders.rs
nicoburns: Merged

parley/src/style/mod.rs
```diff
@@ -20,12 +20,16 @@ pub use styleset::StyleSet;
 
 use crate::util::nearly_eq;
 
-/// How whitespace is collapsed.
-#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+/// Whitespace collapsing and hanging behavior.
+///
+/// Note: Currently only the `TreeBuilder` collapses white space. The `RangedBuilder` and `StyleRunBuilder`
+/// builders expect already transformed text and use only this property to influence whitespace hanging behaviour
```
tomcur:
```suggestion
/// Note: Currently only the [`TreeBuilder`](crate::TreeBuilder] collapses white space. The
/// [`RangedBuilder`](crate::RangedBuilder) and [`StyleRunBuilder`](crate::StyleRunBuilder)
/// builders expect already-transformed text and use this property only to influence whitespace hanging behaviour.
```

tomcur (approved):
Looks good; thank you!

nicoburns:
> Have we had the discussion about what the longer-term plan is? The current proposal that one of the builders does whitespace collapsing and the others don't at all is extremely odd.

We have discussed that long-term we want proper/generalized text transformation support (where exactly that sits is TBD) with an offset mapping table, etc. Whitespace collapsing could(/should?) probably be handled as part of that long-term.

Until then, I suggest we leave things as are for expediency. The status quo is that the `TreeBuilder` is effectively the "web semantics" builder (with whitespace collapsing being pretty web-specific), with the other two following a model more common in traditional GUIs, word processors, etc.

Long-term we could also consider dropping the `RangedBuilder` and `StyleRunBuilder` entirely if we wanted to (or making the `RangedBuilder` a wrapper around the `TreeBuilder`). Non-overlapping ranges are a strict subset of a tree model, and overlapping ranges can be resolved into non-overlapping ranges (as the `RangedBuilder` does).

> Also, has there been any thinking about how to handle some of the other line-specific cases - e.g. steps 1 and 3 of https://drafts.csswg.org/css-text-3/#white-space-phase-2?

I've thought about this, but I'm not yet sure how invasive the change will be and thought it made sense to fix the bits of whitespace handling which fit within the existing model first. We may also want to think about implementing UAX#9 L1 as part of that, as I think that code will sit in a similar place.

PR #815, opened 2026-09-14
Fix out-of-flow box indexing issue

tomcur:
LLM Contributions: Review.

The box's index is taken from the committed text, so text has to be committed for both in-flow and out-of-flow boxes.

**Changelog**

> ## Parley
> ### Fixed
>
> - When using the `TreeBuilder`, out-of-flow boxes are now correctly positioned within the text.


nicoburns (approved):
LGTM.

---

Does this issue exist in `0.11`? If not then the changelog entry is not required.

tomcur:
Yeah, I believe so:

https://github.com/linebender/parley/blob/v0.11.1/parley/src/builder.rs#L227-L239

Incidentally, an LLM also appears to have spotted this in your #814 (but it's quite orthogonal).

PR #816, opened 2026-09-14
Tighten `bytemuck` feature forwarding

tomcur:
LLM Contributions: Review.

Quick follow-up.

**Changelog: None**


nicoburns (approved):
I think there's also a feature on `font-types`. Not sure if we want to enable that?

tomcur:
> I think there's also a feature on `font-types`. Not sure if we want to enable that?

Good question, and I don't know. I think at least the forwarding to `parlance` you had introduced definitely makes sense as it's quite central in the API (and this just follows the convention of the workspace to forward to all workspace members that themselves forward), so merging as-is for now.

Issue #14, opened 2022-01-08
panic on linux


dvc94ch:
playing around with piet-gpu and parley using your piet-gpu integration. however I haven't gotten it to work yet. my guess is that since fount mentions it only supports windows and macos that this is currently expected.

```
MESA-INTEL: warning: Performance support disabled, consider sysctl dev.i915.perf_stream_paranoid=0

thread 'main' panicked at 'range end index 1 out of range for slice of length 0', /home/dvc/cloudpeer/parley/src/layout/line/greedy.rs:510:26
stack backtrace:
   0: rust_begin_unwind
             at /rustc/f1edd0429582dd29cccacaf50fd134b05593bd9c/library/std/src/panicking.rs:517:5
   1: core::panicking::panic_fmt
             at /rustc/f1edd0429582dd29cccacaf50fd134b05593bd9c/library/core/src/panicking.rs:100:14
   2: core::slice::index::slice_end_index_len_fail
             at /rustc/f1edd0429582dd29cccacaf50fd134b05593bd9c/library/core/src/slice/index.rs:41:5
   3: <core::ops::range::Range<usize> as core::slice::index::SliceIndex<[T]>>::index
             at /rustc/f1edd0429582dd29cccacaf50fd134b05593bd9c/library/core/src/slice/index.rs:240:13
   4: core::slice::index::<impl core::ops::index::Index<I> for [T]>::index
             at /rustc/f1edd0429582dd29cccacaf50fd134b05593bd9c/library/core/src/slice/index.rs:15:9
   5: <alloc::vec::Vec<T,A> as core::ops::index::Index<I>>::index
             at /rustc/f1edd0429582dd29cccacaf50fd134b05593bd9c/library/alloc/src/vec/mod.rs:2496:9
   6: parley::layout::line::greedy::commit_line
             at /home/dvc/cloudpeer/parley/src/layout/line/greedy.rs:510:26
   7: parley::layout::line::greedy::BreakLines<B>::break_next
             at /home/dvc/cloudpeer/parley/src/layout/line/greedy.rs:222:12
   8: parley::layout::line::greedy::BreakLines<B>::break_remaining
             at /home/dvc/cloudpeer/parley/src/layout/line/greedy.rs:257:15
   9: parley::layout::Layout<B>::break_all_lines
             at /home/dvc/cloudpeer/parley/src/layout/mod.rs:110:9
  10: <piet_parley::ParleyTextLayoutBuilder as piet::text::TextLayoutBuilder>::build
[...]

PR #819, opened 2026-09-15
fix: decouple UAX #29 word boundaries from line break opportunities

GoCoder7:
LLM Contributions: None.

### Summary

Decouple UAX #29 word boundaries from UAX #14 line break opportunities in `Cluster::is_word_boundary()`.

### Problem

In `analyze_text`, line break boundaries (`Boundary::Line` / `Boundary::Mandatory`) and word boundaries (`Boundary::Word`) were merged into a single `Boundary` enum where line breaks took precedence.

Because CJK scripts (including Korean Hangul) allow soft line breaks between almost every character under UAX #14 rules, nearly every Korean character was marked as `Boundary::Line`. Consequently, `Cluster::is_word_boundary()` (which checked `boundary != Boundary::None`) treated every Korean syllable as a word boundary. In text editors and downstream consumers (like Blitz/Dioxus), word navigation (`Option + Left/Right`) and word deletion (`Option + Backspace`) stopped at every single syllable instead of jumping word by word across whitespace.

### Fix

- Retain UAX #29 `is_word` boundary information from `WordSegmenter` in a dedicated bit (`WORD_BOUNDARY_SHIFT`) within `CharInfo`.
- Pass this flag to `ClusterInfo::is_word_boundary`.
- Update `Cluster::is_word_boundary()` to check this dedicated flag rather than line break boundaries.
- Line break behavior (`Boundary::Line`) remains unchanged for layout wrapping.

**Changelog**

> ### Fixed
>
> - Fixed CJK and Korean word boundary navigation in `Cluster::is_word_boundary()`, properly distinguishing UAX #29 word boundaries from UAX #14 line break opportunities.


tomcur (approved):
This looks like an improvement to me, thanks for the fix! I'll open two follow-ups to this.

Before this change, we were accidentally getting word boundaries "nearly right" in the case where dictionaries were *not* available, because following Unicode, the boundaries then need to be at every ideograph. In the common case that coincides with the line break opportunities, which we were then seeing as word boundaries. The ICU4X word segmenter doesn't do that, though, so one of the follow-ups fixes a degradation this PR causes to the boundary detection when there are no dictionaries available (i.e., especially without the `complex-scripts` feature, but note that also *with* that feature we don't carry dictionaries for some scripts).

(Sorry for the mess of commits; I had to use GitHub's web UI because this PR came from an organization's fork, meaning I couldn't push to it from my local checkout.)

PR #820, opened 2026-09-15
Implement `WhiteSpaceCollapse::BreakSpaces`

nicoburns:
LLM Contributions: Generated with Fable 5.1 Low.

Builds on:

- https://github.com/linebender/parley/pull/812
- https://github.com/linebender/parley/pull/814

Adds support for `WhiteSpaceCollapse::BreakSpaces`. This maps to the existing `Preserve` mode for actual whitespace collapsing, but has it's own unique behaviour during line-breaking and hanging. This is the remaining (widely implemented) value of `WhiteSpaceCollapse` needed to complete our whitespace handling.

This is an additional ~19% perf hit on `calculate_content_widths`, but https://github.com/linebender/parley/pull/823 is able to recover that and more with a fast-path.

Blitz PR (updates to Parley with this PR): https://github.com/DioxusLabs/blitz/pull/884

**Performance comparison**

Numbers for `calculate_content_widths`

| build | median ms | min ms | vs 0.11 | vs main |
| --- | --- | --- | --- | --- |
| v0.11.0 | 0.67–0.68 | 0.66 | +0% | −41% |
| main (3fdcdca) | 1.14–1.15 | 1.12 | +70% | +0% |
| #820 (this PR) | 1.35–1.37 | 1.31 | +100% | +19% |
| #822 (opt ccw) | 0.77 | 0.75 | +14% | −33% |
| #823 (#820 + #822) | 0.81–0.82 | 0.78 | +20% | −29% |

**Changelog**

> ### Added
>
> - Support for `WhiteSpaceCollapse::BreakSpaces`


parley/src/layout/whitespace.rs
```diff
@@ -39,13 +39,43 @@ pub(crate) fn whitespace_hangs<B: Brush>(whitespace: Whitespace, style: &Style<B
...
+            character.info.whitespace(),
+            Whitespace::Space | Whitespace::Tab | Whitespace::IdeographicSpace
+        )
+}
+
+/// Whether a soft line break opportunity exists before the character at `index`.
+///
+/// Following [CSS Text 4 § 5.1][css-break-spaces], `white-space-collapse: break-spaces` permits
+/// wrapping after each preserved space or tab, but not before the first one. Other Unicode
+/// separators retain their UAX #14 opportunities.
```
tomcur, line 64: I believe this follows § 4.3.1, not § 5.1.

```suggestion
/// Following [CSS Text 4 § 4.3.1][css-break-spaces], `white-space-collapse: break-spaces` permits
/// wrapping after each preserved space or tab, but not before the first one. Other Unicode
/// separators retain their UAX #14 opportunities.
```

§ 4.3.1 says the following.

> For [break-spaces](https://www.w3.org/TR/css-text-4/#valdef-white-space-collapse-break-spaces), a soft wrap opportunity exists after every space and every tab.

Further, does it need to say "but not before the first one"? I think just "after" carries the same meaning.

```suggestion
/// Following [CSS Text 4 § 4.3.1][css-break-spaces], `white-space-collapse: break-spaces` permits
/// wrapping after each preserved space or tab. Other Unicode separators retain their UAX #14
/// opportunities.
```

parley/src/layout/whitespace.rs
```diff
@@ -39,13 +39,43 @@ pub(crate) fn whitespace_hangs<B: Brush>(whitespace: Whitespace, style: &Style<B
...
     }
 }
 
+fn is_break_space<B: Brush>(character: Character, styles: &[Style<B>]) -> bool {
+    styles[character.style_index as usize].white_space_collapse == WhiteSpaceCollapse::BreakSpaces
+        && matches!(
+            character.info.whitespace(),
+            Whitespace::Space | Whitespace::Tab | Whitespace::IdeographicSpace
+        )
+}
```
tomcur, line 58: It might be clearer to just fold this into `soft_line_break`, as in, just inlining the code there. That leaves just one place the document. (I was a bit confused by what this function was calculating.)

parley_tests/tests/whitespace.rs
```diff
@@ -107,3 +120,258 @@ fn overflowing_whitespace_hangs_without_adding_a_break_opportunity() {
...
+                nearly_eq(line.metrics().hanging_advance, 0.);
+                nearly_eq(layout.calculate_content_widths().max, full);
+                layout.align(Alignment::Right, AlignmentOptions::default());
+                nearly_eq(layout.get(0).unwrap().metrics().offset, 20.);
+                layout.align(Alignment::Center, AlignmentOptions::default());
+                nearly_eq(layout.get(0).unwrap().metrics().offset, 10.);
+            }
+        }
+    }
+}
```
tomcur, line 213: This is running 30 cases in CI, where most code paths probably aren't very interesting. Maybe prune some separators and endings?

tomcur (approved):
LGTM! Please do take a look at my comments on `soft_line_break` and `is_break_space`.

Non-blocking, but it seems like there's quite a bit of overlap between the added tests. I didn't look too closely at them, but just noting this as there is a small maintenance cost to tests. I'll leave it to your judgement. It feels like tests 2 and 4 could be merged, 7 could perhaps also be folded in there (or just dropped), and 1 and 3 could probably be merged (checking the advance and checking the source is retained).

nicoburns:
Superceded by https://github.com/linebender/parley/pull/833

PR #821, opened 2026-09-16
Remove direct `glifo` dependency

nicoburns:
LLM Contributions: Generated with Devin (Fable 5.1 Low)

**Changelog: None**

## Changes made

- Remove direct `glifo` dependency
- Upgrade `vello_cpu` to latest `0.2.0`
- Snapshot comparison is now exact. `Pixmap::into_png` unpremultiplies with rounding but `Pixmap::from_png` re-premultiplies with truncation, so a decoded snapshot was never bit-identical to the pixmap it was written from. We. now roundtrip new snapshots through png encode/decode to work around this. Allows us to entirely remove the tolerance for several snapshots (otherwise had to increase from ~5 to over 100)

Most of the changes are mechanical changes around the switch from `glifo` to `vello_cpu` in the test runner, and adapting to the new `vello_cpu` API in examples.

## Snapshots

- Almost all updated, but most are just AA noise from the vello_cpu upgrade.
- The `draw_colr_emoji-{nohint,hint,2x_nohint,2x_hint}_skew` COLR glyphs are now skewed noticeably less than before. This was an intentional upstream bugfix in glifo.
- The macOS bitmap emoji bitmaps are resampled with a softer filter (looks better to me)

AI-generated diff images included below for the snapshots with non-AA changes:

colr skew old vs new

macOS emoji old vs new


parley_tests/tests/util/env.rs
```diff
@@ -264,6 +254,13 @@ impl TestEnv {
             ));
         }
 
+        // `Pixmap::into_png` unpremultiplies with rounding but `Pixmap::from_png`
+        // re-premultiplies with truncation, so a decoded snapshot is not bit-identical
+        // to the pixmap it was written from. Pass the current image through the same
+        // round-trip so both sides carry the same (deterministic) error.
+        let current_img =
+            Pixmap::from_png(Cursor::new(current_img.clone().into_png().unwrap())).unwrap();
```
DJMcNab: This makes me sad. Is there an upstream Vello issue for this, if either `from_png` or `into_png` is wrong?

DJMcNab: Separately, can you speak to why you chose to run conversion into png twice for snapshots we regenerate? It's defensible, but it seems a nice mirroring for the png data we actually save to be the png data we compare using.

parley_tests/tests/util/renderer.rs
```diff
@@ -198,17 +196,24 @@ pub(crate) fn draw_layout(
...
+    Renderer {
+        ctx: renderer,
+        resources,
+    }
 }
 
 /// Render the layout to a [`Pixmap`].
 ///
 /// If given [`RenderingConfig::size`] is not specified, [`Layout::width`] and [`Layout::height`]
 /// are used.
```
DJMcNab: Where is the RenderingConfig argument?

DJMcNab (approved):
Thanks!

parley_tests/tests/util/env.rs
nicoburns: Not sure. One for ROH maybe? One is adding 0.5 and one isn't.

nicoburns: It was a smaller diff, but yeah, we should probably avoid converting twice.

DJMcNab (approved):
Thanks! I haven't carefully reviewed the snapshots, but I guess there's no way this can actually change layout 🙃

PR #822, opened 2026-09-16
Optimize `calculate_content_widths`

nicoburns:
LLM Contributions: Generated with Fable 5.1 Low

Builds on:

Gets almost all of the 0.11 performance back for `calculate_content_widths` (only 10% slower than 0.11 with this PR vs ~60% slower on main). A little slower is expected as this function is doing a lot more work now (for correct whitespace hanging).

Draft pending whitespace handling feature work landing. Will write-up later.


tomcur (approved):
This is a nice speed-up!

I'm also working on some changes that may grow the `ShapedCluster` flags, which means either some of this would have to change again, or perhaps growing the flags doesn't cost too much. I was thinking it may be possible to remove style indices, which would free up bits.

An LLM flagged a subtle issue, which is slightly orthogonal, so I'll just open a follow-up for that (edit: #842).

Completely non-blocking, but I think the spaced path could be roughly the same shape, so the atoms loop could in theory be dropped.

PR #823, opened 2026-09-17
Optimize `calculate_content_widths` (`break-spaces` edition)

nicoburns:
Variant of https://github.com/linebender/parley/pull/822 (Optimize `calculate_content_widths`) on top of https://github.com/linebender/parley/pull/820 (Implement `WhiteSpaceCollapse::BreakSpaces`)

Draft pending #820 landing. Will write up later. This takes `calculate_content_widths` down from ~1.3ms to ~0.8ms (vs. ~0.67ms on `0.11`). Those may seem like small numbers, but this a very hot path that can get called thousands of times for a single page layout.


nicoburns:
https://github.com/linebender/parley/pull/822 is now the relevant PR as we merged https://github.com/linebender/parley/pull/833 rather than https://github.com/linebender/parley/pull/820

PR #824, opened 2026-09-17
Simplify CRLF model

tomcur:
LLM Contributions: Review.

In #667 we introduced a look-ahead in the line breaker, such that a CR character followed by an LF would no longer cause two forced breaks.

Later, we fixed our grapheme handling and the line breaker now sees CRLF as a unit. The look-ahead is no longer necessary for the common case.

This proposes removing the check. I think this improves consistency, and also makes it somewhat simpler to perform an optimization that I'll open a PR for soon.

Separately, note that while we currently treat CR as a forced break, CSS Text 4 says we should treat it as a space... this PR does not move us to that model -- and Gecko/Blink also don't do that, they mostly just ignore the CR -- but it's something to keep in mind.

The reason I say this improves consistency, is because #667 intentionally forms two breaks if an inline box is placed inside the sequence. That means the check now only prevents the CR from breaking if it's in a different run from the LF; and that happens precisely when the user creates a shaping-relevant style boundary, or, in fact, places an inline box inside the sequence. Across such item boundaries, we intentionally reset grapheme segmentation; for example, like browsers, we also don't form emoji across these boundaries.

(Orthogonally, do note we may work on grapheme segmentation a bit more: we probably should actually run the segmentation *after* itemization, instead of intersecting the segmentation with items.)

**Changelog: None**


nicoburns (approved):
I think I'm happy with this. And no WPT tests affected. I would like an explanation for why the unit test was removed. Presumably that means there is a case we're now not handling (even if not important)?

tomcur:
> I think I'm happy with this. And no WPT tests affected. I would like an explanation for why the unit test was removed. Presumably that means there is a case we're now not handling (even if not important)?

The test was specifically testing that we were treating CRLF as a unit even if it got split across two runs due to itemization through a shaping-relevant style change. That's opposite from how we handle items elsewhere: we intentionally do not form graphemes across item boundaries, and CRLF is a grapheme.

nicoburns:
> CRLF is a grapheme

Hmm... I guess. Either way, this seems like an unlikely edge case.

tomcur:
> Either way, this seems like an unlikely edge case.

Indeed.

PR #827, opened 2026-09-17
Render a hyphen when a line breaks at a soft hyphen

### LINKED ISSUE #704: Handle soft hyphens
These are treated as an in-word line breaking opportunity (possibly uniquely without breaking kerning?) but not as a visual space.
If the line breaking opportunity is taken, they're rendered as a hyphen.

Currently we treat these as a line breaking opportunity, but don't render the hyphen (I believe).

sawy3r:
LLM Contributions: This branch was co-authored with Claude Code — the implementation and the tests were written with it.

## Problem description

Parley treats a soft hyphen (U+00AD) as an in-word line breaking opportunity, but a line broken there ends with nothing at all: `hyphe\u{AD}nation` wrapped inside the word gives `hyphe` / `nation`, with no hyphen anywhere. This makes the break render as a hyphen.

## Which glyph

The hyphen is the glyph the font's cmap maps for **U+2010 HYPHEN**, falling back to **U+002D HYPHEN-MINUS** when the font has no U+2010. It comes from the font the soft hyphen itself was shaped with, at that run's font size and variation coordinates, and it is looked up directly rather than shaped.

The soft hyphen's own cmap entry is never used, even in the fonts that give it a distinct glyph — that matches Chrome, and since HarfBuzz drops the soft hyphen during shaping there is no such glyph to reuse in the first place. When a font maps neither U+2010 nor U+002D there is no hyphen and the break behaves exactly as it does today.

## The advance

The hyphen takes part in the fit decision: a break opportunity right after a soft hyphen is only an opportunity if the line's advance up to the soft hyphen *plus the hyphen* fits in the maximum advance. When it doesn't fit, the opportunity is skipped as if the soft hyphen weren't there and the breaker uses whatever earlier opportunity it has, or overflows as before.

Once a break is taken, the advance is part of the line's measured content: it's included in `LineMetrics::advance`, in the last `Run` of the line, in the `GlyphRun` that ends it and in the soft hyphen's `Cluster`, so cluster advances still sum to run advances and run advances to the line advance. Alignment and justification work on the wider line.

It stays a glyph rather than a character. The line's text range still ends at the soft hyphen, cluster counts and text ranges are unchanged, and a soft hyphen that isn't at the end of a line is still glyph-less and zero-width — a paragraph that fits on one line lays out exactly as before.

In visual order the hyphen sits at the line's *end* edge: last for a left-to-right run, first for a right-to-left one.

## How it's wired

* `parley_engine::hyphen_glyph` does the cmap lookup and the advance. It lives in the engine because `skrifa` is a hard dependency there and only an optional one of `parley`.
* The line breaker records the hyphen on the *saved* break opportunity, with the saved advance already grown by it, and simply doesn't save an opportunity whose hyphen doesn't fit. Reverting to a saved state therefore reverts the hyphen with it, so re-breaking a layout at different widths is unaffected. `commit_line` moves the hyphen onto the line's logically last text run.
* `Cluster::advance`, `Run::advance`, the `Glyphs` iterator behind `Run::glyphs_in` and the atom iterator behind `Line::items` all account for it, so renderers get the glyph and the metrics agree.

## Not covered

`ContentWidths` is untouched: the min-content width of a word containing a soft hyphen still ignores both the break and the hyphen. That felt like a separate change, but I'm happy to fold it into this PR if you'd rather have it together. Character-count breaking via `break_next_with_length` also doesn't render hyphens.

## Tests

`parley_tests/tests/soft_hyphen.rs` adds a snapshot of `hyphe-` / `nation` / `word` broken at a narrow width, plus assertions for the glyph id, advance and baseline position, the cluster/run/line advance sums, the unbroken layout being unchanged, the fit rule in both directions (just too narrow for the hyphen, and just wide enough), and the right-to-left ordering.

**Changelog**

> ### Fixed
>
> - A line broken at a soft hyphen (U+00AD) now renders a hyphen at the end of the line.

Fixes #704.


DJMcNab:
I have no judgement on the code yet (the PR description pretty much matches my understanding, but I haven't checked the code). I'll aim to review this next week.

Excluding content widths doesn't seem right to me, but I'm not certain either way.

However, as a note, please remove the commit footer which names Claude as a co-author; you are fully responsible for all of the PR's content in Linebender projects, not the tools you happened to use. I don't expect you to have known this - we don't currently have this written as a defined policy.

PR #828, opened 2026-09-18
Split shaping runs when the line height changes

virtualritz:
A run's line height is read from the style its item was shaped under, but a style change that alters *only* the line height does not split the item. One run then covers both styles, and the text before the change is laid out with the later style's line height.

Font size, locale, variations, features, letter spacing and word spacing all split the item already; line height is the one run-affecting property missing from that list.

## Repro

```rust
let mut builder = lcx.ranged_builder(&mut fcx, "tall\nshort", 1., false);
builder.push_default(StyleProperty::LineHeight(LineHeight::Absolute(40.)));
builder.push(StyleProperty::LineHeight(LineHeight::Absolute(10.)), 5..10);
```

Line 0 holds only the 40.0 text and reports 10.0; the paragraph is 20.0 tall instead of 50.0. Swapping the two values swaps the result, so both lines take the value of whichever style the single run was shaped under.

In an application this shows up as a mixed-size paragraph laid out at the wrong height with every baseline moved: 21 px text at line height 25.2 followed by 13 px at 15.6 came out 31.2 tall instead of 40.8. Font sizes stay correct, because a size change does split; per-glyph brushes stay correct, because glyph runs are split by the per-cluster style index. Only run-level metrics take the wrong style, which makes it easy to miss.

## Fix

One condition: split when `line_height` differs, next to the properties that already do.

## Test

`tests::test_layout::line_height_is_per_run` asserts the two lines of the repro report 40.0 and 10.0 and the paragraph is 50.0 tall. The rest of the suite is unchanged (71 tests pass).

Note for 0.11.x: on that release the item also kept the style index of the character that *ended* it, so the finished run was attributed to the following style. `main` already fixed that half in the itemizer rework; only the split condition is needed here. The 0.11.1 backport carrying both halves is on branch `citra/line-height-0.11.1` of the fork, if a point release is wanted.


DJMcNab:
Welcome!
Can you speak to how this is different to https://github.com/linebender/parley/pull/761?

Also, I'm interested in learning what tool you used to open this pull request. We need to report to them their failure to surface our pull request template. Please edit the description to follow it.

virtualritz:
Yeah, we hit the same issue that #761 fixes and I had a clanker file the PR w/o checking open  ones. Sorry about that. 
I will follow up on that PR.

For context, we use `parely` in our [Creator Studio app](https://www.phyron.com/inventory-videos) to replace Skia for text rendering (and other parts of the Linebender stack for vectors).
Foremost as we can't have cached glyps (we render text with motion blur accross any axis or xform and Skia e.g. caches with 1px size in y and 0.25px in x which then stutters).

PR #829, opened 2026-09-18
Fix complex script word boundaries in the absence of dictionaries

tomcur:
LLM Contributions: Investigation, review, added tests.

For complex scripts that require a dictionary, ICU4X reports runs of characters as a single word. However, according to [UAX #29 Rule WB999](https://www.unicode.org/reports/tr29/#WB999) (which applies if there is no dictionary), there should be word boundaries around every ideograph. Potentially this is something that could be improved in ICU4X; in the meantime, I propose to patch this up ourselves.

Prior to https://github.com/linebender/parley/pull/819, we were accidentally quite close for this fallback case. As https://github.com/linebender/parley/pull/819 made it such that we correctly classify word boundaries separately from line boundaries, we are now wrong for complex scripts if we don't have the dictionary (which is always, if we don't have `complex-scripts`; and otherwise we're still wrong for complex scripts that don't get a dictionary, like (apparently) Ahom).

This spends one bit in `parley_data` to track the relevant Unicode property. I did consider implementing this check with some heuristics, but it didn't feel great to spend serious runtime on processing some table and still be incorrect for at least a handful of code points. I've added a test that starts failing if the ICU4X segmenter implements WB999 in the future, which says to clean all of this up again (including freeing the bit from `parley_data`). Do note that, unfortunately, `has_word_dictionary` is also just approximate, but the alternative is tracking an ICU4X implementation detail.

I *hope* the samples in the added tests are correct. I don't speak those languages, so I'm not fully sure. While most other LLM-generated tests have been correct, some were effectively just checking against output that the code happened to generate at the time of it generating the test. At the very least this will detect accidental changes.

**Changelog**

> ## Parley
> ### Fixed
>
> - **Note to the editor:** this entry should be merged with #819. Word boundaries of complex scripts, in the absence of dictionary data, now follow Unicode rules.


parley_data_gen/src/lib.rs
```diff
@@ -53,6 +53,11 @@ pub fn generate(out: std::path::PathBuf, config: &Config) {
         let is_emoji = emoji_data.contains32(cp);
         let is_extended_pictographic = extended_pictographic_data.contains32(cp);
 
+        // This encodes in which instances ICU4X uses a dictionary to determine a word break. (And
+        // deviates slightly from Unicode.)
```
DJMcNab: Is the deviation from unicode only in the Han/Hirigana special casing, or something else?

DJMcNab (approved):

parley_data_gen/src/lib.rs
tomcur: I've rewritten it a bit ("deviates" was a bit inaccurate; it's really that this unfortunately needs to track an ICU4X implementation detail), but yes. Han / Hiragana are not "Complex Context", but Unicode permits tailoring them. ICU4X chooses to tailor them, but doesn't tailor some other scripts.

PR #830, opened 2026-09-18
Remove `Boundary::Word`

tomcur:
LLM Contributions: Rewritten tests, review.

On top of #829.

Since #819, we now correctly track word and line boundaries separately. `Boundary::Word` is no longer read (in favor of `ClusterInfo::is_word_boundary`). This now removes that variant.

I'm thinking of a slightly bigger overhaul, and this PR itself is a tactical change as these things go, but I think it's worth doing: `is_word_boundary()` no longer implies `boundary >= Boundary::Word`.

I'm thinking to just track the `(grapheme, word, line break opportunity)` as separate bits. Note that would drop `Boundary::Mandatory`, which I believe is effectively unread already, because we encode that case `Whitespace::Newline`. That probably requires a bit more scrutiny than this change on its own, though.

**Changelog: None**


nicoburns (commented):
Change overall makes sense to me. With this change, I think all `Boundary` variants are now specifically line boundaries. So would it make sense to rename things to be explicitly `LineBoundary` and variants to something like:

```
pub enum LineBoundary {
    /// Not a boundary.
    None = 0,
    /// Soft line break opportunity
    Soft = 1,
    /// Hard line break.
    Hard = 2,
}
```

Or `Soft`/`Mandatory` or `Opportunity`/`Mandatory`. Bearing in mind there will likely be an additional variant for hyphenation at some point.

tomcur:
I'm working on a related change that removes `Boundary` altogether. Currently I'm thinking to fold the soft wrap opportunities into the flags. `Boundary::Mandatory` is redundant with `Whitespace::Newline`; and the `Mandatory` variant isn't read at all by `parley`.

This PR is really just to prevent `Boundary::Word` from being used again, as it's wrong.

Depending on how we handle hyphens, they might or might not be orthogonal. I believe Blink calculates hyphen positions lazily, and if we follow that model, they would not be known to analysis. If we do calculate them eagerly, it could be another bit alongside the soft wrap opportunity (or indeed by a variant in some enum). Literal soft-hyphen characters are separate (they impact analysis, so can't really be used to encode the possible positions, I think).

nicoburns (approved):
Let's merge this in the short term then

PR #833, opened 2026-09-20
Implement `WhiteSpaceCollapse::BreakSpaces` (soft break computation in analysis)

nicoburns:
LLM Contributions: Generated with GPT Astra 6 High

Builds on:

- https://github.com/linebender/parley/pull/812
- https://github.com/linebender/parley/pull/814

Supercedes:
- https://github.com/linebender/parley/pull/820

## Summary

Adds support for `WhiteSpaceCollapse::BreakSpaces`. This maps to the existing `Preserve` mode for actual whitespace collapsing, but has it's own unique behaviour during line-breaking and hanging. This is the remaining (widely implemented) value of `WhiteSpaceCollapse` needed to complete our whitespace handling.

## Context

When looking into @tomcur's review comments, I looked more closely at what `soft_line_break` from #820 was doing, and I noticed that it was computing additional soft line break opportunities during layout / content width computation. That seemed like a bit of a code smell to me, as in Parley we normally compute line break opportunities during analysis and just consume them during line-breaking / content width computation.

This PR implements `break-spaces` in that way (line breaking opportunities computed during analysis), and is both simpler (subjective, but true IMO) and faster (measured: no observable overhead vs. main!) than #820 (and this benefit composes with the optimizations from #822 / #823).

## Changes made

- Add `WhiteSpaceCollapse::BreakSpaces` enum variant
- Add `break_spaces: Vec<Range<usize>>` to `LayoutContext`. This populated with text ranges that have `WhiteSpaceCollapse` set to `BreakSpaces` similar to the existing `word_break`  Vec (so empty if `break-spaces` is not used).
- An additional check for `break-spaces` when computing line breaking opportunities in `analyze_text` (essentially free if `break-spaces` is not used (overhead not detectable by benchmarks at all), very cheap even when it is (something on the order of 0.5-1%))
- Minimal changes in line-breaking / content width computation to map the new enum variant to existing whitespace modes.

## Changelog

> ### Added
>
> - Support for `WhiteSpaceCollapse::BreakSpaces`


parley_engine/src/analysis.rs
```diff
@@ -656,6 +660,26 @@ pub(crate) fn analyze_text(
...
+        }
+        let is_break_space = break_spaces_iter
+            .peek()
+            .is_some_and(|range| range.contains(&byte_pos))
+            && matches!(ch, ' ' | '\t' | '\u{3000}');
+        if prev_is_break_space {
+            is_line = true;
+        } else if is_break_space {
+            is_line = false;
+        }
```
tomcur: We should ensure there isn't a soft wrap opportunity right before a mandatory break. I also think the `else if` branch needs to be dropped, otherwise e.g. a zero-width space followed by a tab inside a break-spaces region won't allow a line break, but [UAX #14 Rule LB8](https://www.unicode.org/reports/tr14/#LB8) says that case should allow a line break.

See also the related test fix in https://github.com/linebender/parley/pull/833#discussion_r4062726755.

```suggestion
        if prev_is_break_space && !properties.is_mandatory_linebreak() {
            is_line = true;
        }
```

Elsewhere we explicitly drop the soft wrap opportunity in case it's followed by a mandatory break: https://github.com/linebender/parley/blob/fbfdf958e9cb82e84f60e008ad512bb89b5950cd/parley_engine/src/analysis.rs#L659-L668.

parley/src/tests/test_analysis.rs
```diff
@@ -1189,3 +1191,99 @@ fn test_color_emoji_with_presentation() {
...
+            0..5,
+        );
+    })
+    .expect_boundary_list(vec![
+        Boundary::Word,
+        Boundary::Word,
+        Boundary::Line,
+        Boundary::Mandatory,
+        Boundary::Line,
+    ]);
```
tomcur, line 1258: With the fix from above, there shouldn't be a soft wrap opportunity right before the mandatory break.

```suggestion
    .expect_boundary_list(vec![
        Boundary::Word,
        Boundary::Word,
        Boundary::Word,
        Boundary::Mandatory,
        Boundary::Line,
    ]);
```

parley/src/tests/test_analysis.rs
```diff
@@ -1189,3 +1191,99 @@ fn test_color_emoji_with_presentation() {
         .expect_is_emoji_or_pictograph_list(vec![true, false])
         .expect_force_normalize_list(vec![false, false]);
 }
+
+#[test]
+fn test_break_spaces_suppresses_opportunity_before_first_space() {
```
tomcur: This name probably needs changing, as no opportunity is being suppressed, there's only one being added.

```suggestion
fn test_break_spaces_adds_opportunity_between_spaces() {
```

parley/src/tests/test_analysis.rs
```diff
@@ -1189,3 +1191,99 @@ fn test_color_emoji_with_presentation() {
...
+            0..4,
+        );
+    })
+    .expect_boundary_list(vec![
+        Boundary::Word,
+        Boundary::Word,
+        Boundary::None,
+        Boundary::Line,
+    ]);
+}
```
tomcur, line 1225: This one isn't really testing break spaces, so it's a bit of an odd-one-out. Maybe rename to `test_break_spaces_not_applied_under_preserve`?

parley/src/tests/test_analysis.rs
```diff
@@ -1189,3 +1191,99 @@ fn test_color_emoji_with_presentation() {
...
+        Boundary::Mandatory,
+        Boundary::Line,
+    ]);
+}
+
+#[test]
+fn test_break_spaces_across_style_boundary() {
+    // The opportunity after a preserved space applies even when the following character is in
+    // another style run, and no opportunity is created before the first space of a sequence
+    // spanning a style boundary.
```
tomcur: I'd tighten this a bit, as opportunities are never created before a space (`break-spaces` only creates additional ones after a space, regardless of whether the following character is a space).

```suggestion
    // The opportunity after a preserved space applies even when the following character is in
    // another style run.
```

parley_engine/src/analyzer.rs
```diff
@@ -52,6 +52,13 @@ pub struct AnalysisOptions<'a> {
     /// the text. Empty ranges are ignored. Gaps use [`WordBreak::Normal`].
     pub word_break: &'a [(Range<usize>, WordBreak)],
 
+    /// Ranges of the source text in which spaces and tabs are preserved and a soft wrap
+    /// opportunity follows each of them, i.e. CSS `white-space-collapse: break-spaces`.
+    ///
+    /// Ranges must be sorted and non-overlapping, and must start and end on character boundaries
+    /// of the text. Empty ranges are ignored.
+    pub break_spaces: &'a [Range<usize>],
```
tomcur: A bit more detail:

```suggestion
    /// Ranges of the source text in which a soft wrap opportunity follows every space, tab, and
    /// ideographic space.
    ///
    /// This implements the additional soft wrap opportunities of
    /// [CSS's white-space-collapse: break-spaces][css-break-spaces].
    /// 
    /// Outside these ranges, [UAX #14 § 6][uax-14-algorithm] is followed, which defines where soft
    /// wrap opportunities exist. For example, rules LB7 and LB18 give a sequence of spaces an
    /// opportunity only at its end. The ranges specified here add an opportunity after every space,
    /// tab, and ideographic space, except directly before a mandatory break. This overrides the
    /// "non-tailorable" UAX #14 Rule LB7, and so deviates from Unicode's line breaking algorithm.
    ///
    /// Ranges must be sorted and non-overlapping, and must start and end on character boundaries
    /// of the text. Empty ranges are ignored. Gaps apply the default rules.
    ///
    /// [css-break-spaces]: https://www.w3.org/TR/css-text-4/#valdef-white-space-collapse-break-spaces
    /// [uax-14-algorithm]: https://unicode.org/reports/tr14/#Algorithm
    pub break_spaces: &'a [Range<usize>],
```

tomcur (approved):
Nice! Looks good to me.

Approved conditional on suppressing the soft wrap opportunities before mandatory breaks (see the comments/suggestions inline). I've also proposed a bit more detail to put in the public docstring of `AnalysisOptions::break_spaces`.

Non-blocking: an LLM flagged that with the implementation as-is, a space followed by e.g. ZWSP or WJ also allow a line break, which I'm not sure is intentional. [CSS Text 4 § 6.5](https://www.w3.org/TR/css-text-4/#line-break-details) says those must be honored unless "explicitly defined otherwise".

nicoburns:
> Non-blocking: an LLM flagged that with the implementation as-is, a space followed by e.g. ZWSP or WJ also allow a line break, which I'm not sure is intentional. [CSS Text 4 § 6.5](https://www.w3.org/TR/css-text-4/#line-break-details) says those must be honored unless "explicitly defined otherwise".

My LLM claims that the current behaviour matches Chrome/Firefox/Safari, so I think I'll leave this as a follow-up (if indeed we decide we want to implement it).

